PR #55, opened 2015-06-03
Map page and back button

tpievila:
When user does:
1) Open map
2) Move map 
3) Click stop icon
4) Move to stop page
5) click back

Map is now in different zoom and place than in step 4. We should keep map in same zoom level and location in these kind on cases.

### DIFF
--- a/app/client.cjsx
+++ b/app/client.cjsx
@@ -2,6 +2,7 @@
 React             = require 'react'
 Router            = require 'react-router'
 FluxibleComponent = require 'fluxible/addons/FluxibleComponent'
+isEqual           = require 'lodash/lang/isEqual'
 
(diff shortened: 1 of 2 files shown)


tpievila:
Route summaries work, Android Chrome bug is quite elusive. Needs more specific reproduction steps to debug.

samuliheljo:
This is still broken. What I do is:
1) Use Remote debuggin like so: https://developer.chrome.com/devtools/docs/remote-debugging
2) Connect to localhost with mobile
3) on desktop browse to chrome://inspect/
4) Open Map on mobile
5) Move map a little bit
6) Now I can see in chrome://inspect/ that mobile is continuously updating browser location which seems to jam it

PR #58, opened 2015-06-03
Route page - map

hannesj:
As defined in concept


samuliheljo:
Noticed some bugs:
- Map does not work with mobile browser nor with chrome mobile simulation
- Refresh of page will crash whole app
- Cannot scroll itinerary downwards
- Desktop responsiveness is not ready yet
- When user turns phone to landscape itinerarys are not responsive "correctly"
- In itinerary text "1 pysäkkiä" (plural)

PR #91, opened 2015-06-25
Use changed time for fetching new itinerary

hannesj:


app/component/navigation/index-sub-navigation.cjsx
```diff
@@ -1,78 +1,22 @@
-React       = require 'react'
-TimeStore   = require '../../store/time-store' 
-TimeActions = require '../../action/time-action'
+React         = require 'react'
```
hannesj: These should probably be removed

tuukka: @hannesj What do you mean? Remove what from where?

hannesj: the requires on lines 2-4 are never used on the page

tuukka: I see, right. When I opened the link to the comment from chat, Github showed only line 1. "View full changes" helps :-)

Issue #65, opened 2015-06-09
Application does not load with IE9


samuliheljo:
Check what is status with IE9

At some point we got:
SCRIPT5022: matchMedia not present, legacy browsers require a polyfill bundle.js, rivi 119 merkki 16511"

Also, check that UI won't do infinite front page reload

hannesj:
Fixed (partially) by Financial-Times/polyfill-service#407 waiting for new release from them

PR #93, opened 2015-06-29
Use the fetch API (and polyfill it for unsupporting browsers)

hannesj:
Currently blocked, while waiting for a new version of polyfill-service which would fix issue created by matchMedia


hannesj:
@tpievila, could you take a look at this, would be nice to get it finally merged, as a new release including Financial-Times/polyfill-service#407 was released. Should also help fixing #65

Issue #46, opened 2015-05-20
Route summaries page - change start or stop location


samuliheljo:
- [ ] Start and stop address works like frontpage search
- [ ] After changing start or stop address user initiates search by mobile browser keyboard search button
- [ ] Search can be initiated with "magnifying class" icon as specified in concept
- [ ] Logic for search boxes should be similar to front page

samuliheljo:
Will be implemented in https://github.com/HSLdevcom/digitransit-ui/issues/71

PR #94, opened 2015-07-01
[WIP, do not merge] New search

hukka:
Current bugs:

- Fourth suggestion in list cannot be selected with touch (at least on Android), because the mouseEnter never fires and react-autosuggest hasn't saved the index
- On frontpage, the input field does not clear after using it to set origin


tpievila:
Related to #46

hukka:
Outdated by two search fields

Issue #77, opened 2015-06-09
Departure page


samuliheljo:
As defined in concept:
https://projects.invisionapp.com/share/MY2F0CQ2W#/screens/79444040?maintainScrollPosition=false

hannesj:
Related to #107

PR #98, opened 2015-07-09
Navigation page

hannesj:


app/store/location-store.coffee
```diff
@@ -73,14 +79,26 @@ class LocationStore extends Store
   getLocationString: () ->
     "#{@address}::#{@lat},#{@lon}"
 
+  storeWatchId: (watchId) ->
+    @watchId = watchId
+
+  clearWatchId: ->
+    @watchId = undefined
+
+  getWatchId: ->
```
tpievila: Is the id meant to be stored in the caller (function doesn't bind this)?

app/util/geo-utils.coffee
```diff
@@ -0,0 +1,11 @@
...
+toDeg = (rad) -> rad * 180 / Math.PI
+
+getBearing = (lat1,lng1,lat2,lng2) ->
+  lonScale = Math.cos toRad (lat1+lat2)/2
+  dy = lat2-lat1
+  dx = (lng2-lng1)*lonScale
+  (toDeg(Math.atan2(dx, dy)) + 360) % 360
+
+module.exports =
+  getBearing: getBearing
```
tpievila: lint:

  ⚡ app/util/geo-utils.coffee
     ⚡ #4: Spaces are required after commas. ,.
     ⚡ #4: Spaces are required after commas. ,.
     ⚡ #4: Spaces are required after commas. ,.
     ⚡ #5: Operators must be spaced properly. +.
     ⚡ #5: Operators must be spaced properly. /.
     ⚡ #6: Operators must be spaced properly. -.
     ⚡ #7: Operators must be spaced properly. -.
     ⚡ #7: Operators must be spaced properly. *.

app/component/map/navigation-map.cjsx
```diff
@@ -0,0 +1,124 @@
...
+  componentWillUnmount: ->
+    @context.executeAction LocateActions.stopLocationWatch
+    if @intervalId
+      clearInterval(@intervalId)
+      @intervalId = undefined
+
+  render: ->
+    <div id="map" className="fullscreen"/>
+
+module.exports = NavigationMap
```
tpievila: lint

  ✗ app/component/map/navigation-map.cjsx
     ⚡ #65: Invoking a constructor without parens and with arguments.
     ⚡ #77: Invoking a constructor without parens and with arguments.
     ✗ #87: Line exceeds maximum allowed length. Length is 105, max is 100.
     ⚡ #89: Invoking a constructor without parens and with arguments.

app/component/map/navigation-map.cjsx
```diff
@@ -0,0 +1,124 @@
...
+class NavigationMap extends React.Component
+  @contextTypes:
+    getStore: React.PropTypes.func.isRequired
+    executeAction: React.PropTypes.func.isRequired
+
+  componentDidMount: ->
+    @context.executeAction LocateActions.startLocationWatch
+
+    require.ensure ['mapbox-gl/dist/mapbox-gl-dev', './mapbox-gl.css'], =>
+      require './mapbox-gl.css'
```
tpievila: Shouldn't the callback take require as an argument?

Unusual import (not at the top, like everything else) is worthy of a comment too.

app/component/map/navigation-map.cjsx
```diff
@@ -0,0 +1,124 @@
...
+          plan.legs[0].to.lat,
+          plan.legs[0].to.lon)
+
+      mapLoaded = new Promise (resolve) -> map.on 'load', resolve
+
+      Fulltilt = require('exports?window.FULLTILT!fulltilt/dist/fulltilt.js')
+      fulltiltLoaded = Fulltilt.getDeviceOrientation(type: 'world')
+
+      mapLoaded.then =>
+        #debug("map loaded")
```
tpievila: No dead code into repos

app/component/map/navigation-map.cjsx
```diff
@@ -0,0 +1,124 @@
...
+
+        @locationJSONsource = new mapboxgl.GeoJSONSource data: locationAsGeoJSON coordinates
+        map.addSource 'location', @locationJSONsource
+        map.addLayer getLayerforLocation()
+
+        fulltiltLoaded.then(
+          (orientation) => @initializeCompass(map, true, orientation)
+        , (errror) => @initializeCompass(map, false)
+        )
+    , 'mapboxgl'
```
tpievila: Somewhat a PITA to see to which call this is a parameter for, when the actual call might not even fit on the display at the same time. A comment would be good.

Why do we need a named chunk here?

app/component/map/navigation-map.cjsx
```diff
@@ -0,0 +1,124 @@
...
+      newCoordinates = @context.getStore('LocationStore').getLocationState()
+      if newCoordinates.lon != coordinates.lat or newCoordinates.lon != coordinates.lon
+        @locationJSONsource.setData locationAsGeoJSON coordinates
+        coordinates = newCoordinates
+      if compassAvailable
+        euler = orientation.getScreenAdjustedEuler()
+      map.easeTo
+        center: [coordinates.lat, coordinates.lon]
+        bearing: if compassAvailable then 360 - euler.alpha else coordinates.heading
+        pitch: if compassAvailable and isNaN(euler.beta) then Math.max(euler.beta, 0) else 45
```
tpievila: Wait, what? If beta is not a number, take a max of it and 0?

app/page/navigation.cjsx
```diff
@@ -0,0 +1,11 @@
+React              = require 'react'
+DefaultNavigation  = require '../component/navigation/default-navigation'
+NavigationMap      = require '../component/map/navigation-map'
+
+class NavigationPage extends React.Component
+  render: ->
+    <DefaultNavigation className="fullscreen">
+      <NavigationMap hash={parseInt(@props.params.hash)}/>
```
tpievila: It's not obvious what the hash represents and what it's used for.

app/store/location-store.coffee
```diff
@@ -46,6 +46,12 @@ class LocationStore extends Store
     @status = @STATUS_FOUND_LOCATION
     @emitChange()
 
+  storeLocationWithoutEmit: (location) ->
+    @lat = location.lat
+    @lon = location.lon
+    @heading = if location.heading then location.heading else @heading
+    @status = @STATUS_FOUND_LOCATION
+
```
tpievila: Would be good to comment here shortly why we sometimes want to emit and sometimes not.

This is pretty much the same as storeLocation, but heading is missing in the old function. If intentional, would be good to insert a comment on that.

Also, could the storeLocation simply call storeLocationWithoutEmit and then emit, to remove code duplication?

app/util/geo-utils.coffee
```diff
@@ -0,0 +1,11 @@
+toRad = (deg) -> deg * Math.PI / 180
+toDeg = (rad) -> rad * 180 / Math.PI
+
+getBearing = (lat1,lng1,lat2,lng2) ->
```
tpievila: I think something short like:

    // Get direction of movement from two first to second point.
    // North is 0, east is 90, south 180, west 270

server/server.js
```diff
@@ -32,6 +32,10 @@ function setUpStaticFolders() {
   app.use(rootPath +"css", express.static(cssFolder))
   var jsFolder = path.join(staticFolder, 'js')
   app.use(rootPath +"js", express.static(jsFolder))
+  var mapFontsFolder = path.join(staticFolder, 'map', 'fonts')
+  app.use(rootPath +"mapFonts", express.static(mapFontsFolder, {setHeaders: function(res){res.setHeader("Content-Encoding","gzip")}}))
+  var mapFolder = path.join(staticFolder, 'map')
+  app.use(rootPath +"map", express.static(mapFolder))
```
tpievila: We don't use jslint, but I think that one line is pretty long and that there should be whitespace around operators.

app/store/location-store.coffee
hannesj: Fluxible binds it automatically

app/util/geo-utils.coffee
hannesj: Fixed

app/component/map/navigation-map.cjsx
hannesj: Fixed

app/component/map/navigation-map.cjsx
hannesj: I changed the import logic quite a lot, now all of the requires are in the componentWillMount, where they are asynchronously required. The componentDidMount will then continue loading them after they have been downloaded.

app/component/map/navigation-map.cjsx
hannesj: Removed

app/component/map/navigation-map.cjsx
hannesj: The changed require process removed this line. Named chunks are nice with debugging.

app/component/map/navigation-map.cjsx
hannesj: Whoops, there seems to be a exclamation mark missing here after combining the two handlers into one. Fixed

app/page/navigation.cjsx
hannesj: Comment added

app/store/location-store.coffee
hannesj: Fixed according to your suggestion, nicer that way.

app/util/geo-utils.coffee
hannesj: Added

server/server.js
hannesj: Fixed all the static file definitions.

PR #99, opened 2015-07-13
Vehicles on navigation map

hannesj:


hannesj:
Currently all vehicles are pointing to south. For this to be fixed, data-driven styles need to be implemented in mapbox/mapbox-gl-js (coming in v9).

app/action/real-time-client-action.coffee
```diff
@@ -52,28 +52,35 @@ parseMessage = (topic, message, actionContext) ->
 module.exports =
   startRealTimeClient: (actionContext, options, done) ->
     #Fetch initial data
-    xhrPromise.getJson(config.URL.REALTIME + (getTopic(options)).replace('#', '')).then (data) ->
-      parseMessage(topic, message, actionContext) for topic, message of data
+    if !Array.isArray(options)
```
samuliheljo: Should we restrict that options is always array. Now this makes me think: "why it is called in different ways"

samuliheljo:
It Seems that "back"-button is no longer working in navigation

PR #143, opened 2015-09-18
Refactoring styles

samuliheljo:


samuliheljo:
Bugs & something to fix:
- Stop names on map look different than before
- Disruption info text seem more gray than before

Minor nitpicking
- In some style files (e.g. _elements.scss, _base.scss) there are 0-4 empty lines between selectors. It would be nice to have e.g. 1 empty line.

PR #155, opened 2015-09-29
Trip page

MikaelStenstrand:


app/action/real-time-client-action.coffee
```diff
@@ -6,7 +6,8 @@ xhrPromise = require '../util/xhr-promise'
 getTopic = (options) ->
   route = if options.route then options.route else "+"
   direction = if options.direction then parseInt(options.direction) + 1  else "+"
-  '/hfp/journey/+/+/' + route + '/' + direction + '/#'
+  trip = if options.trip then options.trip else "+"
```
samuliheljo: I have wondered this row before. Why options might or might not contain ".trip", I probably know the answer, but we should comment this

app/component/map/vehicle-marker-container.cjsx
```diff
@@ -38,7 +38,8 @@ class VehicleMarkerContainer extends React.Component
       @context.executeAction RealTimeInformationAction.startRealTimeClient
     @context.getStore('RealTimeInformationStore').addChangeListener @onChange
     for id, message of @context.getStore('RealTimeInformationStore').vehicles
-      @updateVehicle(id, message)
+      if !@props.trip || message.tripStartTime == @props.trip
```
samuliheljo: I don't quite get this. why do we sometimes do update but sometimes not?

app/component/route/route-header-container.cjsx
```diff
@@ -31,10 +31,13 @@ class RouteHeaderContainer extends React.Component
     if reverseIds.length == 1
       reverseId = reverseIds[0]
 
+    #TODO: all of trips should be reanmed pattern
```
samuliheljo: "reanmed"?

app/component/route/route-header.cjsx
```diff
@@ -2,24 +2,25 @@ React                 = require 'react'
 Icon                  = require '../icon/icon'
 GtfsUtils             = require '../../util/gtfs'
 Link                  = require 'react-router/lib/Link'
-
+classnames            = require 'classnames'
 
 class RouteHeader extends React.Component
```
samuliheljo: This is a larger question: Should we have somewhere explanation what props is supported by component?
Now there are "@props.className", "@props.trip", "@props.reverseId", "@props.route". It's not obvious what e.g. this means:
```
if @props.reverseId && !@props.trip
       reverse = <Link to="#{process.env.ROOT_PATH}linjat/#{@props.reverseId}">
           <Icon className={"route-header-direction-switch " + mode} img={'icon-icon_direction-b'}/>
         </Link>
```

app/page/trip.cjsx
```diff
@@ -0,0 +1,62 @@
...
+    return hours + mins
+
+  render: ->
+    trip = @getStartTime(@props.trip.stoptimes[0].scheduledDeparture)
+
+    <DefaultNavigation className="fullscreen trip">
+      <RouteHeaderContainer className="trip-header" route={@props.trip.pattern} trip={trip}/>
+      <RouteMapContainer className="map-small" route={@props.trip.pattern} trip={trip} tripId={@props.trip.gtfsId}/>
+      <RouteListHeader
+        headers={["Aika", "Pysäkki", "Pysäkkinumero"]}
```
samuliheljo: Localizations here

app/page/trip.cjsx
```diff
@@ -0,0 +1,62 @@
...
+  getStartTime: (time) ->
+    hours = ('0' + Math.floor(time / 60 / 60 )).slice(-2)
+    mins = ('0' + (time / 60 % 60)).slice(-2)
+    return hours + mins
+
+  render: ->
+    trip = @getStartTime(@props.trip.stoptimes[0].scheduledDeparture)
+
+    <DefaultNavigation className="fullscreen trip">
+      <RouteHeaderContainer className="trip-header" route={@props.trip.pattern} trip={trip}/>
```
samuliheljo: Trip page route number should be clickable. e.g. "51"

app/page/trip-map.cjsx
```diff
@@ -0,0 +1,41 @@
+React                  = require 'react'
```
samuliheljo: This is a general question: Should we make this kind of pages as own pages or like on Route page, where there is only one page?

MikaelStenstrand: This is implemented to the same way as the old front page was done. This implementation enables the map to be shown and go into fullscreen onClick, whereas the route page has the map under a tab. This depends on the requirements for the page

app/component/route/route-header.cjsx
MikaelStenstrand: we could use the PropTypes on every component both documentation (which props are injected to the component) and also to ensure that the type is correct of the props.
https://facebook.github.io/react/docs/reusable-components.html

PR #164, opened 2015-10-07
Input field style fix

samuliheljo:
Fix style bug that causes search fields smash into each other when 'current position' is used as destination location

screen shot 2015-10-07 at 15 20 03


samuliheljo:
let me do this again...

PR #167, opened 2015-10-08
Language ui (DT-183)

ffsiren:


app/component/navigation/lang-select.cjsx
```diff
@@ -0,0 +1,36 @@
...
+    @context.getStore(PREFERENCES_STORE_NAME).addChangeListener @onChange
+
+  componentWillUnmount: ->
+    @context.getStore(PREFERENCES_STORE_NAME).removeChangeListener @onChange
+
+  onChange: (lang) =>
+    @forceUpdate()
+
+  selectLanguage: (lang) ->
+    () =>
```
samuliheljo: Why double function?

app/component/navigation/index-navigation.scss
```diff
@@ -111,6 +111,24 @@ nav.top-bar {
   background: transparent;
 }
 
+.lang-select {
```
samuliheljo: Should we put this in own file?

app/component/navigation/lang-select.cjsx
ffsiren: selectLanguage returns a function that is called on a#onClick

PR #171, opened 2015-10-08
Ui tweaks

Macroz:
Tweak the Nearby Routes UI to match proto 
 - styles closed to proto
 - remove extra scroll
 - include link


app/component/route/route.scss
```diff
@@ -33,7 +34,8 @@ $route-header-favourite-icon-selected-hover-fill-color: $background-color;
   overflow: hidden;
   color: $gray;
   font-size: $font-size-small;
-  margin-left: 1em;
+  margin-left: 0px;
```
tuukka: I don't know if we have a CSS style guide but I think it says use "0" instead of "0px" ;-)

app/component/route/route.scss
```diff
@@ -25,7 +25,8 @@ $route-header-favourite-icon-selected-hover-fill-color: $background-color;
 }
 
 .departure-list-header {
-  border-bottom: solid 1px #eef1f3;
+  border-top: 1px solid $light-gray;
+  border-bottom: 1px solid $light-gray;
```
tuukka: "solid 1px" is better style than "1px solid", I would say.

Macroz: Actually 1px solid is better, W3C would say :)
https://www.w3.org/community/webed/wiki/CSS_shorthand_reference

app/component/route/route.scss
Macroz: Good catch, fixed.

app/component/route/route.scss
```diff
@@ -211,12 +213,17 @@ div.route-tabs {
 }
 
 .nearby-routes {
-  overflow-y: auto;
+  overflow: hidden;
```
tuukka: What can this hide? Should we also have white-space: nowrap?

Macroz: There was a double overflow (the whole tab, and the list below) which was not good. So it's better to just hide the one for the whole tab contents. There's actually not going to be any need to scroll when the extra scrollbar is gone.

PR #178, opened 2015-10-14
Sentry

hukka:
DT-213 DT-214, error logging from client and nodejs server


hannesj:
@hukka could you add the DSN and the secret to the ansible vault and adapt the build scripts in digitransit-deploy for this?

PR #179, opened 2015-10-14
DT-225 Nearby routes: add stop number

Macroz:
[DT-255 Nearby routes: add stop number](https://digitransit.atlassian.net/browse/DT-225)

Adds a component called StopReference that implements the showing of the stop number in a Departure. This new component is taken into use in the RouteListContainer in use in the nearby routes tab.

Also includes the component documentation and a small refactoring thereof.

### DIFF
--- a/app/queries.js
+++ b/app/queries.js
@@ -379,6 +379,9 @@ var DepartureListFragments = {
         realtimeDeparture
         realtime
         serviceDay
+	stop {
+	  code
+	}
       }
     }
   `,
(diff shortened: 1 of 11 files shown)


app/queries.js
```diff
@@ -379,6 +379,9 @@ var DepartureListFragments = {
         realtimeDeparture
         realtime
         serviceDay
+	stop {
```
hannesj: Please use spaces

PR #186, opened 2015-10-15
Improvements for summary page

hannesj:


app/component/summary/summary-row.scss
```diff
@@ -1,5 +1,6 @@
 .itinerary-summary-row {
   padding-left: 0.5em;
+  padding-right: 0.5em;
```
hannesj: Should we use padding-small class instead of where this is used

Macroz: .padding-small is likely good for normal uses, but I think in this case, where there is a definite lack of space, we probably just want to optimize it to the max by itself. E.g. using .padding-small here will add more extra vertical whitespace around this component.

Anyway, I added an overflow hidden here, because the last leg text was causing horizontal scrollbars to appear in some cases.

PR #189, opened 2015-10-16
DT-129 sliders

hukka:


samuliheljo:
Something to fix. Compared with the one deployed to matka.hsl.fi
- Row height has incresed. Too much vertical padding or margin.
- Firefox slider icon size has increased.

samuliheljo:
NEW: new
OLD: old

PR #207, opened 2015-10-22
Use color variable name that is visible also in non hsl version

ffsiren:
Makes the "not implemented" popup display title,icon, links also in non hsl version


samuliheljo:
I created a separate issue for this: https://digitransit.atlassian.net/browse/DT-302

PR #209, opened 2015-10-22
Show 404-not-found when looking at route page with illegal id

ffsiren:
https://digitransit.atlassian.net/browse/DT-255


hukka:
Snap failed again in iPhone tests...

PR #212, opened 2015-10-22
Refactor search

hannesj:


Macroz:
Seems to fix also [DT-249](https://digitransit.atlassian.net/browse/DT-249)

PR #214, opened 2015-10-22
Ui summary  duration and link

VilleMiekkoja:


samuliheljo:
Let's refactor calculation in summary-row.cjsx and duration.cjsx so that we have only one place where this time calcution is done.

PR #219, opened 2015-10-26
Center map on stop page DT-276

ffsiren:
https://digitransit.atlassian.net/browse/DT-276


app/page/stop-map.cjsx
```diff
@@ -20,15 +20,16 @@ class Page extends React.Component
       @context.executeAction FavouriteStopsAction.addFavouriteStop, @props.params.stopId
 
     <DefaultNavigation className="fullscreen">
-      <Map lat={@props.stop.lat+0.0005}
+      <StopCardHeader stop={@props.stop} favourite={favourite} addFavouriteStop={addFavouriteStop} headingStyle="h3" className="stop-page" infoIcon={true}/>
+      <Map lat={@props.stop.lat}
            lon={@props.stop.lon}
-           zoom={18}
+           zoom={16}
```
hannesj: This should not be changed, check with Petro what the zoom level should be.

hannesj:
The stop page becomes slightly too log, so we get dual scroll bars.

app/page/stop-map.cjsx
ffsiren: Checked with Petro: it's ok to make zoom smaller (so that you see slightly more context).

PR #221, opened 2015-10-27
Update realtime views

hannesj:


samuliheljo:
National icon animation position could be better
screen shot 2015-10-29 at 12 10 11

samuliheljo:
Do we have different spacing for bus and tram icons? It seems that animation is "closer to tram" than on bus

app/component/itinerary/transit-leg.cjsx
```diff
@@ -8,12 +8,27 @@ FormattedMessage = intl.FormattedMessage
 class TransitLeg extends React.Component
 
   render: ->
+    originalTime = if @props.leg.realTime and @props.leg.departureDelay >= 180 then [
```
samuliheljo: I would move this "180" to config file

samuliheljo:
Otherwise looks good!

PR #239, opened 2015-11-04
Routing: Make maximum biking distance configurable in file

ffsiren:


samuliheljo:
There is something strange in this. If I:
1) Open hsl version
2) take a trip from opastinsilta to kera
3) Select only "subway" and "bicycling"
4) it still shows biking as one possible option

It seems that otp is not using maxwalkingdistance for biking?

hannesj:
actually the name is not completely accurate, as it is the maximum distance from the start point and to the destination, so the distance can be double to the specified.

PR #248, opened 2015-11-12
Dt 379 refactor stop marker popup

VilleMiekkoja:


app/component/map/marker-popup-bottom.cjsx
```diff
@@ -0,0 +1,19 @@
+React       = require 'react'
+ComponentUsageExample = require '../documentation/component-usage-example'
+Example = require '../documentation/example-data'
```
hukka: These three aren't used. Should they be deleted, or docs added?

PR #257, opened 2015-11-19
Fix DT-394, dragging itinerary map changes tab

hukka:


samuliheljo:
does not work with samsung s4 android :(

PR #258, opened 2015-11-19
Fix DT-376 Too many markers in ui

hukka:


app/component/map/map.cjsx
```diff
@@ -38,9 +38,6 @@ class Map extends React.Component
     if isBrowser
       origin = @context.getStore('EndpointStore').getOrigin()
 
-      if origin?.lat
-        fromMarker = <LocationMarker position={origin} className="from"/>
```
hannesj: Hides marker on front page, where it should be visible, possibly same on eg. stop, route and trip pages.

samuliheljo:
1) now we are missing FROM icon when user is using his own position.
2) look at screen shot, how it should be interpreted? (I admit that this was case also before but it seems that this fix will not make situation that much better)

screen shot 2015-11-20 at 11 04 11

PR #267, opened 2015-11-25
Implement new user position icon

VilleMiekkoja:
Quite much code repetition. Tried find a solution without repetition, but didn't find it in this time. Feel free to suggest a one.


hannesj:
This icon is already used on stop and route pages, icon-icon_mapMarker-location-animated should be replaced instead.

PR #279, opened 2015-11-30
choose popup to open based on distance

tuukka:


samuliheljo:
Generally seems to make stop clicking much better, good. Some review questions, I add those...

app/component/map/generic-marker.cjsx
```diff
@@ -60,9 +60,14 @@ class GenericMarker extends React.Component
...
     return nextProps.id != @props.id
 
+  openPopup: =>
+    @refs['marker'].getLeafletElement().openPopup()
+
   getMarker: ->
     <Marker map={@props.map}
+            ref="marker"
             position={lat: @props.position.lat, lng: @props.position.lon}
+            interactive={false}
```
samuliheljo: What is this, could not find it?
https://github.com/PaulLeCam/react-leaflet/blob/master/src/Marker.js

tuukka: The `react-leaflet/Marker` component forwards all unhandled props like this to the underlying `leaflet/marker` implementation. We use Leaflet 1.0 beta where `clickable` has been renamed as `interactive`: http://mourner.github.io/Leaflet/reference.html#marker Setting `interactive={false}` stops the marker from handling events so that the map gets to handle them.

samuliheljo: I would maybe add that as a comment. (or at least part of that). What do you think.

I have no other comments for pull request..

hannesj: I would still like to have markers handle direct clicks, this e.g. renders markers on route and itinerary pages unclickable.

samuliheljo: Confirmed, those 2 pages are not working

tuukka:
Obsoleted by #333 SVG stop layer.

PR #281, opened 2015-12-01
Dt 389 show terminals on map

VilleMiekkoja:


static/svg-sprite.svg
```diff
@@ -224,12 +224,16 @@
 	<path class="path6" d="M967.118 56.882h-910.235v512.007h910.235c31.422 0 56.89-25.468 56.89-56.89v-398.228c0-31.422-25.468-56.89-56.89-56.89zM881.783 398.221c0 15.711-12.734 28.445-28.445 28.445h-59.371c1.579 4.454 2.482 9.226 2.482 14.222 0 23.564-19.103 42.667-42.667 42.667s-42.667-19.103-42.667-42.667c0-4.996 0.903-9.768 2.482-14.222h-374.747c1.579 4.454 2.482 9.226 2.482 14.222 0 23.564-19.103 42.667-42.667 42.667s-42.667-19.103-42.667-42.667c0-4.996 0.903-9.768 2.482-14.222h-59.371c-15.711 0-28.445-12.734-28.445-28.445v-199.114c0-15.711 12.734-28.445 28.445-28.445h654.231c15.711 0 28.445 12.734 28.445 28.445v199.114z"></path>
 	<path class="path7" d="M711.114 199.107h142.224v85.335h-142.224v-85.335z"></path>
 </symbol>
-<symbol id="icon-icon_station" viewBox="0 0 1024 1024">
```
hannesj: This symbol is used in other places, probably create a new one for this use?

app/component/map/stop/terminal-marker.cjsx
```diff
@@ -1,20 +1,15 @@
...
 Relay         = require 'react-relay'
 queries       = require '../../../queries'
+geoUtils      = require '../../../util/geo-utils'
 isBrowser     = window?
 Icon          = require '../../icon/icon'
-StopMarkerPopup = require './stop-marker-popup'
-provideContext = require 'fluxible-addons-react/provideContext'
 intl          = require 'react-intl'
 GenericMarker = require '../generic-marker'
+L             = if isBrowser then require 'leaflet' else null
```
hannesj: Isn't needed?

app/component/map/stop/terminal-marker.cjsx
```diff
@@ -28,19 +23,14 @@ class TerminalMarker extends React.Component
     Icon.asString 'icon-icon_station', 'terminal-medium-size'
```
hannesj: Why to expose a function instead of static property?

static/svg-sprite.svg
```diff
@@ -224,6 +224,17 @@
 	<path class="path6" d="M967.118 56.882h-910.235v512.007h910.235c31.422 0 56.89-25.468 56.89-56.89v-398.228c0-31.422-25.468-56.89-56.89-56.89zM881.783 398.221c0 15.711-12.734 28.445-28.445 28.445h-59.371c1.579 4.454 2.482 9.226 2.482 14.222 0 23.564-19.103 42.667-42.667 42.667s-42.667-19.103-42.667-42.667c0-4.996 0.903-9.768 2.482-14.222h-374.747c1.579 4.454 2.482 9.226 2.482 14.222 0 23.564-19.103 42.667-42.667 42.667s-42.667-19.103-42.667-42.667c0-4.996 0.903-9.768 2.482-14.222h-59.371c-15.711 0-28.445-12.734-28.445-28.445v-199.114c0-15.711 12.734-28.445 28.445-28.445h654.231c15.711 0 28.445 12.734 28.445 28.445v199.114z"></path>
 	<path class="path7" d="M711.114 199.107h142.224v85.335h-142.224v-85.335z"></path>
 </symbol>
+<symbol id="icon-icon_station--onmap" viewBox="0 0 2048 2048">
+	<title>icon_station</title>
+	<circle key="halo" class="stop-halo" cx="1024" cy="1024" r="989" stroke-width="70" stroke="#333" stroke-opacity="0.4" fill="#fff" fill-opacity="1"/>
+	<circle class="station" cx="1024" cy="1024" r="860"/>
+	<svg x="28%" y="28%" viewBox="0 0 2348 2348">
+		<path class="path1 fill-color13" d="M-0.007 80.935l1023.776-80.823v80.826l-1023.776 269.413v-269.417z"></path>
```
hannesj: Why are these lines duplicated from the other icon? They could be included by using an use-tag. Also, you probably shouldn't use an svg element inside an svg, but rather just define a new group using a g-element.

app/util/geo-utils.coffee
```diff
@@ -98,6 +98,25 @@ getDistanceToNearestStop = (lat, lon, stops) ->
     distance: minDist
   }
 
+getDistanceToFurthestStop = (lat, lon, stops) ->
```
hannesj: This could be simplified a bit with writing it something like 

```
getDistanceToFurthestStop = (coordinates, stops) ->
  stops.map stop ->
    stop: stop
    distance: coordinates.distanceTo new L.LatLng stop.lat, stop.lon
  .reduce (previous, current) ->
    if current.distance > previous.distance then current else previous
  , stop: null, distance: 0
```

app/queries.js
```diff
@@ -345,6 +345,11 @@ var StopMarkerLayerFragments = {
           name
           lat
           lon
+          stops {
```
hannesj: It might be a bit excessive to ask for parent stops for each stop and each child stop for all of those. It might a bit too big to refactor that in this task, but just a s a future remark.

static/svg-sprite.svg
VilleMiekkoja: 1. The lines are not identical. 2. g-elements have no x and y attributes

hannesj: The only difference is the class, which you can also give to the parent, right? Also, with use, you can use x and y

PR #285, opened 2015-12-03
DT-118 fix icons

hukka:


app/component/map/map.scss
```diff
@@ -88,14 +88,32 @@ div.current-location-marker > span > svg.icon {
 }
 
 div.leaflet-marker-icon.from, div.leaflet-marker-icon.to {
-  > span > svg.icon {
-    width: 24px;
-    height: 24px;
-    stroke: white;
-    stroke-width: 60px;
+  > span {
+    &::before {
```
hannesj: Can you make the ::before to the svg instead of span?

app/component/map/map.scss
```diff
@@ -88,14 +88,32 @@ div.current-location-marker > span > svg.icon {
...
+      width: 24px;
+      height: 24px;
+      stroke: white;
+      stroke-width: 60px;
+    }
   }
 }
 
+div.leaflet-marker-icon.from > span::before {
+  content: "A";
```
hannesj: Same as above?

app/component/map/map.scss
```diff
@@ -88,14 +88,32 @@ div.current-location-marker > span > svg.icon {
...
-  > span > svg.icon {
-    width: 24px;
-    height: 24px;
-    stroke: white;
-    stroke-width: 60px;
+  > span {
+    &::before {
+      color: white;
+      position: absolute;
+      left: 6.4pt;
```
samuliheljo: For mac chrome and firefox:
left: 6.0pt;
top: 0pt;

works better. what do you think?
screen shot 2015-12-03 at 14 01 50

PR #290, opened 2015-12-07
Dt 487

ffsiren:


app/store/local-storage.coffee
```diff
@@ -0,0 +1,15 @@
...
+      catch error
+        if error.name == 'QuotaExceededError'
+          console.warn('[localStorage] Unable to save state; localStorage is not available in Safari private mode')
+
+  getItem: (k) ->
+    if window?.localStorage?
+      window.localStorage.getItem k
+
+  removeItem: (k) ->
+    window?.localStorage?.removeItem STORAGE_KEY
```
hannesj: this should probably be `window?.localStorage?.removeItem k` ?

PR #293, opened 2015-12-08
prevent selecting text of tab buttons

ffsiren:


sass/base/_base.scss
```diff
@@ -140,6 +140,13 @@ body select  {
   margin-bottom: 0;
 }
 
+.no-select {
+  -webkit-user-select: none;
+  -moz-user-select: none;
+  -ms-user-select: none;
+  user-select: none;
```
hannesj: This is taken care by autoprefixer, you only need to have `user-select`

PR #296, opened 2015-12-09
DT-496 front page push state

Macroz:
Contains
* [DT-496 Use history push state for front page tabs, so that they can be closed by back button
](https://digitransit.atlassian.net/browse/DT-496)
* [DT-504 Use history push state for offcanvas](https://digitransit.atlassian.net/browse/DT-504)

Had to fix rackt/history. Currently using fork at [Macroz/history](https://github.com/Macroz/history). See PR https://github.com/rackt/history/pull/179 and issue https://github.com/rackt/history/issues/178


Macroz:
Now without fixed/forked history but using instead pinned version 1.13.0 (1.13.1 is already broken).

PR #297, opened 2015-12-10
Search field fixes

ffsiren:
fixes: DT-432, DT-454, DT-491


app/component/search/autosuggest.cjsx
```diff
@@ -110,15 +105,11 @@ class Autosuggest extends React.Component
           # 2) when we notice that everything is cleared, we remove location from flux store (not ok)
           # react-autosuggest should support second case, but it currently doesn't
           if input == ""
-            @context.executeAction @props.onEmptyAction
-          return input.trim().length >= 2
+            @props.onEmptyAction()
```
hannesj: This could be renamed to just ´onEmpty´ as it is not a Fluxible action anymore

hannesj:
Seems to work really nicely, the only bug I found is that the search fields won't update when going forwards/backwards in history, but I'm not sure how we should handle that.

hannesj:
Ok, that doesn't work currently either. I'll suggest we create new issues for those and merge this as is.

PR #299, opened 2015-12-10
DT-479: url encode names

ffsiren:


app/component/search/search-two-fields-container.cjsx
```diff
@@ -107,7 +107,7 @@ class SearchTwoFieldsContainer extends React.Component
       # Then we can transition. We must do this in next
       # event loop in order to get blur finished.
       setTimeout(() =>
-        @context.history.pushState(null, "/reitti/#{from}/#{to}")
+        @context.history.pushState(null, "/reitti/#{encodeURIComponent(from)}/#{encodeURIComponent(to)}")
```
hannesj: This is only one of the uses of linking to routing. Probably should implement something like [getLocationString](https://github.com/HSLdevcom/digitransit-ui/blob/master/app/store/position-store.coffee#L80)(which should also be updated to use `encodeURIComponent`) and use that everywhere

PR #304, opened 2015-12-15
search field fixes 

ffsiren:
fixes: DT-495, DT-526, DT-528, DT-505


samuliheljo:
Bug:
1) Let UI get your position
2) Set destination Kamppi
3) You are shown a route
4) activate destination field, and just click on map. (this clears destination, which is weird)
5) Click back, you are now on front page
6) Click nearby routes, UI does a route search. <-- Bug

*Some thoughts*
It seems that this is quite hard to get "right". This might be due that whole concept is not that clear now ui should work in which situation.

*What to do?*
We should:
a) revisit concept
b) create clear spec how search should behave in different situations.

PR #313, opened 2015-12-20
DT-430 document meta

Macroz:
Implements [DT-430](https://digitransit.atlassian.net/browse/DT-430) with document meta handling using [react-helmet](https://github.com/nfl/react-helmet/). 

All the meta tags are moved to a new TopLevel -component (wrapper of all routes) and meta.coffee.


ffsiren:
It looks like the metadata is not populated (not even the title) for pages that utilize relay. After discussions it was decided that this is fine. New tasks will be added to backlog that will fix this limitation.

PR #321, opened 2016-01-08
Dt 525

ffsiren:


Macroz:
Testing manually works just fine. Automatic test problems should be fixed in other pull requests, because they don't concern this feature.

PR #323, opened 2016-01-11
fix for 'DOM Exception 22'

ffsiren:
session storage is not available in mobile safari private browsing mode


app/store/itinerary-search-store.coffee
```diff
@@ -176,7 +176,12 @@ class ItinerarySearchStore extends Store
 
   storeItinerarySearch: (data) ->
     @data = data
-    window?.sessionStorage?.setItem STORAGE_KEY, JSON.stringify @data
+    try
+      window?.sessionStorage?.setItem STORAGE_KEY, JSON.stringify @data
+    catch error
+      if error.name == 'QuotaExceededError'
```
tuukka: else re-throw, to not hide unexpected errors?

PR #325, opened 2016-01-12
Dt 449

ffsiren:
Starts position watch early in page (outside of react) and then when react starts replaces the functions that retrieve position updates with react-aware functions.


app/action/position-actions.coffee
```diff
@@ -63,6 +37,13 @@ runReverseGeocodingAction = (actionContext, lat, lon, done) ->
 
 debouncedRunReverseGeocodingAction = debounce(runReverseGeocodingAction, 60000, {leading: true})
 
+broadcastCurrentLocation = (actionContext, pos) ->
+  if pos
```
samuliheljo: Could this method just be:
```if pos then actionContext.dispatch "GeolocationFound", pos```

File: static/geolocation.js [hunk 2]
```diff
@@ -0,0 +1,21 @@
...
+window.retrieveGeolocation = function retrievePos(pos) {
+  window.position.pos=pos;
+}
+
+window.retrieveError = function retrieveError(error) {
+  window.position.error = error;
+}
+
+if(window.location.search.indexOf('mock') === -1){
+  window.geoWatchId = navigator.geolocation.watchPosition(function geoPosition(position) {
```
samuliheljo: Are we sure that navigator.geolocation always exists?

PR #327, opened 2016-01-12
Test log session and commit

Macroz:
Check if more information can be added to tests to connect Snap and Browserstack


Macroz:
In Snap logs you can see e.g.
 bs-chrome   commit=69aad84
 bs-chrome   launch_url=default
 bs-chrome   session id=f60bf804f7dbb468b029fdc3da69e135c825d7ca

PR #332, opened 2016-01-13
serve Intl polyfill always (as gated)

ffsiren:


app/server.cjsx
```diff
@@ -59,10 +60,10 @@ getPolyfills = (userAgent) ->
       'fetch': flags: ['gated']
       'Promise': flags: ['gated']
       'String.prototype.repeat': flags: ['gated']
-      'Intl': flags: ['gated']
-      'Intl.~locale.en': flags: ['gated']
-      'Intl.~locale.fi': flags: ['gated']
-      'Intl.~locale.sv': flags: ['gated']
+      'Intl': flags: ['always', 'gated']
```
hannesj: Should 'always' be set, now that we added LG to the list of vendors not to trust?

PR #333, opened 2016-01-13
DT-493 SVG stop layer

hannesj:


hannesj:
Requires HSLdevcom/digitransit-deploy#73

app/component/map/stop/stop-marker-tile-layer.cjsx
```diff
@@ -0,0 +1,169 @@
...
+provideContext = require 'fluxible-addons-react/provideContext'
+intl          = require 'react-intl'
+
+class SVGTile
+  constructor: (@coords, done, @map) ->
+    @el = document.createElementNS 'http://www.w3.org/2000/svg', 'svg'
+    @el.setAttribute "class", "leaflet-tile"
+    @el.setAttribute "viewBox", "0 0 4096 4096"
+    @el.addEventListener "click", @onMapClick
+    if @coords.z < 14
```
samuliheljo: Could we check this before creating  SVGTile? I mean, is there a reason to create empty SVGTile if zoom level is low?

hannesj: Leaflet doesn't like if we don't return any element to be shown. Thus generate an empty SCG to show.

hannesj:
All issues found should be fixed now

PR #335, opened 2016-01-14
DT-580 do not update if location (from props) stays the same

ffsiren:


hannesj:
It still updates quite frequently when on mobile, as the location shifts slightly constantly on GPS.

ffsiren:
I believe that's because you have map tracking on? and the query is for the stops? that's separate from this issue.

PR #337, opened 2016-01-18
DT-297 Have a means to prevent local server from starting when running tests on BS

ffsiren:
The only required change is that an option to prevent starting of local server is added (test/script/run-ui-tests.sh).

This is because the docker pipeline uses the created docker container in tests,


ffsiren:
@hukka Can you please check that you're ok with the changes I did with the script. I basically just added a 4th parameter "noserver" that prevents the local server from starting.

After that we can merge this into master and start building new containers from master.

PR #349, opened 2016-01-28
Added location awareness in autosuggest/location searching.

seime:
Configurable and default false

### DIFF
--- a/app/config.default.coffee
+++ b/app/config.default.coffee
@@ -47,8 +47,8 @@ module.exports =
   autoSuggest:
     sortOrder: locality: 1, address: 2, stop: 3
     sortOthers: 4
-  # Let Pelias suggest based on current user location
-  locationAwareSuggestions: false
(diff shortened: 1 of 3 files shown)


app/config.default.coffee
```diff
@@ -47,6 +47,8 @@ module.exports =
   autoSuggest:
     sortOrder: locality: 1, address: 2, stop: 3
     sortOthers: 4
+  # Let Pelias suggest based on current user location
+  locationAwareSuggestions: false
```
hannesj: Our config is currently a bit messy with most things being directly on the root level. I would appreciate if you could move the option to be within autoSuggest in order to group options that are relevant together.

hannesj:
Hi,

Thanks for your pull request. It looks good except for the one comment I wrote. If you could fix that, I'm more than happy to merge this.

hannesj:
Merged, thanks!

PR #350, opened 2016-01-28
Airplane now an explicit choice of mode instead of always active

seime:


samuliheljo:
Thanks for the pull requests!

Some background first:
We are producing two journey planners:
1) National version
2) Helsinki city area version

So if I remember correctly, our situation differs from yours in this. 

Now, the problem with this pull request is that in Helsinki city area we don't want to show airplanes as option for routing. In national version we do. Actually, In Helsinki area we DO want to show city bikes, but DON'T want in national version.

In order to merge this, we would need transport modes to be configurable.

Configuration should include:
1) Whether particular mode (e.g. bus, tram, airplane) is available for selection
2) Default value (on/off) whether routing uses that mode or not. 

for Example, in HSL we would have:
- Bus: available, default on
- Tram: available, default on
- Train: available, default on
- Subway: available, default on
- Ferry: available, default on
- Citybike: available, default off
- Airplane: not available

Would you be interested in making this change?

csolem:
Hi. Can you please have another look at this pull request? Initial transport modes used for trip planning are now configurable. And it is also possible to configure if a transport mode toggle button is displayed or not. For instance, airplane toggle button is disabled for HSL.

samuliheljo:
Merged. Thanks a lot!

PR #357, opened 2016-02-04
Dt 626

siren:
implements DT-626, DT-627


VilleMiekkoja:
Some small improvements:

- Add top border to gray bar
- Add little more spacing

PR #377, opened 2016-02-15
fix: downgrade webpack, doesn't work

Macroz:


Macroz:
rm -rf node_modules; npm install
Now it works.

PR #385, opened 2016-02-18
Dt 667 add component

Macroz:
[DT-667](https://digitransit.atlassian.net/browse/DT-667)

Implement favourite routes with the same component as the nearby routes.

The query is very slow with a long list of routes so it must be optimized in the future.


Macroz:
Now using geo-utils distance to nearest stop

siren:
I know this is not strictly in the scope of this issue but i just noticed that the following resources could now be removed:
app/component/favourites/favourite-stop-card-list-container.cjsx
app/component/stop-cards/stop-card-list.cjsx
"geta6/masonry"

PR #393, opened 2016-02-22
NRP-126 Make the disruption info button configurable

csolem:


hannesj:
Merged, thanks @csolem

PR #395, opened 2016-02-22
Reposition map when previous leg is clicked

VilleMiekkoja:


hannesj:
Why have separate unFocus prop instead of giving it generic leafletEvents object?

PR #397, opened 2016-02-22
Wait leg component

csolem:
Separate component for wait legs. Will for instance fix the current issue with `NaN m`


samuliheljo:
Cheers!

PR #402, opened 2016-02-24
Rutebanken load scss from config name

csolem:
- Load sass theme in webpack config based on env variable name.
- Git ignore themes other than hsl and default
- Git ignore config other than hsl, client and default. 
- Git ignore svg-sprite.*.svg other than svg-sprite.hsl.svg and sv-spring.default.svg.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,16 @@ test/credentials.json
 test_output
 test/binaries
 selenium-debug.log
+sass/themes/*
+!sass/themes/default
+!sass/themes/hsl
+
+app/config.*.coffee
+!app/config.client.coffee
+!app/config.default.coffee
+!app/config.hsl.coffee
+
+
+static/svg-sprite.*.svg
+!static/svg-sprite.hsl.svg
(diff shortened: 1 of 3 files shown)


samuliheljo:
Hi, we'll dig into this tomorrow more closely.

We'll probably need to make slight changes because this approach requires us to define theme in the build phase. That causes us to build two different docker images, which is something that we don't want to do. We try to keep all these params as run options for docker containers.

Anyways, I'm sure we find some solution that works for us both.

I'll continue with this tomorrow...

csolem:
Hi. I see. What about leaving the hsl theme, and just configuring an additional theme? I have updated the pull request.

samuliheljo:
Yep, that was also my initial thought. I'll try how this new version works...

hannesj:
Also see webpack/webpack#370

csolem:
New version that just takes everything in sass/themes. Does not require env variable.

hannesj:
Thanks, looks good to me!

PR #407, opened 2016-02-25
Rutebanken configurable leg component mapping

csolem:
* Separate itinerary legs from itinerary tabs
* Attempt to make itinerary legs component configurable


csolem:
Closed because we will try to have only one leg mapping in ItineraryLegs.

PR #408, opened 2016-02-26
Configure available languages in config and use selected language for momentjs

csolem:
* Configure available languages in config
* Configure default language in config
* Set various language/locale values based on the selected language in PreferenceStore
* Configure moment in TopLevel with the selected locale
* Add configuration for moment thresholds used for humanizing times
* Load Intl locales based on the configured languages
* Configure webpack ContextReplacementPlugin with available languages


csolem:
Maybe there should be a mapping between selected language and what locale is used with moment. For instance, *en* could be mapped to *en-gb*. *no* could be mapped default to *nb*.

app/client.cjsx
```diff
@@ -36,8 +36,8 @@ Raven?.setUserContext piwik: piwik.getVisitorId()
 require('react-tap-event-plugin')()
 
 # English data added as default
-addLocaleData require "react-intl/lib/locale-data/fi"
-addLocaleData require "react-intl/lib/locale-data/sv"
+for language in config.availableLanguages
+  addLocaleData require "react-intl/lib/locale-data/" + language
```
hannesj: This will pull all existing languages into the js bundle, could we maybe add the current locale to the page on the server (from the selected locale), and fetch a new locale definition from the server when the user decides to switch?

app/config.hsl.coffee
```diff
@@ -36,6 +36,8 @@ module.exports =
     bucketSize: 100
   maxWalkDistance: 2500
   maxBikingDistance: 10000
+  availableLanguages: ['fi', 'sv', 'en', 'nb']
```
hannesj: nb could be removed

app/config.default.coffee
```diff
@@ -31,6 +31,8 @@ module.exports =
     bucketSize: 1000
   maxWalkDistance: 10000
   maxBikingDistance: 40000
+  availableLanguages: ['fi', 'sv', 'en', 'nb']
```
hannesj: nb could be removed

app/client.cjsx
hannesj: Actually, the server couuld just bundle in all the available locales and add them to the page, and we could load them here

app/util/configure-moment.js
```diff
@@ -0,0 +1,21 @@
+var config = require('../config')
+
+var moment = require('moment');
+
+// Configure moment with the selected language
+// and with the relative time thresholds used when humanizing times
+var configureMoment = function(language) {
+  moment.locale(language);
+  if(language != "en") {
+    require('moment/locale/' + language + '');
```
hannesj: Same goes here as to react-intl

webpack.config.js
```diff
@@ -36,16 +39,21 @@ function getLoadersConfig(env) {
 }
 
 function getPluginsConfig(env) {
+  if(!config.availableLanguages) {
+    throw "availableLanguages needs to be configured in the config file";
+  }
+  var languageExpression = new RegExp("^./(" + config.availableLanguages.join('|') + ")$");
```
hannesj: This requires that we specify a config file during build-time, which is something we don't want to do. Maybe we could loop through all config files and gather the files from there.

csolem: We could loop through the config files, but should we have to depend on configuration at all during build time? Maybe it's hard to avoid.

hannesj: I think that is ok, as it is possible to only include the config files present during build time.

webpack.config.js
```diff
@@ -35,17 +37,40 @@ function getLoadersConfig(env) {
...
+}
+
 function getPluginsConfig(env) {
+  var languageExpression = new RegExp("^./(" + getAllPossibleLanguages().join('|') + ")$");
+
   if (env === "development") {
     return([
       new webpack.HotModuleReplacementPlugin(),
-      new webpack.ContextReplacementPlugin(/moment(\/|\\)locale$/, /fi|sv|en\-gb/),
+      new webpack.ContextReplacementPlugin(/moment[\\\/]locale$/, languageExpression),
```
hannesj: Nice, now the same has to be done for `react-intl/lib/locale-data`, as it still tries to put evertyhing to the webpack context

app/html.cjsx
```diff
@@ -21,7 +21,7 @@ class Application extends React.Component
         <script dangerouslySetInnerHTML={ __html: @props.polyfill }/>
         <script dangerouslySetInnerHTML={ __html: @props.state }/>
         <script dangerouslySetInnerHTML={ __html: @props.config }/>
-        <script dangerouslySetInnerHTML={ __html: @props.locale }/>
+        <script dangerouslySetInnerHTML={ __html: @props.windowLocale }/>
```
hannesj: What about changeing this to `'window.locale="' + @props.locale + '"'` to save a prop.

PR #411, opened 2016-02-26
Make elements in the right menu configurable.

csolem:


app/component/summary/customize-search.cjsx
```diff
@@ -64,85 +64,95 @@ class CustomizeSearch extends React.Component
...
-          min={0.5}
-          max={3}
-          step={0.1}
-          minText={@context.intl.formatMessage(
-            {id: 'slow', defaultMessage: "Slow"})}
-          maxText={@context.intl.formatMessage(
-            {id: 'run', defaultMessage: "Run"})}
-        />
-      </section>
+      {if config.customizeSearch.walkReluctance.available
```
hannesj: It would probably be better to loop through these and conditionally add to an array, or what do you think?

csolem: Maybe. If we had more of the variables configured, we could loop through the configured sliders and create them dynamically. For now I think this is viable.

csolem: Please let me know if I should look further into looping through and adding to an array, or if the current solution is viable.

PR #416, opened 2016-03-01
Display destination name together with alight text. Hide alight text if airplane

csolem:
* Show the destination name together with the alight text, not on the next line.
* If transit mode is AIRPLANE, do not show the alight text

### DIFF
--- a/app/component/itinerary/transit-leg.cjsx
+++ b/app/component/itinerary/transit-leg.cjsx
@@ -80,11 +80,15 @@ class TransitLeg extends React.Component
               =1 {one minute}
               other {# minutes}})' />
         </div>
-        <div><FormattedMessage
-          id='alight'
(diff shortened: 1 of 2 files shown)


hannesj:
Just talked with our graphical designer, and we thought of if we really need to show this information at all, as it is shown already on the next row. We really haven't really thought of the exact texts to be shown on the instructions and are open to suggestions if you have feelings about this.

csolem:
Actually, the alight text is removed here: https://github.com/HSLdevcom/digitransit-ui/pull/422/files#diff-d13f74769ef3add9e8d9307ca135d41bL83

hannesj:
What I meant was for all transport types, not just for airplanes

csolem:
In #422 it will be removed for all transport types, because the whole if statement is removed.

PR #425, opened 2016-03-02
Update and refactor styles, DT-713 and DT-681

hukka:
- Change HSL typeface
- Remove all font related hardcoding from SCSS files
- Always specify font-weight when specifying font-family, and vice versa
- Update styleguidelines
- Refactor Foundation override out of themes and component styles


app/html.cjsx
```diff
@@ -12,6 +12,7 @@ class Application extends React.Component
       {head?.meta.toComponent()}
       {head?.link.toComponent()}
       <style dangerouslySetInnerHTML={ __html: @props.fonts }/>
+      <link rel="stylesheet" type="text/css" href="https://cloud.typography.com/6364294/6653152/css/fonts.css" />
```
hannesj: This should probably be added in the config file, where tje current font CSS comes from

PR #428, opened 2016-03-03
Enable city bike available bike count 

csolem:


tuukka:
Could you add a config option for display 0 as the count? We would still need this for a month or so at HSL.

samuliheljo:
Cheers!

samuliheljo:
I made small tweaks:
- we want to show citybikes for both of our installations (makes whole config option now obsolete?)
- we want to use location aware pelias search

PR #429, opened 2016-03-04
Dt 691 disable horizontal swipe

VilleMiekkoja:


app/page/itinerary.cjsx
```diff
@@ -34,11 +98,64 @@ class ItineraryPage extends React.Component
...
+            index={parseInt(@props.params.hash)}
+            className="itinerary-swipe-views-root"
+            slideStyle={{height: "100%"}}
+            containerStyle={{height: "100%"}}
+            onChangeIndex={@switchSlide}>
+            {@getSlides(itineraries)}
+          </SwipeableViews>
+          <div className="itinerary-tabs-container">
+            <Tabs
+              onChange={() => return}
```
VilleMiekkoja: This prop function was required in order to the Tabs component to update itself when slide occurs in desktop display mode. In real mobile device, the tab dots updated correctly even without this, but on desktop google chrome display mode, they did not unless this prop function. In material-ui example, there was also two change listeners, one in tabs and one in the react swipeable views: http://www.material-ui.com/#/components/tabs (swipeable example)

app/component/itinerary/itinerary-tabs.cjsx
```diff
@@ -7,98 +6,65 @@ EndLeg             = require './end-leg'
...
-      "lat": lat
-      "lon": lon
+      lat: lat
+      lon: lon
 
   render: ->
     numberOfLegs = @props.itinerary.legs.length
-    leafletObj = <ItineraryLine key="line" legs={@props.itinerary.legs} showFromToMarkers={true} showTransferLabels={true}/>
+    legs = []
+    @props.itinerary.legs.forEach (leg, j) =>
```
samuliheljo: Use ItineraryLegs

PR #433, opened 2016-03-06
Fix too frequent appearance of the nps query.

VilleMiekkoja:


tuukka:
:tada:

hannesj:
Should we use the existing time from time-store? It can be accessed from `client.cjsx` with `context.getComponentContext().getStore('TimeStore').getCurrentTime()`, which returns a Moment object.

PR #445, opened 2016-03-14
Dt 802 refactor transit leg

VilleMiekkoja:


samuliheljo:
review comment 1:
Check texts; now different languages have different bundle keys available:
e.g.:
- tram-with-route-number

You can try to run ./language-check.sh (osx, linux) or I can run it for you

PR #452, opened 2016-03-19
Add French translation

kalon33:
Add french translation for digitransit-ui interface.


kalon33:
Close as some other commit went into this.

PR #460, opened 2016-03-23
DT-836 Measure frame jank in map and nearby routes

hukka:
DT-836

Windows phone support is pretty bad with map tracking; the events just don't seem to fire reliably and even then the requestAnimationFrame seems to fire only once. Debugging this should wait for Win10 dev laptop.


samuliheljo:
I tested this with "Test 3" Windows phone. Basically, it makes map unpannable and in addition it does not send any perf data to server on windows phone.

In addition, I am not sure if we should enable this kind of tracking on every device since we know that windows phones are the problem and we can verify this using project phones.

Therefore, suggested changes:
- Make it work on windows phone
- Put this feature behind some kind of toggle (url param maybe?)

samuliheljo:
Ok, so windows phone has some problems. Let's put this on and see what happens.

PR #468, opened 2016-03-30
Show spinner if itinerary rows empty

csolem:
Show a spinner while waiting for itinerary search result. To let the user know that the search is ongoing.


hannesj:
This was implemented in #470

PR #470, opened 2016-03-31
Dt 451 use graphql in itinerary

Macroz:
GraphQL based planning works mostly the same as the previous REST planning, but with some missing features and improvements needed. I think it would be useful to add these as separate tasks, if this PR is deemed good enough for now.

- More old code for itinerary search action can be removed
- Prop and context types that are required should be checked.
- Requires should be checked
- WAIT legs etc. special legs that were added previously in the itinerary search action are not added at the moment
- Intermediate stops are not displayed the map anymore. This needs to be refactored in OTP as well. In the future they are supposed to be shown on the new itinerary page map, not on the summary map. ShowIntermediateStops is not required after this.
- Route not found errors should be tested more.
- Summary components can be refactored to basic presentation components as well as Relay wrappers. This is something that needs to be checked together when merging the new summary page changes from Norway.


Macroz:
This PR requires also that OTP PR is merged https://github.com/HSLdevcom/OpenTripPlanner/pull/48

Macroz:
**Navigating from summary to itinerary does not work at the moment.** Need to implement that part.

Macroz:
This is more or less complete now. Remaining tasks to cleanup more and introduce Wait & Airport legs in a new way.

samuliheljo:
It seems that UI send multiple GraphQL queries to server (3) when route is tried from current location to set destination

I have set piwik id and endpoint correctly

app/page/itinerary.cjsx
```diff
@@ -19,78 +14,57 @@ class ItineraryPage extends React.Component
...
+    plan = <Relay.RootContainer
+      Component={ItineraryPlanContainer}
+      route={new queries.SummaryPlanContainerRoute(
+        fromPlace: @props.params.from
+        toPlace: @props.params.to
+        numItineraries: 3
+        modes: modes
+        date: selectedTime.format("YYYY-MM-DD")
+        time: selectedTime.format("HH:mm:ss")
+        walkReluctance: walkReluctance + 0.000099
```
samuliheljo: is this "stuff that will be removed once some upstream pull request is merged"?

samuliheljo:
This does not work, but just keeps loading

http://dev.digitransit.fi/reitti/Raikan%20Rauta%2C%20Helsinki%3A%3A60.197131%2C24.882565/Rovaniemi%2C%20Rovaniemi%3A%3A66.498048%2C25.705348

Time 07:19 ish

samuliheljo:
Just as a general comment for 3rd parties. Itinerary
- wait "legs"
- airplane checking
- airplane checkout

Will be implemented outside this pr

app/page/itinerary.cjsx
Macroz: Yes, the + 0.000099 is a big in graphql-java and will be removed once we can use the patched version. Patch has been merged already.

Macroz:
Found a case where the shortName length is checked. It can sometimes be missing. This fixes the problem above.

PR #478, opened 2016-04-05
Configure if the search button should be rendered

csolem:
* Configure if the search button should be rendered
* Use border radius on the lower search field in case there is no search button


Macroz:
We are right now merging a bunch of our PRs from the sprint, including one search field changes. Let's have a look at including this next week.

csolem:
Closed for now, because the search fields were changed.

PR #480, opened 2016-04-06
Dt 809 support for orig and dest

VilleMiekkoja:
This pull request does not include the removal of the second search field. That will be done in DT-808


app/translations.coffee
```diff
@@ -14,12 +14,13 @@ translations =
...
     'no-position': 'Ei sijaintia'
-    'origin': 'Mistä? - osoite tai pysäkki'
+    'origin-placeholder': 'Mistä? - osoite tai pysäkki'
+    'origin': 'LÄHTÖPAIKKA'
     'to-stop': 'Pysäkille'
     'line': 'Linja'
     'leaves': 'Lähtee'
     'next': 'Seuraava'
-    'destination': 'Määränpää'
+    'destination': 'MÄÄRÄNPÄÄ'
```
samuliheljo: These should be made uppercase using css style text-transform
- origin
- destination
- etc

app/component/search/search.scss
```diff
@@ -323,7 +299,25 @@ div#search-panel-close-tab {
...
+    position: absolute;
+    color: $gray-blue;
+  }
+
+  .search-header__button {
+    margin-bottom: 0;
+    letter-spacing: -0.025em;
+  }
+
 }
```
samuliheljo: Tab bar is missing "blue top border" for selected as in design: https://projects.invisionapp.com/share/5X5RNXGGU#/screens/130683408

app/component/search/search-modal.cjsx
```diff
@@ -36,40 +25,20 @@ class SearchModal extends React.Component
...
-          <Icon img={'icon-icon_arrow-left'} />
-          <span className="search-header-separator"/>
+        <div className="small-12 medium-6 medium-offset-3 columns cursor-pointer search-header">
+          <span className="search-header__back-arrow" onClick={@props.closeModal}>
+            <Icon img={'icon-icon_arrow-left'}/>
+            <span className="search-header-separator"/>
+          </span>
+          <Tabs
+            inkBarStyle={{display: "none"}}
+            tabItemContainerStyle={{backgroundColor: "#eef1f3", lineHeight: "18px", marginLeft: "28px", width: "calc(100% - 28px)"}}
```
samuliheljo: Could we do this using classes?

app/component/search/search-modal.cjsx
```diff
@@ -13,22 +15,9 @@ class SearchModal extends React.Component
...
-    position = @context.getStore('SearchStore').getPosition()
-    @setState("value": position?.address || "")
-
-  closeModal: () =>
-    @context.executeAction SearchActions.closeSearch
-
   render: =>
     style = {}
-    if @context.getStore('SearchStore').isModalOpen() == false
+    if @props.modalIsOpen == false
```
samuliheljo: There might be reason why this is done like this. However, I'm wondering why are we rendering the modal in the first place when it is closed.

Why not render it only when "modalIsOpen" ?

app/component/favourites/add-favourite-container.cjsx
```diff
@@ -144,7 +151,31 @@ class AddFavouriteContainer extends React.Component
...
+        ref="modal"
+        modalIsOpen={@state.searchModalIsOpen}
+        selectedTab={"favourite-place"}
+        closeModal={@closeSearchModal}>
+        <Tab
+          className="search-header__button"
+          label={searchTabLabel}
+          ref="searchTab"
+          value={"favourite-place"}
+          style={{
```
samuliheljo: I would rather see these styles coming from css classes

app/component/search/search-input.cjsx
```diff
@@ -106,14 +104,11 @@ class SearchInput extends React.Component
...
       onSuggestionSelected={@currentItemSelected}
       focusedItemIndex={@state.focusedItemIndex}
       inputProps={
-        "id": "autosuggest-input"
-        "value": @state?.value || ""
+        "id": @props.id
+        "value": if @state.value?.length >= 0 then @state?.value else @props.initialValue
         "onChange": @handleUpdateInputNow
         "onKeyDown": @handleOnKeyDown
         "placeholder": @context.getStore('SearchStore').getPlaceholder()
```
samuliheljo: What's this? store contains "undefined"

app/component/search/search-field.cjsx
```diff
@@ -1,6 +1,5 @@
 React           = require 'react'
```
samuliheljo: Was it so that "dummy search field" will be made simpler in following tasks?

app/component/search/search.scss
VilleMiekkoja: That's true, for some reason I didn't see it in design. Perhaps the blue top bar confused me

app/component/favourites/add-favourite-container.cjsx
```diff
@@ -68,6 +71,13 @@ class AddFavouriteContainer extends React.Component
       'icon-icon_shopping'
     ]
 
+  focusInput: () =>
+    @refs["searchInputfavourite"]?.refs.autowhatever?.refs.input?.focus()
```
hannesj: Why not use dot notation for searchInputFavourite?

app/component/search/search-two-fields-container.cjsx
```diff
@@ -120,22 +147,79 @@ class SearchTwoFieldsContainer extends React.Component
...
+        <Tab
+          className="search-header__button"
+          label={originSearchTabLabel}
+          ref="searchTab"
+          value={"origin"}
+          onActive={@onTabChange}
+          style={{
+            color: if @state.selectedTab == "origin" then "#333" else "#7f929c",
+            fontSize: "11px",
+            fontFamily: "Gotham Rounded SSm A, Gotham Rounded SSm B, Arial, Georgia, Serif",
```
hannesj: These should come from the CSS, as they are theme-specific

app/component/search/search-two-fields-container.cjsx
```diff
@@ -46,12 +53,29 @@ class SearchTwoFieldsContainer extends React.Component
...
 
+  onTabChange: (tab) =>
+    @setState
+      selectedTab: tab.props.value
+      () =>
+        if tab.props.value == "origin"
+          @context.executeAction SearchActions.executeSearch, @context.getStore('EndpointStore').getOrigin()?.address || ""
+        if tab.props.value == "destination"
+          @context.executeAction SearchActions.executeSearch, @context.getStore('EndpointStore').getDestination()?.address || ""
+    setTimeout((() => @focusInput(tab.props.value)), 0) #try to focus, does not work on ios
```
hannesj: Why not do it in the callback of the setState?

app/component/search/search-modal.cjsx
VilleMiekkoja: Only override is provided by material-ui. They do not provide className for that. We could find the div and that way override line-height and margin-left. But the others we could not override without !Important.

app/component/search/search-two-fields-container.cjsx
VilleMiekkoja: Yes, they should. But I didn't find a way to do it without !important. And as we discussed about this earlier, we decided we would then use inline styles. Though now I discussed it with Samuli, and we agreed that we could use !important this time.

app/component/search/search-field.cjsx
VilleMiekkoja: Yes, that was the goal

app/component/search/search-input.cjsx
VilleMiekkoja: Good question! According to design, there is no placeholder in the search. So that is superfluous and can be deleted

app/component/favourites/add-favourite-container.cjsx
```diff
@@ -76,8 +86,10 @@ class AddFavouriteContainer extends React.Component
       id: 'address'
       defaultMessage: 'Address')
 
-    focusInput = () =>
-      @refs.modal?.refs.searchInput?.refs.autowhatever?.refs.input?.focus()
+    searchTabLabel = @context.intl.formatMessage(
+      id: 'favourite-target'
+      defaultMessage: 'FAVOURITE PLACE'
```
hannesj: Should not be capitalized

app/component/search/search-modal.cjsx
```diff
@@ -6,70 +6,38 @@ FormattedMessage = intl.FormattedMessage
...
-    else
-      style.right = "0px"
-
-    <div style={style}
-      className="search-modal">
-      <div className="row">
-        <div className="small-12 medium-6 medium-offset-3 columns cursor-pointer search-header" onClick={@closeModal}>
-          <Icon img={'icon-icon_arrow-left'} />
-          <span className="search-header-separator"/>
+    if @props.modalIsOpen
```
hannesj: Just a nitpick, but it is usually nice to exit immediately rather than in else, i.e. transforming this to

```
if !@props.modalIsOpen
  return false
<div ...
```

app/component/search/search-two-fields-container.cjsx
```diff
@@ -9,9 +9,16 @@ SearchField      = require './search-field'
 intl             = require 'react-intl'
 FormattedMessage = intl.FormattedMessage
 SearchModal      = require './search-modal'
+SearchInput      = require './search-input'
+Tab              = require 'material-ui/lib/tabs/tab'
 
 class SearchTwoFieldsContainer extends React.Component
```
hannesj: Should this be renamed to something else, like SearchTabs?

app/component/search/search-two-fields-container.cjsx
```diff
@@ -86,32 +110,36 @@ class SearchTwoFieldsContainer extends React.Component
     origin = @context.getStore('EndpointStore').getOrigin()
     destination = @context.getStore('EndpointStore').getDestination()
 
-    focusInput = () =>
-      @refs.modal?.refs.searchInput?.refs.autowhatever?.refs.input?.focus()
+    originSearchTabLabel = @context.intl.formatMessage(
+      id: 'origin'
+      defaultMessage: 'ORIGIN'
```
hannesj: Capitalized, same with 'DESTINATION'

app/component/search/search-two-fields-container.cjsx
```diff
@@ -86,32 +110,36 @@ class SearchTwoFieldsContainer extends React.Component
     origin = @context.getStore('EndpointStore').getOrigin()
     destination = @context.getStore('EndpointStore').getDestination()
 
-    focusInput = () =>
-      @refs.modal?.refs.searchInput?.refs.autowhatever?.refs.input?.focus()
+    originSearchTabLabel = @context.intl.formatMessage(
```
hannesj: No need for parenthesis, same with destination?

app/translations.coffee
```diff
@@ -14,12 +14,13 @@ translations =
...
     'no-position': 'Ei sijaintia'
-    'origin': 'Mistä? - osoite tai pysäkki'
+    'origin-placeholder': 'Mistä? - osoite tai pysäkki'
+    'origin': 'lähtöpaikka'
     'to-stop': 'Pysäkille'
     'line': 'Linja'
     'leaves': 'Lähtee'
     'next': 'Seuraava'
-    'destination': 'Määränpää'
+    'destination': 'määränpää'
```
hannesj: The first letter can still be capitalized, so that it looks ok in other places, same goes with destination, etc.

PR #484, opened 2016-04-06
Green departure time if realtime

csolem:
When real time is available for the itinerary, make the departure time green. This is a temporary solution until the full design is implemented.(with two departure times).

### DIFF
--- a/app/component/summary/summary-row.cjsx
+++ b/app/component/summary/summary-row.cjsx
@@ -24,8 +24,11 @@ class SummaryRow extends React.Component
       if data.legs[i].transitLeg
         noTransitLegs = false
 
+    realTimeAvailable = false
+
     for leg, i in data.legs
       if leg.transitLeg or noTransitLegs
+        if leg.realTime then realTimeAvailable = true
         legs.push <RouteNumber
                     mode={leg.mode}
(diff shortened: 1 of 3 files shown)


app/component/summary/summary-row.scss
```diff
@@ -52,6 +52,10 @@
     margin-top: 4px;
   }
 
+  & .realtime-available {
+    color: #00985f;
```
hannesj: This should probably come from the sass variable configured in the theme.

app/component/summary/summary-row.cjsx
```diff
@@ -48,7 +51,7 @@ class SummaryRow extends React.Component
           {geoUtils.displayDistance(data.walkDistance)}
         </div>
       </div>
-      <div className="itinerary-start-time">
+      <div className={cx "itinerary-start-time", if realTimeAvailable then "realtime-available"}>
```
hannesj: You can just use `{cx "itinerary-start-time", "realtime-available": realTimeAvailable}` to simplify a bit

hannesj:
Also, should we show this only if the first vehicle has realtime information or any of them have it?

csolem:
Would the start time from OTP be affected if one of the vehicles has realtime information, and is delayed?

PR #486, opened 2016-04-07
Remove arrow link from summary page

hannesj:


app/component/util/util.scss
```diff
@@ -246,19 +246,6 @@ $btn-heigth: 17px;
...
 }
 
-/*
-  ARROW LINK
-*/
-
-.icon.right-arrow-blue-background {
-  padding: 13px;
-  height: $arrow-link-size;
-  width: $arrow-link-size;
```
VilleMiekkoja: The variable should be deleted as well

PR #487, opened 2016-04-07
Use full height when showing nearby/favourites and add exit cross 

csolem:
* Add heading and exit cross on nearby routes panel and favourites panel
* Use full height for panels for nearby routes and favourites
* Move panel for nearby routes to a separate component


samuliheljo:
Yes, you should keep on working with this :+1: 

Some more detailed designs for you:
hsl_reittiopas_ios_etusivu_lahellasi_02_720
hsl_reittiopas_ios_etusivu_suosikit_02_720

So we added that "X" to close tab also.

I'll check that "front page loads twice" problem...

csolem:
Maybe it's better to store the state of the front page panel in a store, similar to the feedback-store?

samuliheljo:
I havent't had time yet to check that state part. Couple quick insights:
- Back button should close tab as it does now in master version. This is now handled with browser state.
- My initial feel is that removing when tab is opened will not produce any performance improvements and it even might make things slower (it takes time to put it back once tab closes)

Anyways, plan remains the same; let's make that tab as designed in above images. And as you were interested in making tabs take less space, put that as configuration option.

samuliheljo:
You should check how location.state is used here:
https://github.com/HSLdevcom/digitransit-ui/blob/master/app/component/front-page/front-page-panel.cjsx#L52

samuliheljo:
So to sum it up, you have basically two options:
a) wait a bit and let us check whether this "do not render map when x" is actully good way to proceed
b) try to do something for state.

I think that it could be better to go with a) as work with b) might not be needed in the end.

csolem:
I have the same opinion as you. To remove the map and rerender it will probably not gain performance. So I will remove this logic for now.

csolem:
Also, I think the X will do fine for us as well. Because of the X, we can also have full height for panels. Because the original thought was that it should be easy to close favourites/nearby routes. Which means less configuration.

csolem:
Can you please have another look at this now, Samuli?

samuliheljo:
Brilliant! Looks good, Merging.

Extra points for panel refactoring :)

PR #492, opened 2016-04-11
Configurable items in the left menu

csolem:
Make it configurable to render some of the elements in the left menu (off canvas menu). This makes it possible to only render the language selector,


samuliheljo:
In future this left menu will be moved to right side and it'll be available on every page.
I except that to happen in one month.

When that change is made, I would always keep that menu available but have better support on what links are shown in the menu.

I'm merging this for now as temporary solution

PR #493, opened 2016-04-11
Configurable card header description for stop place card

csolem:
Making it possible to not show the stop card header description in stop cards.

### DIFF
--- a/app/component/stop-cards/stop-card-header.cjsx
+++ b/app/component/stop-cards/stop-card-header.cjsx
@@ -19,13 +19,12 @@ class StopCardHeader extends React.Component
 
   getDescription: ->
     description = ""
-    if config.stopCard.showCardHeaderDescription
-      if @props.stop.desc
-        description += @props.stop.desc + " // "
-      if @props.stop.code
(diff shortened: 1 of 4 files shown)


samuliheljo:
Could you elaborate little bit, why not show stop info?

csolem:
The reasons for this are

1. Our values for `@props.stop.code` is not always a public code that is interesting for the user of the application
2. Our current gtfs data is somewhat not yet consistent enough
3. If one of the fields are empty, we end up with trailing `//`

kalon33:
To add my two cents: In GTFS data I use in France, stops numbers or ids are not known from the public (only from people who have to deal with data). Only stop names are, so in this situation it would be better to hide them.

csolem:
Maybe we could make each one of them (desc/code/distance) configurable, and do something about the slashes?

kalon33:
That seems great to me, but I'm not the one who decides ;)

samuliheljo:
Based on discussion above, I would make all 3 configurable. Once those modifications are here, I shall merge.

samuliheljo:
Thanks!

PR #497, opened 2016-04-12
Dt 808 one search field

VilleMiekkoja:


test/search.js
```diff
@@ -1,45 +0,0 @@
-var suite = require('./api/suite.js').suite;
```
hannesj: Why remove tests entirely instead of adjusting them?

app/component/search/fake-search-with-button.cjsx
```diff
@@ -0,0 +1,30 @@
+React       = require 'react'
+Icon        = require '../icon/icon'
+
+FakeSearchWithButton = (props) ->
+  <div className="search-form">
```
hannesj: Could we simplify the dom structure?

VilleMiekkoja: Managed to simplify it little

test/api/browser-upgrade.js
```diff
@@ -72,32 +72,53 @@ module.exports = function (browser) {
...
       browser.click('#destination');
     },
 
-    enterText: browser.origin.enterText,
-
+    enterText: function type(text) {
+      browser.expect.element('#search-destination').to.be.enabled.before(ELEMENT_VISIBLE_TIMEOUT);
+      browser.pause(100);
+      browser.setValue('#search-destination', text);
+      browser.pause(1000); //wait for suggestions
```
hannesj: Couldn't this be done with something like `browser.expect.element('#suggestions').to.be.visible.before(ELEMENT_VISIBLE_TIMEOUT)`

app/component/search/search-main-container.cjsx
```diff
@@ -191,6 +173,7 @@ class SearchTwoFieldsContainer extends React.Component
           className={"search-header__button" + if @state.selectedTab == "destination" then "--selected" else ""}
           label={destinationSearchTabLabel}
           value={"destination"}
+          id={"destination"}
```
hannesj: Parenthesis not needed around strings

PR #498, opened 2016-04-12
DT-842 Implement stops and city bikes using tiled canvas layer

hannesj:


kalon33:
Will other bike systems (provided via OTP) will still show up on the map (and be usable) with this PR?

hannesj:
@kalon33 they should be showing all ready with the current implementation. The new implementation is meant for improved performance on mobile devices.

kalon33:
@hannesj OK so the new implementation will keep them and make them faster? I feared that it would replace existing one by something HSL specific. Thanks for replying :)

samuliheljo:
Map scrolling is now considerably better on windows phone, great!
Noticed, that citybike popup:
- Has 0/NaN
- Does not include "bar that shows how many available"

screen shot 2016-04-14 at 14 44 52

app/component/map/tile-layer/select-stop-row.cjsx
```diff
@@ -0,0 +1,54 @@
...
+  patterns.push(
+    <div key="first" className="route-detail-text">
+      <span className={patternData[0].type.toLowerCase() + " vehicle-number no-padding"}>{patternData[0].shortName}</span> {#Repalce with RouteNumber}
+      <RouteDestination mode={patternData[0].type} destination={patternData[0].headsign} />
+    </div>)
+
+  if patternData.length > 1
+    patterns.push(
+      <div key="second" className="route-detail-text">
+        Lisäksi{patternData[1..].map((p) -> if p.shortName then <span key={p.shortName} style={padding: "0 2px"} className={p.type.toLowerCase() + " vehicle-number"}>{p.shortName}</span> else null)}
```
samuliheljo: Missing translation

app/component/map/tile-layer/marker-select-popup.cjsx
```diff
@@ -0,0 +1,19 @@
...
+class MarkerSelectPopup extends React.Component
+  render: ->
+    options = @props.options.map (option) ->
+      if option.layer == "stop"
+        <SelectStopRow key={option.feature.properties.gtfsId} {...option.feature.properties}/>
+      else if option.layer == "citybike"
+        <SelectCitybikeRow key={option.feature.properties.stationId} {...option.feature.properties}/>
+
+    <div className="card">
+      <h3 className="padding-normal">Valitse pysäkki</h3>
```
samuliheljo: Translation

PR #500, opened 2016-04-12
Dt 697 refactor routing

siren:


app/history.coffee
```diff
@@ -0,0 +1,8 @@
+useRouterHistory     = require 'react-router'
```
hannesj: The first one is probably a typo?

PR #512, opened 2016-04-15
Make it easier to close search popup

samuliheljo:
Click on div instead of span


app/component/search/search-modal.cjsx
```diff
@@ -24,10 +24,10 @@ class SearchModal extends React.Component
         className="search-modal">
         <div className="row fullscreen">
           <div className="small-12 medium-6 medium-offset-3 columns cursor-pointer search-header">
-            <span className="search-header__back-arrow" onClick={@props.closeModal}>
-              <Icon img={'icon-icon_arrow-left'}/>
+            <div className="search-header__back-arrow" onClick={@props.closeModal}>
```
hannesj: Add `cursor-pointer`?

samuliheljo: It's actually for whole row already. See above.

hannesj: Hups

PR #515, opened 2016-04-17
DT-896 Update dependencies

hannesj:
(And fix lint issues)


app/page/itinerary.cjsx
```diff
@@ -98,7 +98,7 @@ class ItineraryPage extends React.Component
           walkBoardCost: search.walkBoardCost
           minTransferTime: search.minTransferTime
           walkSpeed: search.walkSpeed + 0.000099
-          maxWalkDistance: search.maxWalkDistance
+          maxWalkDistance: search.maxWalkDistance + 0.1
```
hukka: Why the 0.1? A comment would be good

kalon33:
Hi @hannesj , following yesterday changes, I got #529 bug in digitransit-ui. Is there any chance that it may be linked to your commits? I tested right after #527 merging and that was still working :(

PR #518, opened 2016-04-19
Dt 814 show origin popup

siren:


app/component/map/map.scss
```diff
@@ -232,6 +232,25 @@ div.leaflet-marker-icon.vehicle-icon {
     }
   }
 }
+
+div.origin-popup {
```
samuliheljo: cursor: pointer; would be cool

samuliheljo:
Let's make margin slightly smaller on .leaflet-popup-content

PR #519, opened 2016-04-19
Dt 844 clear button

siren:


app/component/search/search.scss
```diff
@@ -229,3 +254,14 @@ div.search-modal {
   position: relative;
   bottom: -4px;
 }
+
+.geolocation-selected {
```
samuliheljo: This should not be as bold as it is now

app/component/search/search.scss
```diff
@@ -132,11 +132,13 @@ ul#react-whatever-suggest {
...
+  margin-right: 10px;
 }
 
-.address-box .geolocation-bar-location-icon svg.icon {
+.geolocation-bar-location-icon svg.icon {
   fill: $hilight-color;
+  padding-top: 2px;
 }
 
 div.search-modal {
```
samuliheljo: We should refactor this css file so that we have:

.search-form {
...
}

.search-modal {
...
}

And we would then use SASS nesting for those elements. Now we have 
.geolocation
id:clear-input
.geolocation-selected

etc. and it's not clear how these are nested.

samuliheljo:
height:
screen shot 2016-04-20 at 13 58 26

samuliheljo:
iPhone comments:
- on iphone "my position" icon is too positioned "too high"
- on iphone "X" is hard to click. It seems that it very easily takes click to whole input.

Functionality:
- When input is empty, "X" should not be rendered

Bugs:
- seach does not work with white windows test phone ("test3"), keyboard is not shown


Css:
- Add some padding to "my position" bar. Now there isn't enough white space around bar
- Make "X" icon slightly smaller

PR #521, opened 2016-04-19
Dt 915 open citybike popup

VilleMiekkoja:


hannesj:
Why not just set the statest.stops to contain the single bikeshare station (and possibly renamestate.stops to something more relevant)?

PR #528, opened 2016-04-26
Dt 911 enable citybike icons

VilleMiekkoja:


app/component/map/tile-layer/tile-container.coffee
```diff
@@ -9,10 +10,18 @@ class Tile
     @eventratio = @ratio * @scaleratio
     @el = @createElement()
 
-    if @coords.z < 14 or !@el.getContext
+    if @coords.z < config.markersMinZoom or !@el.getContext
```
hukka: Remove markersMinZoom and check both other configs here, or just remove this whole if?

PR #530, opened 2016-04-27
DT-938 Fix stops & citybikes

hannesj:


app/component/map/tile-layer/city-bikes.coffee
```diff
@@ -45,5 +44,6 @@ class CityBikes
       (geom[0][0].x / @tile.ratio) - citybikeImageSize / 2,
       (geom[0][0].y / @tile.ratio) - citybikeImageSize / 2
 
+CityBikes.getName = () -> "citybike"
```
hukka: @hannesj: @getName: () -> ?

PR #531, opened 2016-04-27
DT-943 Use specific css class for origin popup

hukka:


app/component/map/map.scss
```diff
@@ -234,6 +234,10 @@ div.leaflet-marker-icon.vehicle-icon {
 }
 
 div.origin-popup {
+  @include font-bold;
+  color: $gray-blue-light;
+  font-size: 1qpx;
```
hannesj: qpx?

PR #533, opened 2016-04-27
DT-940 fix search alignment

hukka:


hukka:
Based on top of #531 and #532

PR #534, opened 2016-04-27
DT-955 summary to itinerary

### LINKED ISSUE #529: [regression] can't retrieve itinerary details in digitransit-ui
With latest Tuesday 26th April changes, it is not possible to display itinerary details anymore (was possible just before these commits on 25th evening, I tested it). Spinner is just turning again and again... Setting origin and destination and retrieving itinerary list is possible, but not displaying the detailed view.

Is this possible to get a look at it? I reviewed my diff between master and my local repo, and I only have changes in config.default.coffee and in app/util/leg-text-util.js (to disable line name replacement by "M" for subway lines).

Thanks!

hukka:
Fix #529


hannesj:
:+1: :clap:

PR #536, opened 2016-04-28
DT-910 Set citybikes on frontpage if URL has ?citybikes

hukka:


app/page/index.cjsx
```diff
@@ -8,12 +8,17 @@ Icon               = require '../component/icon/icon'
...
 FeedbackPanel      = require '../component/feedback/feedback-panel'
+ItinerarySearchAction = require '../action/itinerary-search-action'
 
 class Page extends React.Component
   @contextTypes:
     getStore: React.PropTypes.func.isRequired
     executeAction: React.PropTypes.func.isRequired
 
+  componentDidMount: ->
+    if window.location.search?.indexOf('citybikes') > -1
```
samuliheljo: use this.context.location

samuliheljo: also, this should be moved to TopLevel -component, so that it is not page specific

samuliheljo: also, how about refactoring ?mock also

PR #557, opened 2016-05-03
Fixes application on lumia 625

samuliheljo:


hannesj:
Fixed already in #550

PR #570, opened 2016-05-09
Dt 920 show first vehicle

Macroz:
Implements [DT-920](https://digitransit.atlassian.net/browse/DT-920)

Should test with mobiles and various route combinations.

### DIFF
--- a/app/component/documentation/example-data.coffee
+++ b/app/component/documentation/example-data.coffee
@@ -45,9 +45,16 @@ station =
 favouriteLocation =
   locationName: "Työ"
   favouriteLocationIconId: "icon-icon_place"
-  arrivalTime: "14:38"
-  departureTime: "3 min"
+  arrivalTime: 1444175199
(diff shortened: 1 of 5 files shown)


samuliheljo:
Overflow:

<img width="139" alt="screen shot 2016-05-10 at 12 18 13" src="https://cloud.githubusercontent.com/assets/1298522/15141588/6e084222-16a9-11e6-83e9-7de033268c07.png

samuliheljo:
screen shot 2016-05-10 at 12 18 13

samuliheljo:
Why is it not showing "first itinerary" but in some case e.g. second?

Macroz:
I'm going over the different mobiles to fit the text.

First itinerary is not shown because we take the first (i.e. "best") route from OTP. In the summary screen we show three "best" routes in the order of departure time (i.e. not in the "best first" order). So the route will be one of the three but not necessarily the first.

PR #574, opened 2016-05-10
Disable zoom snapping

hannesj:


tuukka:
Do we really want this? (Raster tiles become fuzzy.)

samuliheljo:
I think it needs some work with lumia phones

PR #580, opened 2016-05-12
DT-1014 Fallback to default endpoint to avoid undefined center in map

csolem:
When running with `map.useVectorTiles = false`, using digitransit with `?mock` in the URL fails because the leaflet center is not defined.


samuliheljo:
Jira ticket: https://digitransit.atlassian.net/browse/DT-1014

csolem:
#581

PR #584, opened 2016-05-15
Refactor departure list

hannesj:


app/component/favourites/FavouriteRouteListContainer.js
```diff
@@ -0,0 +1,115 @@
...
+
+FavouriteRouteListContainer.propTypes = {
+  routes: React.PropTypes.array,
+  currentTime: React.PropTypes.object,
+  searching: React.PropTypes.bool,
+  location: React.PropTypes.oneOfType([React.PropTypes.object, React.PropTypes.bool]).isRequired,
+};
+
+// TODO: This should be moved above in the component hierarchy
+const FavouriteRouteListContainerWithTime = connectToStores(
```
Macroz: Maybe this could be split into multiple wrappers with separate time and position?

app/component/front-page/NearestRoutesContainer.js
```diff
@@ -0,0 +1,70 @@
...
+    modes: PropTypes.array.isRequired,
+  };
+
+  constructor() {
+    super();
+    this.useSpinner = true;
+  }
+
+  componentDidMount() {
+    this.useSpinner = false;
```
Macroz: Why is this set to true in constructor?

hannesj: Works as before. See [this](https://github.com/HSLdevcom/digitransit-ui/pull/584/files#diff-86eaa8a44fc413fd1b866b4c942534bbL11) for previous version. It is done in order to show the spinner on initial load, but not when updating data every 30 seconds.

PR #585, opened 2016-05-16
DT-980 Initialize state.suggestions

hannesj:


app/component/search/GeolocationOrInput.js
```diff
@@ -0,0 +1,44 @@
...
+    }),
+    initialValue: PropTypes.string,
+  }
+
+  state = {
+    geolocation: false,
+  };
+
+  componentWillMount() {
+    if (typeof this.props.endpoint !== 'undefined') {
```
hukka: This will fail if props.endpoint is null

app/component/search/GeolocationOrInput.js
```diff
@@ -0,0 +1,44 @@
...
+      this.setStateFromEndpoint(this.props.endpoint);
+    }
+  }
+
+  setStateFromEndpoint = ({ useCurrentPosition }) =>
+    (useCurrentPosition ? this.setState({ geolocation: useCurrentPosition }) : null)
+
+  getInitialValue = () =>
+    (typeof this.props.endpoint !== 'undefined' ?
```
hukka: Ditto

PR #587, opened 2016-05-16
DT-1041 Change now to set departure time

hannesj:


hukka:
Coffee, coffee, coffee

hannesj:
Note that as the time control is an uncontrolled component we can't change its value from the store, but only set the default value. If we change it to a controlled component, we must change the debounce logic

hukka:
LGTM apart from the binds in constructors instead of fat arrows

samuliheljo:
Are you sure that this works? On top bar, when I:
1) Use current master (caf6850)
2) Use my location as origin
3) route to kamppi (time 21:10, "lähtee") - works correctly
4) click "later" (time 21:22, "lähtee") - works correctly
5) click "earlier" (time 21:52, "saapuu") -- ?? seems like a bug?

hannesj:
Yes, the logic is not completely straight-forward due to the complexity of how to find the previous departures. You should find that the buses arriving in Kamppi leave at 21:52 departing at approx 21:10.

This issue was to fix the behaviour when the user clicks at now after having used the earlier button. Previously the arrival time was set at the current time but now the departure time is set to the current time. Also the timepicker in the bar above the map should change its time when clicking on later/now/earlier, previously it didn't update.

PR #591, opened 2016-05-17
DT-1011 Hide "Toggle Map Tracking" if no geolocation

rrudd:


app/component/map/MapWithTracking.js
```diff
@@ -0,0 +1,150 @@
...
+import Map from './map';
+import ToggleMapTracking from '../navigation/toggle-map-tracking';
+import connectToStores from 'fluxible-addons-react/connectToStores';
+import withReducer from 'recompose/withReducer';
+import onlyUpdateForKeys from 'recompose/onlyUpdateForKeys';
+
+function mapStateReducer(state, action) {
+  switch (action.type) {
+    case 'enable':
+      return Object.assign({}, state, {
```
hukka: Let's move to spread ({...}) for new objects, even if it's still stage 2.

PR #595, opened 2016-05-18
DT-1012 No redirect to summary page

rrudd:


app/action/PositionActions.js
```diff
@@ -0,0 +1,137 @@
...
+        return actionContext.dispatch('GeolocationDenied');
+      } else if (error.code === 2) {
+        return actionContext.dispatch('GeolocationNotSupported');
+      } else if (error.code === 3) {
+        return actionContext.dispatch('GeolocationTimeout');
+      }
+    }
+  };
+
+  if (window.position.error !== null) {
```
hukka: This will only catch nulls. That is, if window.position.error = null. If it's undefined or defined to undefined, the check will fail. != null with only one = will catch those too.

app/action/PositionActions.js
```diff
@@ -0,0 +1,137 @@
...
+    window.retrieveError(window.position.error);
+    window.position.error = null;
+  }
+
+  if (window.position.pos !== null) {
```
hukka: Ditto

app/action/PositionActions.js
```diff
@@ -0,0 +1,137 @@
...
+    done();
+    return;
+  }
+
+  actionContext.dispatch('GeolocationSearch');
+
+  let timeoutId = window.setTimeout(() => actionContext.dispatch('GeolocationWatchTimeout'), 10000);
+
+  window.retrieveGeolocation = (position) => {
+    if (window.position.pos !== null) {
```
hukka: See below

PR #600, opened 2016-05-18
Use configured font from theme in leaflet-container

csolem:
Make it possible to configure the font and weight for leaflet markers.


hannesj:
Uh, sorry merged this too soon. Could you replace this with the `@mixin font-book` mixin for consistency.

PR #603, opened 2016-05-19
CO2 and weather configurable

csolem:
Make route information configurable


kalon33:
@csolem is there currently some way to source these informations or are they still empty?

csolem:
I think they are empty. Currently, it's not our main focus to implement, but we need to hide the headings.

PR #620, opened 2016-05-23
DT-1025 No position fixed for search modal

rrudd:
This is a temporary solution for the beta release desktop version.


hannesj:
Does not work properly when adding favourite location

PR #631, opened 2016-05-25
Dt 950 dynamic citybike status

VilleMiekkoja:
The PR is done - it:

- [ ] all changed components

  * [ ] have examples
  * [ ] have unit tests


app/relay-route/CityBikeRoute.js
```diff
@@ -0,0 +1,17 @@
+import Relay from 'react-relay';
+
+export default class CityBikeRoute extends Relay.Route {
+  /* eslint-disable graphql/template-strings */
```
hannesj: Why are you disabling the rule. Shouldn't you just configure the variable for the query to fix the issue?

PR #640, opened 2016-05-26
Fix setOriginToDefault bug.

apkoponen:
The problem occured, when inside([pos.lon, pos.lat], config.areaPolygon)
evaluated to true == User location was outside of the configured area
polygon. E.g. when users outside of Helsinki area would go to a site using
HSL config.

The bug was caused by trying to import defaults from EndpointActions.js,
when there are none.


apkoponen:
This bug was caused by this [commit](https://github.com/HSLdevcom/digitransit-ui/commit/0311b48f687f2ca8407551452b187a3498658482), where module.exports where converted to ES6 exports, but no default was given.

hannesj:
Thanks!

PR #651, opened 2016-05-28
DT-1057 show disuptions in favourite route row

rrudd:


app/component/favourites/FavouriteRouteListContainer.js
```diff
@@ -14,6 +14,11 @@ const getNextDepartures = (routes, lat, lon) => {
   const seenDepartures = {};
 
   for (const route of routes) {
+    let hasDisruption = false;
```
Macroz: This can be simplified?

Macroz:
I think Firefox needs the fill style so I added it.

PR #659, opened 2016-05-31
DT-1049 Wait leg zooms map to wrong position

rrudd:


Macroz:
@rrudd I tried simplifying the code a little bit. Can you check my commits?

rrudd:
@Macroz All excellent improvements, thanks!

PR #661, opened 2016-05-31
DT-1092 show user position on trip route page

rrudd:


hannesj:
- [ ] Add example and description for the component
- [ ] Add the component to styleguide
- [ ] Add gemini tests

PR #679, opened 2016-06-03
Messages

hukka:
The PR is done - it:

- [x] follows the style and naming rules (passes `npm run lint`)
- [x] doesn't break anything (passes `npm run test-local` and `npm run test-browserstack`)
- [x] design is as expected. NOTE! visuals are compared using HSL theme (`CONFIG=hsl npm run dev`).
- [x] any changed files are transformed to ES6
- [ ] all changed components

  * [ ] have examples
  * [ ] have unit tests

To review, check all pages, itinerary that is short, itinerary that is long (needs scrolling), etc. The flex changes affect pretty much every page.


hukka:
This is DT-1017 and DT-1005

siren:
Looks like the map-tracking toggle has moved under the 'near you' button

siren:
I don't see the initial message appear on IE11 at all.

siren:
this problem remains: when the initial message is displayed and then closed there's a gray bar on bottom of the map when scrolled down util there's a tile boundary

app/component/navigation/MessageBar.js
```diff
@@ -0,0 +1,131 @@
+import React, { Component, PropTypes } from 'react';
```
siren: Two minor things I would consider:
  -move (unread)message filtering function to store (or atleast to connectToStores part)
  -change read status through action (and not directly through the store) so that we get rid of the component->store dependency

PR #686, opened 2016-06-06
DT-1136 use material-ui time picker in browsers with no native time input support

rrudd:


app/component/util/inputTypeCheck.js
```diff
@@ -0,0 +1,38 @@
+function supportsType(inputType) {
```
hannesj: This file should be in `app/utlis`, as it is not an utility component, similarly it should use and ES6 style export, i.e `export default function supportsType(inputType) {` Also the filename should match the exported function name according to AirBnB styleguide.

PR #691, opened 2016-06-07
Itinerary feedback (optional)

csolem:
Although there might be comprehensive plans for user feedback, it would be nice to be able to allow the user to comment on the itinerary summary, using Piwik. This feature is default turned off by a configuration toggle called: `itinerary.enableFeedback`. So everything should look and work like today, unless turned on. If the feature is turned on, some sass adjustments might be required in the theme to make it look all right. There are also three new translations.

The PR is done - it:

- [x] follows the style and naming rules (passes `npm run lint`)
- [ ] all changed components

  * [ ] have examples
  * [ ] have unit tests


hukka:
Starting review on this.

hukka:
We still haven't written down our extra style rules on top of airbnb, but we are using arrows instead of binds so that it's easier to see just from the function definition that it is using bound this. I also added translations to our branch.

One final thing is that you are using a speach [sic] bubble icon that is missing from our sprites. An empty bubble is a bit odd if the feature is enabled just from config, so should there be some text, or would you like to add the icon too?

PR #697, opened 2016-06-08
make texts better

siren:
The PR is done - it:

- [ ] all changed components

  * [ ] have examples
  * [ ] have unit tests


kalon33:
@siren Would it be possible to make these texts translatable, to allow desktop wrapper use in any country?

siren:
@kalon33 yeah, makes sense. I changed the logic so that now the content comes from config (desktopWrapperText) so you can customize it by installation. It is also possible to disable the desktop wrapper functionality completely: enableDesktopWrapper=true/false

If you like you could take a stab at making the texts translatable.

PR #706, opened 2016-06-09
Corrections to popup and card header styles

rrudd:
This PR changes the way stop codes are presented in card headers and fixes a styling issue with popup tips.


app/component/card/CardHeader.js
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import Icon from '../icon/icon';
+import cx from 'classnames';
+import ComponentUsageExample from '../documentation/ComponentUsageExample';
+
+const Favourite = ({
```
hannesj: Do you think this should be extracted and also used in RouteHeader?

app/component/card/CardHeader.js
```diff
@@ -0,0 +1,66 @@
...
+  addFavourite,
+  favourite,
+}) => (
+  <span className="cursor-pointer favourite-icon right" onClick={addFavourite}>
+    <Icon className={cx('favourite', { selected: favourite })} img="icon-icon_star" />
+  </span>);
+
+const CardHeader = ({
```
hannesj: Visual tests for this component should be added

PR #710, opened 2016-06-09
Only use M for subway if shortName not available

csolem:
The PR is done - it:

- [ ] all changed components

  * [ ] have examples
  * [ ] have unit tests

### DIFF
--- a/app/util/leg-text-util.js
+++ b/app/util/leg-text-util.js
@@ -4,7 +4,7 @@
 const MAX_ROUTE_LENGTH = 6;
 
 function getLegText(leg) {
-  if (leg.transitLeg && leg.mode.toLowerCase() === 'subway') {
+  if (leg.transitLeg && leg.mode.toLowerCase() === 'subway' && !leg.route.shortName) {
     // TODO: Translate these characters.
     return 'M';
   } else if (leg.transitLeg && leg.route.shortName &&


hannesj:
Hi, you can actually remove the hard coded `M` altogether. This is a data issue for us and should not require any extra configuration

tuukka:
@hannesj Will you fix the data?

hannesj:
This seems to be fixed in JORE for the new IDs 31M1 and 31M2.
13.6.2016 11.05 ap. "Tuukka Hastrup" <notifications@github.com> kirjoitti:

tuukka:
@hannesj It'll take half a year or so before those are in operation. I think we want to merge this earlier. And in the latest GTFS data, they are still broken:
```
31M1,HSL,,Helsinki-Siuntio,,1,https://www.vr.fi/cs/vr/fi/paakaupunkiseudun_lahiliikenne
31M1B,HSL,B,Helsinki-Siuntio,,1,https://www.vr.fi/cs/vr/fi/paakaupunkiseudun_lahiliikenne
```

hannesj:
@tuukka This is still a data issue that should be fixed upstream and not special cased by us, if we want to have a generic solution. Also there is only one subway, and the icon still reflects that.

tuukka:
@hannesj If you want to change the design to remove the M, I'm personally fine with it as long as it doesn't end up saying "null" or "undefined" ;-) By the way, I notice we call this function only in SummaryRow while we use shortName directly a lot.

hannesj:
Yes, after removing the hardcoded value, it looks like this:

screen shot 2016-06-13 at 14 36 06
screen shot 2016-06-13 at 14 36 18
screen shot 2016-06-13 at 14 36 41

The last picture is unchanged from the [current one](https://beta.reittiopas.fi/linjat/HSL:1300V:0:02), only hte summary page is affected.

hannesj:
@tuukka I'll merge this as is. We can remove the config after the shortName is included on the GTFS bundle.

tuukka:
:+1: Thanks @hannesj and @csolem!

PR #715, opened 2016-06-13
DT-552 Fix always scrolling route stop list

rrudd:


app/component/route/RouteStopListContainer.js
```diff
@@ -0,0 +1,107 @@
...
+      stopObjs.push(
+        <RouteStop
+          key={stop.gtfsId}
+          stop={stop}
+          mode={mode}
+          vehicles={vehicleStops[stop.gtfsId]}
+          distance={
+            (isNearest && nearest.distance < config.nearestStopDistance.maxShownDistance) ?
+            nearest.distance : null}
+          ref={
```
hannesj: I think you could use a [ref callback](https://facebook.github.io/react/docs/more-about-refs.html#the-ref-callback-attribute) here, i.e. something like `ref={(nearest) => nearest == null ? void 0 : nearest.scrollIntoView(false)}` and not needing to do it in the componentDidMount

PR #725, opened 2016-06-15
DT-1142 only include required timezone

rrudd:


app/util/configure-moment.js
```diff
@@ -1,14 +1,15 @@
...
 
-import moment from 'moment-timezone';
+import moment from 'moment-timezone/moment-timezone';
 
 // Configure moment with the selected language
 // and with the relative time thresholds used when humanizing times
 function configureMoment(language) {
   moment.locale(language);
 
   if (config.timezone) {
```
hannesj: This should probably be `config.timezoneData`

PR #727, opened 2016-06-16
Fix route schedule timezone.

apkoponen:
The timetable showed the times in the current locale
instead of UTC, which caused the times to be wrong
depending on the offset between the browser timezone
and UTC.

Cf. difference between HSL timetables and the view in
E.g.
http://aikataulut.reittiopas.fi/linjat/fi/h72_72n.html
https://dev.reittiopas.fi/linjat/HSL:1072:1:02

### DIFF
--- a/app/component/route/RouteScheduleContainer.js
+++ b/app/component/route/RouteScheduleContainer.js
@@ -62,8 +62,10 @@ class RouteScheduleContainer extends Component {
         </div>);
     }
     return trips.map((trip) => {
-      const departureTime = this.formatTime(trip.stoptimes[stops[from].id].scheduledDeparture);
-      const arrivalTime = this.formatTime(trip.stoptimes[stops[to].id].scheduledArrival);
+      const fromSt = trip.stoptimes[stops[from].id];
+      const toSt = trip.stoptimes[stops[to].id];
+      const departureTime = this.formatTime(fromSt.serviceDay + fromSt.scheduledDeparture);
+      const arrivalTime = this.formatTime(toSt.serviceDay + toSt.scheduledArrival);
 
       return (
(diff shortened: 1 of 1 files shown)


hannesj:
There will also be problems on dates with DST changes due to how OTP uses time (midnight is calculated as 12 hours before noon, meaning it could be at 23:00 or 01:00). Therefore it would be better to add `serviceDay`field inside stoptimes in the GraphQL query to fetch the midnight for the selected date. Then you can use the existing formatTime function for serviceDay + scheduledDeparture or serviceDay + scheduledArrival.

apkoponen:
I rebased the commit and added serviceDay. Did I understand correctly that serviceDay for stoptime is value that the start of the day is different from actual midnight? I'm just checking, because "serviceDay" is used very differently intripsForDate(serviceDay: $serviceDay). All of the serviceDays I got in the query were 0.

tuukka:
@apkoponen I think that's a bug in the API, the returned `serviceDay` should not be 0 but e.g. 1466629200 for today (start of day in local timezone). (Or it should be -7200 if the day is hardcoded as 1970-01-01).

Btw, if you haven't had a look before, the API implementation starts here: https://github.com/HSLdevcom/OpenTripPlanner/blob/master/src/main/java/org/opentripplanner/index/IndexGraphQLSchema.java

apkoponen:
@tuukka That's what I was wondering. Thanks for the link, if I have time, I'll see if I can find where the actual magic happens.

hannesj:
The timestamp for midnight can be gotten from stoptimesForDate, I fixed the minor issue in #784 and will merge this from there. Thanks for noticing and fixing this!

PR #733, opened 2016-06-17
Add POC for testing out setting location from map click

hannesj:
TODO:

- [x] Add localization for text
- [ ] Add reverse geocoding for addresses
- [ ] Investigate wether this is good UX


tuukka:
Reverse geocoding could be added now, or could be left for later. UX can be investigated later, as this only changes what happens on map long-press / right-click, which didn't do anything before.

PR #744, opened 2016-06-21
DT-1155 new route page header

rrudd:


siren:
@rrudd  there's currently merge conflict that needs to be resolved

rrudd:
Merge conflict resolved. Should something be done about the empty disruptions tab before this is merged/deployed?

siren:
Thanks. There's a separate issue for the disruptions tab (DT-1156) so I would just leave it as it is.

PR #760, opened 2016-07-01
add from-to markers to summary search fields

rrudd:


tuukka:
With the Finland config, labels A and B in the markers are grey (text color from theme) although they should be white. (Personally, I think the markers should have a white border too, but that's not what the design image looks like.)

app/component/summary/OriginDestinationBar.js
```diff
@@ -0,0 +1,110 @@
...
+  constructor() {
+    super();
+
+    this.state = {
+      origin: undefined,
+      destination: undefined,
+      tabOpen: false,
+    };
+
+    this.componentWillMount = this.componentWillMount.bind(this);
```
tuukka: No need to bind the methods that you don't pass as callback functions.

app/component/summary/OriginDestinationBar.js
```diff
@@ -0,0 +1,110 @@
...
+  render() {
+    const ownPosition = this.context.intl.formatMessage({
+      id: 'own-position',
+      defaultMessage: 'Your current location',
+    });
+
+    let initialValue = '';
+
+    if (this.state[this.state.tabOpen]) {
+      initialValue = this.state[this.state.tabOpen].useCurrentPosition ?
```
tuukka: Perhaps use `if` here, as you are assigning the result to a non-const. Or I would at least have the both sides of the colon : on the same line.

app/component/summary/OriginDestinationBar.js
```diff
@@ -0,0 +1,110 @@
...
+        ownPosition :
+        this.state[this.state.tabOpen].address;
+    }
+
+    return (
+      <div className="origin-destination-bar">
+        <div className="field-link from-link" onClick={() => this.openSearch('origin')}>
+          <Icon img={'icon-icon_mapMarker-point'} className="itinerary-icon from" />
+          <span className="dotted-link">
+            {this.state.origin.useCurrentPosition ?
```
tuukka: This would fit on one line (http://airbnb.io/javascript/#comparison--nested-ternaries).

PR #772, opened 2016-07-07
DT-1159 fullscreen route map

rrudd:


app/page/RoutePage.js
```diff
@@ -86,34 +89,68 @@ class RoutePage extends React.Component {
...
+    if (typeof window !== 'undefined' && supportsHistory()) {
+      const state = this.context.location.state;
+      return state && state.fullscreenMap;
+    }
+
+    return this.state && this.state.fullscreenMap;
+  };
+
+  toggleFullscreenMap = () => {
+    if (supportsHistory()) {
```
hannesj: check for supportsHistory is not needed anymore as support for IE 9 was dropped, so this.state. fullscreenMap can be removed

PR #788, opened 2016-07-22
Fix hilightedStops in TileLayer implementation

hannesj:


hukka:
Do we have a design to check this against?

hannesj:
Not with the current stop marker style, so I adapted this from the old marker style. See https://projects.invisionapp.com/share/5X5RNXGGU#/screens/130673127

PR #789, opened 2016-07-22
Fix favourite icon position on stop page

hannesj:


hukka:
Is the gemini error on snap real?

hannesj:
It seems to be the real-time icon on the TripRouteStop, that was fixed in some other branch, but I can pull the fix also to this branch

hukka:
Gemini on snap still failing

PR #798, opened 2016-08-01
iPhone emulator testing in SauceLabs

hukka:
The PR is done - it doesn't touch any js files.


hukka:
Works on Linux. Doesn't work on mac, needs someone to tweak and test that.

PR #814, opened 2016-08-17
DT-1256 Platform code in stop icon

hannesj:
The PR is done - it:

- [ ] all changed components

  * [ ] have examples


app/util/mapIconUtils.js
```diff
@@ -70,6 +71,15 @@ export function drawRoundIcon(tile, geom, type, large) {
       );
 
       tile.ctx.fill();
+
+      if (platformNumber && hubRadius > 7) {
```
samuliheljo: Instead of using "7", I would either:
- pass empty platformNumber to this function if z-level is not what is configured in config; or
- pass z-level to this function and compare against configured value in config.

in any case number 7 should IMHO be in config

hannesj: It is dependent on if it fits inside the hub, which requires exactly 14 pixels. The size of the hub varies e.g. if it is selected and therefore is not bound to any specific zoom level.

hannesj: Added a comment for that

PR #876, opened 2016-09-22
DT-1373

hukka:


hukka:
See also https://github.com/HSLdevcom/digitransit-openshift-deploy/pull/19 and https://github.com/HSLdevcom/digitransit-proxy/pull/15

PR #892, opened 2016-10-04
fix white turku stops

samuliheljo:


hukka:
Tests pass in BS when running against local server - overriding merge status problem from Snap.

PR #897, opened 2016-10-10
Add server-side data fetching for relay

hannesj:
Also re-adds meta-tag generation.


samuliheljo:
I'm not sure whether I'm checking this correctly. But it seems that relayData is empty and browser makes graphql requests on refresh. 

What I do is:
1) Search a route from A -> B
2) Refresh browser page
3) View sources and see: `<script type="application/json" id="relayData">[]</script>`

Am I missing something?

hannesj:
Before #886 is merged it won't work on itinerary/summary page, only on stop/route/trip pages.

PR #899, opened 2016-10-11
Improve Finnish texts DT-1492

samuliheljo:
This modifies only finnish texts, translations will follow up when Finnish texts are ready


hukka:
Otherwise ok, but doesn't pass lint. I'll make changes and merge

PR #906, opened 2016-10-17
Fix memory leak DT-1479

samuliheljo:
You can check node GC by adding following start flags:
--debug --trace-gc

Reason for memory leak is that componentWillUnmount is not called on server, therefore leaving unwanted listeners in heap.

### DIFF
--- a/app/component/summary/OriginDestinationBar.js
+++ b/app/component/summary/OriginDestinationBar.js
@@ -23,12 +23,15 @@ class OriginDestinationBar extends React.Component {
     tabOpen: false,
   };
 
-  componentWillMount = () => {
+  componentWillMount() {
     this.onEndpointChange();
+  }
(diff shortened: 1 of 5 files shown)


app/component/summary/OriginDestinationBar.js
```diff
@@ -23,12 +23,15 @@ class OriginDestinationBar extends React.Component {
     tabOpen: false,
   };
 
-  componentWillMount = () => {
+  componentWillMount() {
```
hukka: IIRC while it wasn't unanimous, we decided that binding lambdas are the default whenever `this` doesn't explicitly need to be unbound, even if react handles the magic automagically for the lifecycle methods.

hukka:
Is the failing test spurious, or a real failure?

PR #913, opened 2016-10-25
DT-1493

siren:
Make tests pass on small screen too
-verify clicks -> fail faster
-try to avoid static pauses


siren:
here are some runs with the tests run on snap:

https://snap-ci.com/siren/digitransit-ui/branch/master

PR #931, opened 2016-11-03
DT-1502 stop popup

Macroz:
Implements the stop popup and adds visual tests.

Some considerations
- Styling. The different popups look slightly different. What is the desired look?
- Uses origin popup classes. Maybe they should be renamed to be more generic? Or do when unifying the styles?
- When you open a stop popup while on the stop page you can change the selected stop. This does not close the more detailed popup and open this new label. Is this ok or not?
- Seems that the stops layer is not redrawn when the selected stop is changed. This causes a bug that the "big circle" does not move to the currently selected stop.


Macroz:
Styling has been updated now to match suggestion. So the first bullet is solved.

Macroz:
@hannesj Is the TimeNavigationControls test fix in master already?

hannesj:
@Macroz It was in https://github.com/HSLdevcom/digitransit-ui/commit/19c78b35b6116e70dc1cab4a844db5337e629b8c, which was merged in https://github.com/HSLdevcom/digitransit-ui/pull/934

PR #982, opened 2016-11-16
DT-1555 use batching for GraphQL queries

hannesj:


hannesj:
URL fixed

samuliheljo:
test passed locally

PR #1031, opened 2016-11-23
DT-1619 Optimize JavaScript build

hannesj:


samuliheljo:
Couple of failing flow tests :(

PR #1034, opened 2016-11-24
DT-1625 Use type icon as terminal symbol

hannesj:


samuliheljo:
What about this cluster popup icon:
screen shot 2016-11-25 at 14 14 54

We should:
1. probably change that also? 
2. remove old terminal icon from svg? or not?

samuliheljo:
Hmm... visual tests passed, without changing the image... maybe not correct result?

PR #1074, opened 2016-12-04
Improve build

hannesj:
The PR is done - it:

- [ ] all changed components

  * [ ] have examples


server/server.js
```diff
@@ -43,7 +44,8 @@ function setUpStaticFolders() {
   const staticFolder = path.join(process.cwd(), '_static');
   // Sert cache for 1 week
   const oneDay = 86400000;
-  app.use(config.APP_PATH, express.static(staticFolder, {
+  app.use(config.APP_PATH, expressStaticGzip(staticFolder, {
```
gmokki: The experss-static-gzip does not correctly implement the Vary header because it does not send the Vary header always for resource that has a precompressed version available.

The Vary header is only used by caching proxies and the failure case can be for example that:
1. user1 fetches app.js without Accept-Encoding header (or with Accept-Encoding value that has no precompressed file such as bzip2)
2. the express-static-gzip does not Vary header in response
3. proxy caches the result
4. user2 asks for the resource with Accept-Encoding: br, gzip
5. the proxy finds the uncompressed version and serves that one (because the response did not have proper Vary)

If the uncompressed also had a Vary header the proxies would not get 'poisoned' with uncompressed versions. This is normally not a problem for internal applications, but with the wild west of internet and it's caches it would be much safer to have the header correctly implemented.

hannesj: https://github.com/tkoenig89/express-static-gzip/issues/6

hannesj: Fixed in 0.2.2

PR #1083, opened 2016-12-07
Dt 1624

samuliheljo:
When deploying this, we need to configure Liikennevirasto to use "CONFIG=matka"


app/config.matka.js
```diff
@@ -0,0 +1,450 @@
+const CONFIG = process.env.CONFIG || 'default';
```
hannesj: Only need values that are different from the default config. See eg. HSL's config

app/config.matka.js
```diff
@@ -0,0 +1,87 @@
+const CONFIG = process.env.CONFIG || 'matka';
+const APP_DESCRIPTION =
+  'Liikenneviraston Matka.fi uudistuu. Apuasi kaivataan kehitystyössä. ' +
+  'Tule palvelun testaajaksi tai tee siitä saman tien parempi.';
+
+export default {
+  CONFIG,
+
```
hannesj: Should probably include 

`
  availableLanguages: ['fi', 'sv', 'en'],
  defaultLanguage: 'fi',
`

samuliheljo: fixed

PR #1084, opened 2016-12-07
DT-1639 remove geolocation.js

hannesj:


samuliheljo:
Flow tests failed in snap (and also locally).

hannesj:
Restructured slightly so that the mock kicks in before rendering intro

samuliheljo:
Some challenges with tests, you probably noticed too...

PR #1093, opened 2016-12-09
DT-1662 Show agency in summary (& nearby)

Macroz:
Metro 'M' should be refactored sometime.


app/component/summary-row.scss
```diff
@@ -90,8 +90,9 @@
     flex-grow: 1; //Needed on IE
 
     .leg {
-      flex: 1 0 0;
```
hannesj: Or use https://github.com/luisrudge/postcss-flexbugs-fixes

Macroz: Seemed to me that the page worked fine even without the fix so I have removed it for now.

app/component/RouteNumberContainer.js
```diff
@@ -0,0 +1,32 @@
...
+  } else if (showAgency && route.agency) {
+    return route.agency.name;
+  }
+  return '';
+};
+
+const getMode = route =>
+  route.mode || (route.rentedBike && 'CITYBIKE');
+
+const SmartRouteNumber = ({ route, ...props }) =>
```
hannesj: Should be RouteNumberContainer?

hannesj:
The issue catched by tests seems to be a real issue

PR #1115, opened 2016-12-14
DT-1607 add max width to about page content

Macroz:
Leave the rest of the task to wait until (disruption) modal is refactored to support back button.

### DIFF
--- a/app/component/front-page.scss
+++ b/app/component/front-page.scss
@@ -436,7 +436,7 @@ svg.icon.footer-icon {
   flex-direction: column;
   align-items: center;
   justify-content: center;
-  & > div {
+  > div {
     max-width: 600px;
   }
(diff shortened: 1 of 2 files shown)


Macroz:
Possibly could have separate SCSS for about page but eventually about page should be under index / front page so maybe the styles can be there for now.

app/component/front-page.scss
```diff
@@ -430,3 +430,13 @@ svg.icon.footer-icon {
   bottom: 40px;
   left: 50px;
 }
+
+.about-page {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  & > div {
```
hannesj: `& ` is not needed when nesting rules

PR #1139, opened 2016-12-18
Add german translation

robbi5:
We've used the digitransit-ui for an quick hack at the #dbhackathon and translated the interface to german. Here are the cleaned and hopefully complete translation entries for de-DE :)


samuliheljo:
Great, thanks!

PR #1155, opened 2016-12-21
handle SecurityError from localStorage

tuukka:
This happens if user has denied cookies in browser settings.

### DIFF
--- a/app/store/localStorage.js
+++ b/app/store/localStorage.js
@@ -8,6 +8,9 @@ function setItem(key, value) {
       if (error.name === 'QuotaExceededError') {
         console.log('[localStorage]' + // eslint-disable-line no-console
           ' Unable to save state; localStorage is not available in Safari private mode');
+      } else if (error.name === 'SecurityError') {
+        console.log('[localStorage]' + // eslint-disable-line no-console
+          ' Unable to save state; access to localStorage denied by browser settings');
       } else {
         throw error;
       }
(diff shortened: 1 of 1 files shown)


app/store/localStorage.js
```diff
@@ -16,8 +19,18 @@ function setItem(key, value) {
 }
 
 function getItem(key) {
-  return (isBrowser && window.localStorage &&
-    window.localStorage.getItem(key)) || null;
+  if (isBrowser && window.localStorage) {
+    try {
+      return window.localStorage.getItem(key);
+    } catch (error) {
+      if (error.name === 'SecurityError') {
```
hannesj: Could the catch block (or even the whole try/catch) be extracted, so it doesn't need to be duplicated?

app/store/localStorage.js
```diff
@@ -1,5 +1,15 @@
 import { isBrowser, isWindowsPhone, isIOSApp } from '../util/browser';
 
+function handleSecurityError(error, logMessage) {
+  if (error.name === 'SecurityError') {
+    if (logMessage) {
+      console.log(logMessage); // eslint-disable-line no-console
+    }
```
hannesj: Should we have else here?

tuukka: No, since the only thing to do is to log any messages.

hannesj: Uh, of course...

PR #1195, opened 2017-01-03
DT-1676

siren:
The PR is done - it:

- [ ] all changed components

  * [ ] have examples


samuliheljo:
1) This seems to change how UI looks even without date. itinerary-start-date span and other style modifications, that is.
-> Let's keep UI as it is when date is not available

2) Some rows have problems like:

screen shot 2017-01-03 at 15 56 53

--> fix needed


3) When itinerary is opened we use different timeformat. (other has weekday e.g. wed. and other not)

screen shot 2017-01-03 at 15 57 03

--> Use "Ke 7.12" format also when itinerary is opened

hannesj:
screen shot 2017-01-10 at 15 03 09

The positioning still seems to be off

app/component/DateWarning.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import moment from 'moment';
+import { FormattedMessage } from 'react-intl';
+
+const DATE_PATTERN = 'dd D.M.';
+
+const DateWarning = ({ date }) => {
+  const now = moment();
```
hannesj: I think this should be the requested date, not the current time. So if you search for a departure for tomorrow, you should get a warning if it departs on the day after tomorrow

app/component/SummaryRow.js
```diff
@@ -100,31 +101,38 @@ export default function SummaryRow(props, { breakpoint, intl: { formatMessage }
...
   const classes = cx(['itinerary-summary-row', 'cursor-pointer', {
     passive: props.passive,
-    'bp-large': breakpoint === 'large',
+    'bp-large': props.breakpoint === 'large',
     open: props.open || props.children,
   }]);
 
+  const NOW = moment();
+
+  const sameDay = dateOrEmpty(startTime, NOW) === '';
```
hannesj: This should also compare the requested date, not today's date

app/component/SummaryRow.js
```diff
@@ -100,31 +101,38 @@ export default function SummaryRow(props, { breakpoint, intl: { formatMessage }
...
-        <FormattedMessage
-          id="itinerary-page.title"
-          defaultMessage="Itinerary"
-          tagName="h2"
-        />,
+        <span className="flex-grow itinerary-heading">
+          <FormattedMessage
+            id="itinerary-page.title"
+            defaultMessage="Itinerary"
+            tagName="h2"
```
hannesj: you shouldn't have a block element inside an inline element.

app/component/SummaryRow.js
```diff
@@ -143,7 +151,7 @@ export default function SummaryRow(props, { breakpoint, intl: { formatMessage }
         <div
           className={cx('itinerary-start-time', { 'realtime-available': realTimeAvailable })}
           key="startTime"
-        >
+        ><span className={cx('itinerary-start-date', { nobg: sameDay })} ><span>{dateOrEmpty(startTime, NOW)}</span></span>
```
hannesj: This should also compare the requested date, not today's date. Also consider splitting the elements to individual rows for readability

app/component/TimeFrame.js
```diff
@@ -1,14 +1,49 @@
...
+    <span className="capitalize">
+      <span className="timeframe-nextday">{momentTime.format(DATE_PATTERN)}</span>
+      &nbsp;
+      <span>{momentTime.format(TIME_PATTERN)}</span>
+    </span>
+  );
+};
+
+const TimeFrame = ({ className, startTime, endTime }) => {
+  const now = moment();
```
hannesj: This should also compare the requested date, not today's date

app/component/TimeFrame.js
```diff
@@ -1,14 +1,49 @@
...
-  <span className={cx(props.className)}>
-    {'// '}{moment(props.startTime).format('HH:mm')} - {moment(props.endTime).format('HH:mm')}
-  </span>
-);
+const TIME_PATTERN = 'HH:mm';
+const DATE_PATTERN = 'dd D.M.';
+/**
+ * Returns date or '' if same day as reference
+ */
+const dateOrEmpty = (momentTime, momentRefTime) => {
```
hannesj: Should this be split into its own file or some kind of date-utils?

app/component/itinerary.scss
```diff
@@ -466,3 +470,31 @@ div.itinerary-container-content {
...
+  padding-top: 15px;
+}
+
+.timeframe-nextday {
+  background-color: $gray;
+  line-height: 1;
+  @include font-narrow-medium;
+  color: $white;
+  border-radius: 2px;
+  padding: 4px 5.5px;
```
hannesj: should these be some of the padding-variables?

app/component/summary-row.scss
```diff
@@ -1,5 +1,5 @@
 .itinerary-summary-row {
-  min-height: 3.3em;
+  min-height: 50px;
```
hannesj: Is this intentional? Check with Raoul.

Macroz: Is the previous value based on something? Zeplin says it should be 50 pixels.

app/component/itinerary.scss
Macroz: The new values are based on the Zeplin spec. Do our padding variables match the spec or how to figure out which should be used?

Macroz:
Implemented preliminary support for using a reference time when showing date warnings. Need to test still and check the Leaves/Arrives behavior.

app/component/summary-row.scss
rrudd: Specifying min-height in px is totally fine, as long as the actual size of the element scales as it does in this case.

app/component/itinerary.scss
rrudd: Since moving to a layout that scales within the breakpoint, relative units are generally suggested. Here the challenge is to select one of the padding variables that closely corresponds with the spec at the intended window size (iphone 5). If a new padding variable is unavoidable, feel free to add it.

Macroz: I changed these to use the closest variables instead. Needed to add one for small border radius.

Macroz:
I think this is ready now. Any more open issues?

PR #1206, opened 2017-01-07
Make route number component a bit wider

ttsirkia:
In Disruption info the current width is not enough to show line numbers such as 206A.

route-number

This wider version should work also in cases such as 123AB.

### DIFF
--- a/app/component/departure.scss
+++ b/app/component/departure.scss
@@ -118,7 +118,7 @@
 
 // route-number component
 .route-number {
-  width: 3.5em;
+  width: 4.5em;
   vertical-align: top;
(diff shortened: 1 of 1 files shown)


hannesj:
Fixed in https://github.com/HSLdevcom/digitransit-ui/commit/6f53412dd2dcbe261ca341ba29dfd7e85a8f7fd6

tuukka:
@ttsirkia In the end, it seems to be even better visually to not set a width at all, as the line number width can vary from that of 1 to that of 989BB. Thank you for the report and proposed fix!

PR #1221, opened 2017-01-13
DT-1728 Empty space on route under vehicle

hannesj:


samuliheljo:
Gemini fails :(

hannesj:
Gemini passes now

Issue #1224, opened 2017-01-14
Dropdown caret is not clickable


ttsirkia:
It is not possible to click the caret:

![hsl-dropdown](https://cloud.githubusercontent.com/assets/2797729/21958366/7f748496-dab5-11e6-9952-53e44ef1f35c.png)

And hovering mouse over the date or time does not show the hand cursor indicating that it is clickable.

ttsirkia:
Caret seems to be clickable now although pointer cursor still not present although #1225 exists.

hannesj:
Fixed by https://github.com/HSLdevcom/digitransit-ui/pull/1330

PR #1225, opened 2017-01-14
Add pointer cursor to dropdown

ttsirkia:
Partially fixes the usability problem in #1224 that the caret is not clickable and the text does not indicate that user can click it.


hannesj:
Merged in https://github.com/HSLdevcom/digitransit-ui/pull/1330

PR #1246, opened 2017-01-23
DT-1701 new time

Macroz:
Now using the time-input 3rd party component.

Notes:
- There are some style fixes in [DT-1635](https://digitransit.atlassian.net/browse/DT-1635)
- Is the UX good enough?
- Which browsers to use this on (Chrome desktop supports native time input but this provides better UX)?


Macroz:
I think this is good enough to be tested already. Specifically check UX first for showstoppers.

Macroz:
Works on the IE Mobile & desktop browsers.

Maybe not the optimal UX for all cases, but I would like to not complicate this component more.

app/component/TimeInput.js
```diff
@@ -0,0 +1,66 @@
+import React, { Component, PropTypes } from 'react';
+
+import debounce from 'lodash/debounce';
+
+const parseValue = (x) => {
+  const value = x.replace(/[^0-9:]/g, '');
```
hannesj: In Finnish the hour and minute should be separated by a full stop, rather than a colon, which we should support [Ref](http://www.kielitoimistonohjepankki.fi/haku/kellonaika/ohje/51)

PR #1250, opened 2017-01-24
Dt 1742

vesameskanen:
The PR is done - it:

- [ ] all changed components

  * [ ] have examples


app/client.js
```diff
@@ -50,13 +51,22 @@ const piwikPlugin = {
...
 
 const ravenPlugin = {
   name: 'RavenPlugin',
   plugContext: plugContext(addRaven),
 };
 
+const addConfig = context =>
+ (context.config = window.config); // eslint-disable-line no-param-reassign
+
+const configPlugin = {
```
hannesj: Isn't this done already on the server?

app/util/Raven.js
```diff
@@ -1,22 +1,23 @@
...
 import { COMMIT_ID } from '../buildInfo';
 
-function getRaven() {
-  if (process.env.NODE_ENV === 'production') {
-    /* eslint-disable global-require */
-    const Raven = require('raven-js');
-    Raven.addPlugin(require('raven-js/plugins/console'));
-    /* eslint-enable global-require */
+function getRaven(config) {
+  return function createRaven() {
```
hannesj: Is this level required?

app/store/PreferencesStore.js
```diff
@@ -18,14 +17,18 @@ function loadLanguage() {
 class PreferencesStore extends Store {
   static storeName = 'PreferencesStore';
 
-  language = loadLanguage();
+  constructor(dispatcher) {
+    super(dispatcher);
+    this.config = dispatcher.getContext().config;
```
hannesj: could we load availableLanguages directly here?

app/store/PreferencesStore.js
```diff
@@ -1,8 +1,7 @@
 import Store from 'fluxible/addons/BaseStore';
 import reactCookie from 'react-cookie';
-import config from '../config';
 
-function loadLanguage() {
+function loadLanguage(config) {
```
hannesj: this could be moved as an instance method inside the store

app/client.js
vesameskanen: In the current setup, this makes the config available via fluxible context. Removewthe plugin -> code stops working. Maybe the same functionality can be achieved in some other (simpler?)  way.

app/store/PreferencesStore.js
vesameskanen: Language initialization (2 code lines) moved into constructor.

app/util/Raven.js
vesameskanen: Right - config is visible in the client.js where plugin is defined.  Wrapper is indeed redundant, removed.

PR #1255, opened 2017-01-26
feat: add HSL FAQ link

Macroz:
Adds FAQ link to HSL with the designated URL.

Link is called FAQ because others would be so long that the menu or footer get too long.


hannesj:
Should this be `UKK` in finnish?

PR #1212, opened 2017-01-09
DT-1655 new metadata

Macroz:
The JIRA-issue lists a bunch of open questions but otherwise review can be started already.

Image has not been updated yet.

PR #1258, opened 2017-01-27
DT-1761 update swedish translations

rrudd:


app/translations.js
```diff
@@ -220,43 +220,49 @@ const translations = {
   'sv': {
     'frontpage': 'Framsidan',
     'stop-page.title-short': 'Hållplats',
+    'stop-page.title': 'Hållplats {stop_name} - {stop_code}',
```
hannesj: It seems the translation is based on an older file, as the variable names, were changed in https://github.com/HSLdevcom/digitransit-ui/pull/1212/files

hannesj: We probably should start the translation fixes by enabling eslint rule `sort-keys` for this file and committing the messages in correct order, in order to make tracking changes like this easier.

Also, we probably should group these messages somehow.

PR #1286, opened 2017-02-01
Dt 1034 - error page

samuliheljo:


app/routes.js
```diff
@@ -327,7 +333,6 @@ export default (config) => {
                   content: TripQueries,
                   meta: RouteQueries,
                 }}
-                render={{ title: ({ props, element }) => React.cloneElement(element, props) }}
```
hannesj: Removing these affect what is shown when moving between route and stop pages. We probably should design what/how to show during the loading of data.

hannesj:
The error page seems to look wrong on safari

siren:
@hannesj to me error page looked similar to what chrome has, if you disagree, feel free to 'polish' it.

PR #1317, opened 2017-02-09
Change routing parameter slider color when set back to initial value

vesameskanen:


hannesj:
Sliders get reset to gray if page is refreshed :(

PR #1329, opened 2017-02-11
DT-1770 Redirect old reittiopas query parameters

hannesj:


vesameskanen:
Should we validate the params? Now redirection to the new service activates even when parameter set is invalid, like when only start point  is set.

PR #1335, opened 2017-02-11
Improve timepicker appearance when editing

ttsirkia:
The current way of editing time in desktop browser is bad because it does not clearly indicate that user should directly edit the field as the previous timepicker component is no longer in use. This PR improves the current situation as follows:

1) Disable selecting all the text in the field to make it act as a typical input field (cursor blinks)
2) Use white background and black text when the field is active

The result in Firefox (`type="text"`):
timepicker-ff
There is a little glitch as the input component does not fully take the space because of the padding of the parent container.

The result in Chrome (`type="time"`)
timepicker-chrome
White background also hides the incorrect caret symbol.

Unfortunately the timepicker component in Chrome is unintuitive to use and not as good as in mobile browsers.

### DIFF
--- a/app/component/TimeInput.js
+++ b/app/component/TimeInput.js
@@ -48,7 +48,6 @@ class TimeInput extends Component {
         className="text-time-selector"
         defaultValue={this.props.value}
         onChange={this.onChange}
-        onClick={() => { this.refs.time.select(); }}
         maxLength="5"
         style={{
           display: 'inline-block',
--- a/app/component/summary.scss
+++ b/app/component/summary.scss
@@ -142,6 +142,11 @@
     }
   }
 
+  input[type="time"]:focus, input[type="text"]:focus {
+    background-color: $white;
+    color: $black;
+  }
+
   .right-offcanvas-toggle {
       span {
         color: $primary-font-color;


tuukka:
Thanks for the proposal! I'm not sure we want this change though, at least not on mobile. Automatically selecting is more efficient for the user, as they can type the five characters without first having to delete the old characters. Also, the box-in-a-box-in-a-box visual appearance is not that nice. I wonder if it would be better to fake the input cursor with something like `border-right: solid 1px black` instead. How to hide it when the real cursor appears though - perhaps on key press?

We also need to support IE.

ttsirkia:
The change does not affect mobile browsers as the style change is inside `.bp-large.time-selector-settings-row`. IE works as FF, therefore no separate screenshot.

Seeing the cursor is quite important. If selecting all the text is really necessary, `onFocus` should definitely be used instead of `onClick`.

ttsirkia:
Changing the background color of the parent container is tricky as there is not a selector for that, `:focus` affects only that item and `:focus-within` is not implemented in any browsers, yet.

The most important issue here is the consistent appearance of the editable text fields. If all the other fields have white background, this should not be an exception.

ttsirkia:
And the little color glitch is visible only when the field is active because of `:focus` selector.

ttsirkia:
Just to concretize with one more picture:
reittiopascolors

Double-clicking "Reittiehdotukset" and clicking the time. The result is the same and even the mouse cursor is the same. No blinking cursor in either of them. Why the time is editable but the another text is not? How the user is supposed to know that the time is editable in this case without just trying? The difference is actually quite remarkable when changing the background color when the field is active, it gives an instant feedback that it is actually an editable field. The visible cursor would even more emphasize this.

I was very confused and did not realize that it is editable after the picker went away because the visual feedback is so poor. And if it was confusing for me, it must be for many less experienced users as well.

hannesj:
This has now been fixed in https://github.com/HSLdevcom/digitransit-ui/pull/1571

PR #1357, opened 2017-02-16
Navigation and hierarchy documentation

rrudd:


vesameskanen:
Some quick notes: 
- Maybe 'Up' button should be roughly explainened first (or maybe even an image link added). The document assumes that the reader already knows which button is under discussion. Especially as left arrow does not associate well with 'up'.
- App does not work as Up button specific docs explain. Currently Up button just duplicates native back button. This might be OK, we just need to fix the app.
- After testing mobile navigation once again, the confusion with arrows and crosses felt kind of annoying. Try Favorites->Add llocation->Search address path and observe how arrows/crosses come and go.

rrudd:
To answer your questions:
- I added a suggestion to first read the link supplied in order to get a hunch of the concept. I also added an image of the Up button. However, the design may be subject to change in the future.
- You are correct. This is currently a defect in the service.
- This may be true, but the issue should be escalated in terms of concepting and design. This task  intends to establish guideline documentation for how UI navigation should be handled.

vesameskanen:
Thanks for the improvements! I found two more issues which would be useful:
- A link from the main README.md to this navigation doc
- Image links do not seem to work with relative paths. They probably require similar url as in the image link of README.md.

PR #1358, opened 2017-02-16
Refactor config/aboutThisService 

vesameskanen:
aboutThisService block of most configurations were still in quite a messy state: some sv translations were in english, some translations missed part of information etc. 

Now per theme configurations inherit all standard parts of aboutThisService from a base theme, so that maintenance in future will be easy. Base theme contents and translations are mostly copied from hsl theme, which presumably was in pretty good shape.
 
Also, this PR adds an automatic host name to theme mapping, when a new theme is initialized.


hannesj:
Could you include https://github.com/HSLdevcom/digitransit-ui/pull/1285

vesameskanen:
Yes I will. #1285 complicates certain things as it changes hashtables into arrays, making property inheritance harder to implement.  Not impossible, though.

hannesj:
One option would be to keep the objects, but have object with header and text as values. This way you could override them, and remove sections by changing the value of the key to null.

vesameskanen:
Works now with arrays. Sub themes can also replace or remove sections from parent theme.

PR #1379, opened 2017-02-24
Offer routing dates which are based on service date range of the routing graph

vesameskanen:


app/component/TimeSelectorContainer.js
```diff
@@ -45,28 +50,39 @@ class TimeSelectorContainer extends Component {
...
+        label = day.format('dd D.M');
+      }
       dates.push(
-        <option value={date.add(1, 'd').format('YYYY-MM-DD')} key={date.format('YYYY-MM-DD')}>
-          {date.format('dd D.M')}
+        <option value={value} key={value} >
+          {label}
         </option>,
       );
+      if (value === endValue) {
```
hannesj: Should this be the end condition in the for loop?

vesameskanen: Surely it could, e.g. with a do-while loop.

PR #1391, opened 2017-02-26
Only apply overflow-fade on search-header__button's contents

hannesj:
Local PR so it runs in CI. Closes #1390


hannesj:
Seems something changed on BrowserStack and we needed to generate couple of images again. Should be fixed in master

hannesj:
Thanks @FruitieX

PR #1403, opened 2017-03-01
NRP-1132 Use 2x tiles if retina tiles configured OR L.Browser.retina

Eightyplus:
Use 2x tiles if retina tiles configured OR L.Browser.retina.

### DIFF
--- a/app/component/map/Map.js
+++ b/app/component/map/Map.js
@@ -198,7 +198,7 @@ class Map extends React.Component {
             tileSize={config.map.tileSize || 256}
             zoomOffset={config.map.zoomOffset || 0}
             updateWhenIdle={false}
-            size={(config.map.useRetinaTiles && L.Browser.retina) ? '@2x' : ''}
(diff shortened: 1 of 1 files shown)


hannesj:
Is there some reason for this? The config option is there to indicate that the map server supports both regular and retina tiles, as the not all providers supply both.

Eightyplus:
The default configuration for useRetinaTiles is true, so right now the browser decides whether to display larger map tiles. We think the map looks way better by forcing 2x, and would like to share this with you.

hannesj:
Sorry that I haven't replied. I just don't really understand the point of the option, as the tiles supplied should be of the same quality as long as the ppi matches. 

I checked your site, and it seems you are using 256px tiles, which means you should be using the following options in your config instead of the supplied:

`
map.tileSize: 256,
map.zoomOffset: 0
`

Eightyplus:
Thanks this actually solves the resolution problem and not the pull request.

PR #1405, opened 2017-03-02
fixes drop-off so they are not interpreted as destination/arrival

Eightyplus:
Some busses in Norway has dropoff only (picktup=NONE) but it is not end destination. Thus these were displayed wrong. 
Adds a check to verify it is the last stop, texts are changed accordingly. 

Trosterud is a drop-off point for many busses
screen shot 2017-03-02 at 14 50 07

Radisson Blu Scandinavia is a endpoint for one of these busses
screen shot 2017-03-02 at 14 50 24

Maybe we should differentiate on the graphics as well?


Eightyplus:
Added svg/icon for drop-off points. 
screen shot 2017-03-03 at 08 49 41

tuukka:
Imporant feature but I don't understand the logic in the proposed icon.

Eightyplus:
It is suppose to symbolize exit. I google "exit icon" and most visualize a door with an arrow from it. Tried to keep it in the same fashion, but hey I'm not a designer, so I understand the confusion :)

Eightyplus:
Graphics updated a bit.
screen shot 2017-03-07 at 14 22 42

tuukka:
The old icon communicates that something ends here. I think it's ambiguous whether it's the whole vehicle journey or only some passengers' leg. The good part: at least it's clear nothing starts here.

I think the problem with the proposed icon is that it looks like something would start here.

How about a new variant of the old icon: leave a gap (white border) on the right of the stop circle, then continue the line after that for a bit? "Something ends here, something continues, nothing starts here."

Eightyplus:
We decided the icon is adequate for our needs and makes sense for our users. You are welcome to change the icon for HSLdevcom, then we can style it differently in our theme setup.

ttsirkia:
I think this is very nice feature. Actually, it is quite confusing that these two cases are currently indicated with the same symbol. I can see Tuukka's point about the symbol. How about if the symbol is just an arrow pointing down?

Something like this:
kuva

The arrow indicates stepping down from the vehicle. Because the arrow is vertical instead of horizontal, it shouldn't mix up with the other symbols.

Eightyplus:
Guess that makes sense as well 👍

PR #1427, opened 2017-03-06
Class on small icons

Eightyplus:
Makes icon size on stop markers configurable + sets appropriate classname on icon with cx to be able to style these individually.


siren:
https://github.com/HSLdevcom/digitransit-ui/pull/1450

PR #1437, opened 2017-03-08
Make intro views more accessible for screen reader users

RonjaO:
With a little fizes the intro views are more accessible for screen reader users. 

I add "a breadcrumb" (now a span element like x/y) which is hidden for sighted users by CSS but screen readers see it. It tells to blind screen reader users information of their location on intro views. Dots, which get the same information for sighted users, don't work for the blind and that's why I also hidden the dots for screen readers using aria-hidden attribute.

The content of intro views (title and text) are inside a button element. Screen readers read them just one element, a button, and don't find any headings from the page. I add the main role (role="main") for the button element: that's way screen readers find both, heading and text right.

### DIFF
--- a/app/component/Intro.js
+++ b/app/component/Intro.js
@@ -43,8 +43,9 @@ export default class Intro extends React.Component {
       key={i}
       tabIndex={0}
       onClick={this.onNextClick}
+      role="main"
     >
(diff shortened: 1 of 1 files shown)


siren:
created https://github.com/HSLdevcom/digitransit-ui/pull/1451

PR #1440, opened 2017-03-09
NRP-1270: fix updates set in production the same day

Eightyplus:
Change caching version from 'CurrentDate' to 'hash', which fixes a problem with multiple deployments the same day (hotfixes etc.).

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -257,6 +257,7 @@ function getPluginsConfig(env) {
       AppCache: {
         caches: ['main', 'additional', 'optional'],
       },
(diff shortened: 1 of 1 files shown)


Eightyplus:
(solves a problem where css files are cached due to a 304 from the server, eventough the css file has been changed)

Eightyplus:
fixes problem in Firefox and Safari (+iOS), but not sure about Chrome - which always return status 200 instead of 304

siren:
created https://github.com/HSLdevcom/digitransit-ui/pull/1449

PR #1442, opened 2017-03-09
makes max and min zoom configurable

Eightyplus:
Makes it possible to configure maxZoom (and minZoom) on map. React-leaflet has an error, so zoom has to be configured on <TileLayer as well.


siren:
created https://github.com/HSLdevcom/digitransit-ui/pull/1448

siren:
merged

PR #1451, opened 2017-03-10
Make intro views more accessible for screen reader users

siren:


siren:
The build is currently failing with the following lint error:

`/var/snap-ci/repo/app/component/Intro.js
  79:47  error  Elements with ARIA roles must use a valid, non-abstract ARIA role  jsx-a11y/aria-role`

hannesj:
Intro screens are not used anymore, closing

PR #1452, opened 2017-03-10
uses pelias as lookup for stops, optional and with configurable mapping

Eightyplus:
Rutebanken uses its own version of Pelias (open source) to search for stops. To map the modality of pelias responses, we map these in the search. The new setup is configurable as well as the mapping. 
We added two methods in config.*.js that handle some special cases, such as mapping of
- stop.properties.layer
- localization (we need to map municipality with a function to add this in the label/name).

If you don't like functions in the config file, we should be able to find a solution to this :)


vesameskanen:
Could you consider adding a suitable rutebanken configuration where the missing parts are defined? It would serve as a useful example for other UI users who might wish to search stops from pelias.

Eightyplus:
Do you want configuration examples in `config.default.js` then?

vesameskanen:
Hmm, how about config.rutebanken.js ?

Eightyplus:
We could add our configuration, but when we would need to do pull requests to do changes to the configuration. Alternatives:
1. Add config to defaults file
2. Add a examples file (with comments?)
3. Add documentation to pelias-to-modality-mapper.js

vesameskanen:
Hi @eightyplus, I guess option 3 is quite OK.  This is not a big issue after all.

Eightyplus:
@vesameskanen, Added, let me know if something is unclear and needs more/better text.

PR #1464, opened 2017-03-13
DT-1893: Added instructional image

lapim:
Jira Link: https://digitransit.atlassian.net/browse/DT-1893
Zeplin link: https://app.zeplin.io/project.html#pid=57c6a64d2d48014e718dec4a&sid=584812367fd449e002f94f03

Added the instructional image to the favourites panel.


app/component/no-favourites-panel.scss
```diff
@@ -0,0 +1,35 @@
...
+
+@media (max-width: 900px) {
+  .nofavs-img {
+    height: 240px;
+    width: 100%;
+    max-width: 240px;
+  }
+}
+
+@media (max-width: 380px) {
```
rrudd: Is there an explicit reason to start using media queries in steps instead of the .bp-s and .bp-l classes for major breakpoint and just scale the content within .bp-s? If there is, should we move to using this practice everywhere?

app/component/front-page.scss
```diff
@@ -92,6 +92,12 @@ $heading-color: #858585;
       padding-top:10px;
       background: #eef1f3;
     }
+
+    .nofavs.row {
+      background: $white;
+      font-size: 13px;
```
hannesj: Is there some reason to use $font-size variable?

PR #1468, opened 2017-03-13
NRP-1280: show platform code/number in itinerary search

Eightyplus:
Adds platform code/number to itinerary search. Configurable.


Eightyplus:
I added minor refactor of the component PlatformNumber to be able to show short and not short version of platform text.

Eightyplus:
I realize you use this for terminals, but it would also be nice to have this on itinerary search where people need a fulfilling description. Epesially train stations.
Our data is unfortunately not mature enough to indicate terminal status.

hannesj:
I think this has been an issue on our side also, I'll look into this tomorrow.

Eightyplus:
Looks like this (platform number right next to "Oslo S>"
screen shot 2017-03-14 at 08 03 20

Eightyplus:
Something that relates to this is platform number on departure list, again we would like this on Rail, but does not have information regarding terminal. So a temporary commit on my behalf is the code below. I should have made it configurable, but that was a lot of work for a function (not class) without context. Should we consider this as well?

app/component/Departure.js
```Javascript
  if (props.isTerminal || mode === 'rail') {
    platformNumber = <PlatformNumber number={props.departure.stop.platformCode} />;
  }
```

screen shot 2017-03-14 at 08 04 14

tuukka:
@Eightyplus Rather than base the visibility on mode, I'd configure whether to show the platform only at terminals or everywhere the platform number data is available.

Eightyplus:
I think we have an issue with OTP, which returns the last platform added no matter what the stop has been imported with. I'll keep you updated.

hannesj:
This will be merged from https://github.com/HSLdevcom/digitransit-ui/pull/1596

PR #1504, opened 2017-03-27
Update dockerfile to use yarn instead of npm

vesameskanen:


hannesj:
Waiting for https://github.com/docker-library/official-images/pull/2950

PR #1509, opened 2017-03-30
DT-1817

siren:
... DT-1911, DT-1912

### DIFF
--- a/app/component/HSLAdformTrackingPixel.js
+++ b/app/component/HSLAdformTrackingPixel.js
@@ -1,0 +1,8 @@
+import React from 'react';
+
+const HSLAdformTrackingPixel = () => (
+  <img src="https://track.adform.net/Serving/TrackPoint/?pm=785700&ADFPageName=K%C3%A4yneet%20sivustolla&ADFdivider=|" width="1" height="1" alt="" />);
+
+HSLAdformTrackingPixel.displayName = 'HSLAdformTrackingPixel';
+
+export default HSLAdformTrackingPixel;
--- a/app/component/HSLAdformTrackingPixel.js
+++ b/app/component/HSLAdformTrackingPixel.js
@@ -1,7 +1,7 @@
 import React from 'react';
 
 const HSLAdformTrackingPixel = () => (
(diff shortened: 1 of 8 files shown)


app/component/TopLevel.js
```diff
@@ -41,6 +43,21 @@ class TopLevel extends React.Component {
...
+    configureMoment(context.intl.locale, context.config);
+    const host = context.headers && (context.headers['x-forwarded-host'] || context.headers.host);
+    const url = context.url;
+
+    const hasTrackingPixel = get(context, 'config.showAdformTrackingPixel', false);
+    this.trackingPixel = host && host.indexOf('127.0.0.1') === -1 && host.indexOf('localhost') === -1 && hasTrackingPixel ? <HSLAdformTrackingPixel /> : undefined;
+
+    this.metadata = meta(this.context.intl.locale, host, url, this.context.config);
+
+    this.topBarOptions = Object.assign({}, ...this.props.routes.map(route => route.topBarOptions));
```
hannesj: this should be done on each render, as the props change during navigation

app/component/HSLAdformTrackingPixel.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+
+const HSLAdformTrackingPixel = () => (
+  <div style={{ width: 0, height: 0 }} ><img id=""src="https://track.adform.net/Serving/TrackPoint/?pm=785700&ADFPageName=K%C3%A4yneet%20sivustolla&ADFdivider=|" width="1" height="1" alt="" /></div>);
```
hannesj: missing space between id and src

app/component/MessageBarMessage.js
```diff
@@ -0,0 +1,42 @@
...
+};
+
+const a = (e) => {
+  if (e.type === 'a') {
+    return (<a href={e.href}>{e.content}</a>);
+  }
+  return null;
+};
+
+const elements = [heading, span, a];
```
hannesj: We could use https://github.com/acdlite/recompose/blob/master/docs/API.md#componentfromprop for these

app/component/TopLevel.js
hannesj: or be changed in componentWillUpdate

Issue #537, opened 2016-04-28
Digitransit should use GTFS route_color in itinerary summary, details and on the map


kalon33:
Using official line colors in Digitransit as in GTFS route_color should make it more intuitive to know which lines you are travelling with. It would also be great to provide an override when no color is set in the GTFS. 
BTW, it would also comply with transit authority color schemes, that the user is used to find in everything relative to the transit lines.

tuukka:
This should be easy enough for you to implement yourself for Paris: in our GraphQL API, routes have `color` and `textColor` that you could add to the queries that currently include `mode`.

AlexLoyko:
Would you like help with this?

kalon33:
@AlexLoyko Why not please, I'm currently busy with other things for my work (far away from transit stuff), and I don't have enough time regarding my skills to understand how the code would work...

AlexLoyko:
I am working on this issue, and faced a problem that once I enter "color" as one of the route's attributes in the request, it returns null. So, I went ahead and figured out where the color is set in the scss. That was the _helper-classes.scss to which I have access to any DOM element. However, it just does not feel right to search through all the components to find every line segment and set color(or add an extra property to it to achieve this issue). My feeling is that we could just modify the QL request with the color attribute for a planner that returns the actual color and set it. Am I right regarding the GraphQL part or I need to search through every element of the UI and add an extra prop color which I set based on my mode(subway in our case) and subway code by updating the _helper-classes.scss?
Thanks in advance!

AlexLoyko:
To make it shorter and more clear, what would be the best way to update colors programmatically, either as a component prop in a Relay store or some other way?

tuukka:
@AlexLoyko The schema looks correct (type Route with fields color and textColor). I don't see a reason to change it. The backend implementation is here: https://github.com/HSLdevcom/OpenTripPlanner/blob/master/src/main/java/org/opentripplanner/index/IndexGraphQLSchema.java#L1610

Where the UI currently queries GraphQL for Route mode and shortName or puts them into props, I think it should additionally handle the colors. When rendering DOM, the code can set element inline styles, if the colors are not null. If the colors are null, there is no need to set inline styles and theme scss should apply as it does now.

There may be additional complications that I'm not taking into account though.

AlexLoyko:
Thank you very much! On it.

AlexLoyko:
@tuukka 
[...]

PR #1513, opened 2017-04-02
Implemented line colors to match the GTFS colors

AlexLoyko:
In this commit I have implemented the feature described in #537. 
A few comments: shouldn't this line https://github.com/HSLdevcom/digitransit-ui/blob/master/app/component/TransitLeg.js#L98 be a RouteNumberContainer instead?
Is there enough testing or should I work on tests myself?"


laidig:
A hint from the docs:
https://github.com/HSLdevcom/digitransit-ui/blob/master/docs/Installation.md#configure-git-hooks

`npm run test-local` will run the tests locally.

kalon33:
@AlexLoyko Would it also be possible to change line number color and the color of the shape containing that number on the map to comply GTFS data? Thanks :)

AlexLoyko:
It does it already I assume.

AlexLoyko:
Just can't push it yet, get the BroserStack credentials missing. Any idea how to fix this quickly?)
Thanks:)

kalon33:
@AlexLoyko Apparently not, I just tested it locally:
screenshot_2017-04-06-21-47-39

As you can see, the route has the proper color on the map, but the "tag" with the line number on the map doesn't have the proper color (around the tag, and the line number text).

AlexLoyko:
Sure. I'll fix it ASAP.

PR #1516, opened 2017-04-03
Don't cancel user's search interaction by a geolocation change

vesameskanen:
Delayed response for setting user's initial geolocation no longer closes the splash screen, if user has already started defining a location using the search modal.
- If user enters an address, it will be used as the start point
- If user closes the search modal without entering an address, geolocation will be used immediately after closing the modal.

### DIFF
--- a/app/action/EndpointActions.js
+++ b/app/action/EndpointActions.js
@@ -44,7 +44,7 @@ export function setEndpoint(actionContext, payload) {
 }
 
 export function setUseCurrent(actionContext, payload) {
-  actionContext.dispatch('useCurrentPosition', payload.target);
+  actionContext.dispatch('useCurrentPosition', payload);
   return actionContext.executeAction(route, payload);
 }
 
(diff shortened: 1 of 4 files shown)


hannesj:
This should be already be handled by https://github.com/HSLdevcom/digitransit-ui/blob/b62780d53b5eff0d1bbf73eddec0e576ea18f6a7/app/component/GeopositionSelector.js#L23 and check why searchModalIsOpen doesn't work there.

vesameskanen:
App state is changed using async executeAction, so it is quite possible that a fast user can open the search modal while setUseCurrent is still being processed.  Indeed, in my tests the behavior was that search modal disappeared while I was was typing an address. This bad behavior is now prevented at SplashOrChildren component, where open modal state keeps the splash alive.

It is a mystery why user's selection for origin does not overwrite the setUseCurrent location, which supposedly finishes before the user picks a location. Location selection launches a setEndpoint action - how can it disappear without any effect?   To tackle this dilemma, setUseCurrent now has an option to prevent accidental overwrite in case the action order gets unexpectedly swapped.

When testing the bug with deliberately delayed geolocation callback,  both firefox and IE behaved as expected: after closing the search modal, quite often UI moved to current location and then immediately to selected location, which is a minor glitch. This demonstrates how setUseCurrent action and setEndpoint actions follow each other.

PR #1529, opened 2017-04-07
Implemented line colors and leg markers to match the GTFS colors

AlexLoyko:
Following the pull request #1513 now I've closed that one and squashed everything into one commit.

@kalon33 
The route tags' colors on the map are now matching the GTFS data as well:)

Since I am not a part of HSL team, I won't pass the credentials for the gemini test. 
So, I'll be waiting on your review, gentlemen!


kalon33:
@AlexLoyko Just found out that in itinerary details, when displaying stops list for a line, the color is not correct (see the following screenshots). When stops are hidden, color is good.
details2
details1

But everywhere else it seems good :).

kalon33:
@alexloyko just activated disruptions info on my digitransit instance, and route colors are not used for it. Could you make a PR for this please? Thanks :)

PR #1554, opened 2017-04-21
Implemented imperial measurements

AlexLoyko:
Implemented the silly American system that the British and Canadians do not even use anymore:)

### DIFF
--- a/app/component/BicycleLeg.js
+++ b/app/component/BicycleLeg.js
@@ -8,9 +8,9 @@ import ComponentUsageExample from './ComponentUsageExample';
 import { displayDistance } from '../util/geo-utils';
 import { durationToString } from '../util/timeUtils';
 
-function BicycleLeg(props) {
+function BicycleLeg(props, context) {
   let stopsDescription;
-  const distance = displayDistance(parseInt(props.leg.distance, 10));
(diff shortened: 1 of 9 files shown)


vesameskanen:
You should use the dynamically assigned configuration which is available as context.config, see for example SearchModal.js.  The config (or distance type) should be passed as a parameter to displayDistance func, not by importing a static default configuration in geoutils.

vesameskanen:
I'm afraid we still need the dynamic configuration option or an explicit UI setting to select imperial measuring. We used to get lots of complaints when the application chose the UI language automatically based on browser settings;  the same applies to measuring units. So, we have to stick to metric system unless  configured otherwise.

laidig:
We also have gotten a fair number of complaints, as Imperial is primarily
used here. It's common to implement some localization for units and number
formatting based upon locale, and for that reason it makes sense to set
this for en_US, but I understand your point that it should be configurable.

Do you have a suggestion for where to implement this configurability?

vesameskanen:
I added the top level switch which allows imperial measuring to the dynamic configuration. It is called config.imperialEnabled. 

I noticed that routing options / walk speed slider still shows speed as meters/minute; you might want to update that, too

vesameskanen:
merged by https://github.com/HSLdevcom/digitransit-ui/pull/1604

PR #1571, opened 2017-05-05
DT-1879: As a user I expect the time input field to always be in 24h format

lapim:
Replaces the native time input/plain text field input with a time picker that works the same for all browsers.


ttsirkia:
Seems to make user experience much worse with mobile browsers.

ttsirkia:
If I erase the minutes, it will display `00`. However, after losing focus, it becomes `45`. It can't work like this. Also, it is very unintuitive that I cannot remove the digits but I have to type nyw digits which will overwrite the existing digits. As entering the destination and time are the most important inputs, usability should be very high and I cannot see how this PR made the situation better.

PR #1589, opened 2017-05-11
Support for imperial measurements

vesameskanen:


AlexLoyko:
@vesameskanen There is a problem with the Distance.js file. Config got wrapped twice, so isImperial never returns true when called from the Distance.js. https://github.com/HSLdevcom/digitransit-ui/blob/imperial/app/component/Distance.js#L18

PR #1601, opened 2017-05-16
DT-1750

siren:


hannesj:
If you have an old search with a stop as result, you will get an invalid stop code in the timetable link

kalon33:
Dear @hannesj @siren How should we alter config.default.js to still be able to use mapzen as PELIAS and PELIAS_REVERSE_GEOCODER provider in Digitransit as before?

PR #1602, opened 2017-05-17
DT-1906: As a user I want to see itinerary's stops on the map with next departures and be able go stop timetables

lapim:
Opens the stop details popup when it's clicked on the itinerary view. When using on mobile, opens the popup only if the view is in fullscreen map mode. If the view returns to the minimized state, the popup closes.


app/component/TransitLeg.js
```diff
@@ -28,6 +28,7 @@ class TransitLeg extends React.Component {
 
   renderIntermediate() {
     if (this.props.leg.intermediateStops.length > 0 && this.state.showIntermediateStops === true) {
+      console.log('row 31');
```
hannesj: Remove debug logging

PR #1603, opened 2017-05-17
Timepicker: Added automatic focus to the minute input after hours are set

lapim:


ttsirkia:
If it works like this, then using backspace should also jump from minutes to hours. For example, typing first `12:50` and then replacing it with `13:10`. Input type should also be `number` to view the numerical keypad in mobile browsers. Still, this cannot beat the native timepickers in mobile browsers.

lapim:
Thanks for the comment! Here are some thoughts regarding your points.

Input type: Though `number` would  technically be the correct input type and enable the use of the numeric keypad on mobile devices, it creates a problem with other requirements. `number` doesn't support certain functions that we want to use (for this reason we're using `setSelectionRange()` instead of `select()` to support iOS browsers) and its crossbrowser support isn't as comprehensive as input `text`'s is. 

Native `time` input: Some browsers we want to support (such as Firefox) don't support `time` and the ones that do get their 12h/24h clock set up by their OS. As we want to have a timepicker that works the exact same way for all browsers (i.e. forced 24h clock that has been requested multiple times), we unfortunately can't utilize the native browser timepicker for the time being.

PR #1606, opened 2017-05-22
Use travis stages

vesameskanen:


test/deploy.sh
```diff
@@ -0,0 +1,23 @@
+#/bin/bash
+set -e
+ORG=${ORG:-hsldevcom}
+yarn install
+
+docker run -d -e CONFIG=hsl -p 127.0.0.1:8080:8080 $ORG/digitransit-ui:ci-$TRAVIS_COMMIT
+wget -N http://chromedriver.storage.googleapis.com/2.29/chromedriver_linux64.zip
```
hannesj: Why do we need chromedriver on deploy?

hannesj: Oh, we are actually running tests here, could those be run in parallel with the visual tests in one stage?

vesameskanen: Sure. I will update that part.

PR #1617, opened 2017-05-24
Timetable component & container

ktj:
Created timetable component and added it to stop page in timetables tab
https://zpl.io/Z1gtg49 & https://zpl.io/2wTJhq


ktj:
Large screen
timetable

Small screen
timetable_small

PR #1635, opened 2017-06-01
DT-2027 Allow selecting date for timetables

hannesj:


vesameskanen:
There seems to be a bug. The selected date changes back to today after a while; sometimes sooner, sometimes later. Maybe related to some realtime update event.

PR #1639, opened 2017-06-02
DT-1963 Do not touch arriveBy selector when clicking earlier/later

hannesj:


vesameskanen:
There is one issue which I find problematic; after applying later/earlier buttons, which swap the arriveBy state, time selector stops obeying  the user. If the user enters a new 'arrive by' time, the internal state should be reset back to obey the displayed arriveBy condition.

PR #1666, opened 2017-06-15
Add commit id and build time to docker image

siren:


siren:
includes https://digitransit.atlassian.net/browse/DT-2040

@hannesj can you please check the excludes I added

PR #1668, opened 2017-06-15
Uncaught empty geometry

AlexLoyko:
We found that we could not load pattern.geometry sometimes which caused calling filter on an undefined object. The necessary check is added.


laidig:
specific exception was:  
```
RouteMapContainer.js:65 Uncaught (in promise) TypeError: Cannot read property 'filter' of undefined
    at n (RouteMapContainer.js:65)
```

PR #1823, opened 2017-09-12
DT-2009: As a user I want to print a stop's timetable

lapim:
JIRA Link: https://digitransit.atlassian.net/browse/DT-2009
Zeplin Link: https://app.zeplin.io/project/57ea6b227dcc263b567e1e6f/screen/598990de4cad74c5de0a7d93


app/component/SecondaryButton.js
```diff
@@ -6,7 +6,8 @@ import Icon from './Icon';
 
 const secondaryButton = (props, context) =>
   <button
-    className="secondary-button"
+    className={`secondary-button ${props.buttonParams.buttonName} ${props
```
hannesj: Why not use classnames library, as is done in other places? [example](https://github.com/HSLdevcom/digitransit-ui/blob/c358a9f99f217aa129d50650d489281db6cda3e3/app/component/map/Line.js#L50)

app/component/StopPageActionBar.js
```diff
@@ -19,7 +24,17 @@ const StopPageActionBar = ({
         dateFormat={DATE_FORMAT}
         onDateChange={onDateChange}
       />
-      <PrintLink className="action-bar" href={printUrl} />
+      <div className="print-button-container">
+        <SecondaryButton
+          buttonParams={{
+            ariaLabel: 'print',
+            buttonName: 'print',
+            buttonClickAction: e => printStop(e),
```
hannesj: why not just `buttonClickAction: printStop`

hannesj: or just window.print, is there a reason for the wrapper function?

app/component/StopPageActionBar.js
```diff
@@ -19,7 +24,17 @@ const StopPageActionBar = ({
         dateFormat={DATE_FORMAT}
         onDateChange={onDateChange}
       />
-      <PrintLink className="action-bar" href={printUrl} />
+      <div className="print-button-container">
+        <SecondaryButton
+          buttonParams={{
```
hannesj: Is there a reason why all params are wrapped in an object?

app/component/stop.scss
```diff
@@ -431,10 +441,75 @@ hr.action-bar {
...
+      font-size: 20px;
+      font-family: $font-family;
+    }
+  }
+}
+
+.print-button-container {
+  display: flex;
+  button {
+  margin-right: 1em;
```
hannesj: The indentation seems off, same in a couple of places.

app/routes.js
```diff
@@ -607,7 +607,7 @@ export default config => {
               .then(content => cb(null, { content: content.default }))
               .catch(errorLoading);
           }}
-          printPage
+          printPageprintPage
```
hannesj: ??

lapim: Probably a copypaste accident. I hope. :-1:

app/component/StopPageActionBar.js
lapim: No particular reason, just a preference of having a generic component getting the props as a "package" instead of multiple pieces for maintenance reasons.

PR #1827, opened 2017-09-13
DT-2127 Remove SplashOrChildren

hannesj:
TODO: Add default starting point for all configs


app/component/map/MapWithTracking.js
```diff
@@ -72,102 +72,120 @@ const onlyUpdateCoordChanges = onlyUpdateForKeys(
...
+      origin.lat != null &&
+      origin.lon != null
+    ) {
+      location = origin;
+      zoom = 16;
+    } else if (mapTracking && position.hasLocation) {
+      location = position;
+      zoom = 16;
+    } else {
+      location = context.config.defaultMapCenter;
```
vesameskanen: Could we use here defaultEndpoint if map center is not defined? It could be the right choice for most configurations.

vesameskanen:
Really many tests seem to fail - is this expected, or is there a problem in the test system itself?

PR #1961, opened 2017-11-20
Dt 2259: Front page placeholder "Lähtöpaikka, Linja tai pysäkki" text should be darker

lapim:
JIRA Link: https://digitransit.atlassian.net/browse/DT-2259


vesameskanen:
- install script is broken
- no visible change in placeholder text darkness (tested with with firefox)

PR #1957, opened 2017-11-16
DT-2262: Add back Lahellasi tab gradient (and make bottom white area smaller) 

lapim:
JIRA Link: https://digitransit.atlassian.net/browse/DT-2262

Made other fixes to the table structure to make the effect work better

PR #1963, opened 2017-11-20
DT-2252: Add blurring to initial map (same that you have in the current version of static map)

lapim:
JIRA Link: https://digitransit.atlassian.net/browse/DT-2252


siren:
Because this is based on #1957 the same problem is present here too...

https://github.com/HSLdevcom/digitransit-ui/pull/1957#issuecomment-345688466

PR #2003, opened 2017-11-30
DT-2319: When there is a notification open the layout breaks on Desktop and Mobile mode

lapim:
JIRA Link: https://digitransit.atlassian.net/browse/DT-2319

### DIFF
--- a/app/component/TopLevel.js
+++ b/app/component/TopLevel.js
@@ -141,6 +141,17 @@ class TopLevel extends React.Component {
     }
 
     const menuHeight = (this.getBreakpoint() === 'large' && '60px') || '40px';
+    // Check which view we are on to decide wether to use flex or grid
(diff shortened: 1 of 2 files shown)


app/component/TopLevel.js
```diff
@@ -141,6 +141,17 @@ class TopLevel extends React.Component {
     }
 
     const menuHeight = (this.getBreakpoint() === 'large' && '60px') || '40px';
+    // Check which view we are on to decide wether to use flex or grid
+    // Index page uses grid - flex for everything else
+    const mainContentDisplay =
+      this.props.location.pathname.indexOf('lahellasi') !== -1 ||
+      this.props.location.pathname.indexOf('suosikit') !== -1
+        ? {
+            display: `grid`,
```
hannesj: This is not supported on IE11, should probably be flex?

PR #2004, opened 2017-11-30
Navigation fix for addresses with special chars

siren:
For example forward slash is breaking navigation currently


siren:
should also fix DT-2340

siren:
and DT-2341

PR #2027, opened 2017-12-05
Minor fixes

hannesj:

PR #2028, opened 2017-12-05
DT-2303 Make Map LazyLoadable

hannesj:
Based on #2027


kalon33:
Sorry I should have written there: @hannesj @vesameskanen  I used to use non vector tiles from thunderforest as advised by another DT dev. So I've no map anymore. How could I fix that problem?

Thanks for your help.

PR #2090, opened 2018-01-08
DT-2384: Bus and tram realtime location icons are too small

lapim:
JIRA Link: https://digitransit.atlassian.net/browse/DT-2384


vesameskanen:
Would it be possible to center the icons on top of the route? For some reason, the middle point of an icon does not match the route line.

PR #2097, opened 2018-01-11
DT-2402: Openable/closable quick settings view from "Settings" button

lapim:
JIRA Link: https://digitransit.atlassian.net/browse/DT-2402
Zeplin Link: https://app.zeplin.io/project/57c6a64d2d48014e718dec4a/screen/59a402a41af13b2b98996f4c


vesameskanen:
This needs some tuning:
- Time selector and day selector are no longer equally high
- Close icon for expanded settings has different size as the close icon for route middle point just above
- Close icon for expanded settings is not aligned with other right side icons above
- On Chrome, open and close expanded settings shows a frame when in focus, which IMO looks a bit strange. Designs do not show such frame.  Maybe check from designers? Other summary icons/buttons don't show the focus.

PR #2110, opened 2018-01-16
DT-2385: When user tries to print an itinerary with only walking mode, the application crashes

vuorisam:
Fixed walk itinerary printing bug.


vesameskanen:
yeah, that's the right way to sync the changes.

PR #2180, opened 2018-03-02
DT-2499: Move Total travel time/Walking amount element to the end of the itinerary suggestion

lapim:
JIRA Link: https://digitransit.atlassian.net/browse/DT-2499


vesameskanen:
Time and walk distance are too close to route end time. There must be some margin space to separate the values.

PR #2187, opened 2018-03-05
Auto-name favourite based on the icon chosen

akx:
As it was, the user had to first enter a name for a favourite place, then choose a descriptive icon, which is a click and type-type-type too many.

This changes the flow to make the UI automagically name the place (unless a name has been manually entered) when the user selects an icon.

Translations for the places exist for Finnish and Swedish. The English names are used when translations are not available.

~I admit I didn't test the code~ Aside from linting, I couldn't test the code, because "OK. 0 tests passed. (4.165s)", so please be careful before merging into any sort of release branch. :sweat_smile:

(In addition, Travis apparently deliberately fails tests from third parties, which is... nice.)


vesameskanen:
Thanks for a nice contribution, I will merge this in the near future.

PR #2243, opened 2018-04-15
Stop autosuggest box from intersecting with search

aapokiiso:
Currently the autosuggest box is intersecting slightly with the search box:

image

The PR removes the negative margins causing this.

### DIFF
--- a/app/component/search.scss
+++ b/app/component/search.scss
@@ -397,7 +397,6 @@ div.autosuggest-panel {
         height: auto;
         position: absolute;
         width: 100%;
-        margin-top: -8px;
         display: block;
         overflow: visible;
         z-index: 1006;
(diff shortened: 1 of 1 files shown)


vesameskanen:
Thanks for contribution! There seems to be some layout problems with search dropdowns in in the itinerary page, which appears after the route search. The same elements are used also there.

Note also that desktop and mobile layouts/styles are a bit different, and the 'Add favorite' tool also uses the search elements. New styles should work with all these cases.

aapokiiso:
@vesameskanen Thanks, I'll try to re-work it

aapokiiso:
Tried to have another go at this, but it started taking too much time for what was meant to be a "drive-by" fix, sorry!

As a side-note, it looks like the top-margins for the [origin](https://github.com/aapokiiso/digitransit-ui/blob/e94797fc21560cfa6a521c56e4a8696a2366bbc2/app/component/search.scss#L446) and [destination](https://github.com/aapokiiso/digitransit-ui/blob/e94797fc21560cfa6a521c56e4a8696a2366bbc2/app/component/search.scss#L462) fields play a role here too.

PR #2246, opened 2018-04-17
DT-2071: As a user I want to be able to differentiate departures with the same route but different patterns in the stop timetable

lapim:
JIRA Link: https://digitransit.atlassian.net/browse/DT-2071
Can be tested with stops pysakit/HSL%3A1414102 and pysakit/HSL%3A4900203


kalon33:
After some use, it appears to me that it may difficult to identify how many stars you have at each stop time... Would it be possible to use numbers, or letters, as indices instead?

image

Thanks!

lapim:
Thanks for bringing this up! It definitely does create a big issue with readability, we'll look into possible solutions for figuring out how to fix this ASAP. Letters and numbers might be confusing as many routes have them included into their shortnames, but the initial idea sounds considerable.

kalon33:
Thanks for looking at this, for now, I found a maximum of 16 different destinations (so the case pictured here is really not the worst!), and 16 stars is even harder to differentiate from 15! ;)

PR #2279, opened 2018-05-16
DT-2551

vesameskanen:
- Working fare zone config (starts working when GTFS gets updated to our  new specs)
- Restore a temporarily disabled flow test

### DIFF
--- a/app/configurations/config.hsl.js
+++ b/app/configurations/config.hsl.js
@@ -286,6 +286,7 @@ export default {
     'HSL:seu2': 'HSL:seu',
     'HSL:seu3': 'HSL:seu',
     'HSL:seu4': 'HSL:seu',
+    'HSL:seu5': 'HSL:seu',
     'HSL:lse': 'HSL:lse',
     'HSL:lse2': 'HSL:lse',
     'HSL:lse3': 'HSL:lse',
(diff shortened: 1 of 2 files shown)


app/configurations/config.hsl.js
```diff
@@ -275,27 +275,41 @@ export default {
...
     'HSL:kse8': 'HSL:kse',
+    'HSL:kse9': 'HSL:kse',
+    'HSL:kse10': 'HSL:kse',
+    'HSL:kse11': 'HSL:kse',
+    'HSL:kse12': 'HSL:kse',
+    'HSL:kse13': 'HSL:kse',
+    'HSL:kse14': 'HSL:kse',
+    'HSL:kse15': 'HSL:kse',
+    'HSL:kse16': 'HSL:kse',
   },
```
jolkkola: Would it be feasible to just drop the numeric part of each identifier (i.e. 'HSL:kse10' -> 'HSL:kse')? Then this fareMapping would not be required to exist at all. Any new numeric extensions in the future would also work right away.

It also seems that the value of each pair is used as a key to retrieve the name of the ticket. Perhaps a unit test to check that the name exists for all possibilities would be in order.

We can revisit these ideas once the fare zones will be updated for the HSL area.

PR #2285, opened 2018-05-20
Update build pipeline

hannesj:
- Update dependencies
- Convert to Webpack 4
- Reduce bindle size by optimizing imports
- Move breakpoint handling to a new style React context
- Enable ASSET_URL env parameter


hannesj:
I'll take a look at it later today!

PR #2291, opened 2018-05-23
Fix the PWA install on iOS

jolkkola:
Some changes made in #2285 (or related libraries) broke the PWA install of Reittiopas on iOS. The purpose of this pull request is to fix the breaking change and update the status bar color.

### DIFF
--- a/app/localStorageHistory.js
+++ b/app/localStorageHistory.js
@@ -43,30 +43,32 @@ const REPLACE = entry => {
 const getEntries = () => history.entries;
 const getIndex = () => history.index;
 
+const getLocationListener = () => event => {
+  switch (event.action) {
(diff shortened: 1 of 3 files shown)


app/localStorageHistory.js
```diff
@@ -43,30 +43,32 @@ const REPLACE = entry => {
...
+    case 'REPLACE':
+      REPLACE(event);
+      break;
+    case 'PUSH':
+      PUSH(event);
+      break;
+    default:
+      console.error('unhandled history event:', event);
+  }
+  if (this && this[event.action] !== undefined) {
```
jolkkola: Most likely these lines (60, 61 and 62) could be removed altogether.

PR #2308, opened 2018-05-30
DT-2150

vesameskanen:
Adds itineraryFiltering parameter to route planning query. Currently the value is read from the configuration.


app/configurations/config.default.js
```diff
@@ -79,6 +79,7 @@ export default {
 
   maxWalkDistance: 10000,
   maxBikingDistance: 100000,
+  itineraryFiltering: 2.5, // drops 40% worse routes
```
jolkkola: What does this mean? Would a value of 5 be equal to 80% worse routes?

vesameskanen: This is documented better in OTP side and also in our gitlab wiki's  OTP parameter doc.

The rule is simply 100%/x . So factor 5 defines 100%/5 = 20%  filtering threshold.

PR #2348, opened 2018-06-13
DT-2682

vesameskanen:
Index page now monitors PreferencesStore language selection only via StoreListeningIntlProvider, which handles context.intl translations. In earlier versions, Index page connected also directly to PreferencesStore, and this lead to wrong event handling order and rendering the page with old translations. 

Also, language selector state was confused because of SSR rendering, which is now disabled.


jolkkola:
A more thorough look into using `react-intl` would be a good idea in the near future as it moves to use the new context API.

Issue #1594, opened 2017-05-11
Walking speed unit m/min is odd


ttsirkia:
Estimating your walking time as meters per minute is very hard because that kind of units is never used.

Most of the people know that 5 km/h is quite typical walking speed, 7 km/h requires quite fast walking and 3 km/h is rather slow. But I had to calculate that 5 km/h is 83 m/min. Didn't know that and cannot estimate any speeds using that unit.

Suggesting to change the unit.

jolkkola:
This has now been changed. The unit will change in the production version once the cycling & walking feature has been completed and merged to the master branch.

PR #2363, opened 2018-06-19
DCW-32: change the walking speed indicator from m/min to km/h

jolkkola:
The purpose of this pull request is to change the walking speed indicator from m/min to km/h. Note that no speed indicator is available for cycling speed, so the rest of the specified work must be done after the new settings panel becomes available.

Link to the JIRA task: https://digitransit.atlassian.net/browse/DCW-32

Screenshot: 
image


ttsirkia:
Fixes #1594

PR #2391, opened 2018-07-03
Fix firefox timetable printing

jolkkola:
The purpose of this pull request is to enable Firefox to print all timetable pages. The printing problem was due to display:flex being set to the container elements. This should not affect other browsers' printing capabilities.

This was previously part of the branch DT-2284, but that may be discarded altogether.

### DIFF
--- a/app/component/map/map.scss
+++ b/app/component/map/map.scss
@@ -62,6 +62,10 @@ div.leaflet-bottom.leaflet-left div.leaflet-control-zoom {
 
 .map .leaflet-bottom {
   z-index: 800;
+
+  @media print {
+    display: none;
+  }
 }
 
 div.leaflet-container {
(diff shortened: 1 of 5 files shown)


jolkkola:
This affects the route, line and timetable (stop/terminal) print modes.

Examples:
- route: http://localhost:8080/reitti/Opastinsilta%206%2C%20Helsinki%3A%3A60.199093%2C24.940536/Rautatieasema%2C%20Kaivokatu%2C%20Helsinki%3A%3A60.170356%2C24.941252/0/tulosta?modes=WALK%2CCITYBIKE%2CSUBWAY
- line: http://localhost:8080/linjat/HSL:1059/aikataulu/HSL:1059:0:01
- stop: http://localhost:8080/pysakit/HSL%3A1220106
- terminal: http://localhost:8080/terminaalit/HSL%3A1000004

PR #2397, opened 2018-07-04
Dt 2696

optionsome:
* Fix favicon and sw.js loading paths when not in root

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -32,8 +32,9 @@ function addMetaData(config) {
   }
 
   const html = stats.html.join(' ');
-  const appPathPrefix =
+  const APP_PATH =
     config.APP_PATH && config.APP_PATH !== '' ? `${config.APP_PATH}'/'` : '/';
+  const appPathPrefix = process.env.ASSET_URL || APP_PATH;
 
   htmlParser.convert_html_to_json(html, (err, data) => {
     if (!err) {
@@ -57,6 +58,11 @@ function addMetaData(config) {
       data.link.forEach(e => {
         // eslint-disable-next-line no-param-reassign
         delete e.innerHTML;
+        if (e.href.startsWith('../icons')) {
+          console.log(e.href);
+          console.log(appPathPrefix);
+          e.href = appPathPrefix + e.href.substring(3);
+        }
       });
 
       // eslint-disable-next-line no-param-reassign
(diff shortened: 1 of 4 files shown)


app/component/AppBarLarge.js
```diff
@@ -25,12 +29,15 @@ const AppBarLarge = ({ titleClicked }, { router, location, config, intl }) => {
     <div>
       <div className="top-bar bp-large flex-horizontal">
         <button className="noborder" onClick={titleClicked}>
-          {config.textLogo ? (
+          {isBrowser && !config.textLogo && logo ? (
```
hannesj: Isn't this wrong, as it makes SSR render the text logo, rather than just having an empty background image?

app/component/AppBarSmall.js
```diff
@@ -6,18 +6,19 @@ import DisruptionInfo from './DisruptionInfo';
...
   { config },
 ) => (
   <React.Fragment>
     <DisruptionInfo />
     <nav className="top-bar">
       {!disableBackButton && <BackButton />}
       <section className="title">
-        {showLogo && !config.textLogo ? (
-          <div className="logo" />
+        {isBrowser && showLogo && !config.textLogo && logo ? (
```
hannesj: Same as above

app/component/ItineraryCircleLine.js
```diff
@@ -15,6 +16,18 @@ class ItineraryCircleLine extends React.Component {
     color: PropTypes.string,
   };
 
+  constructor(props) {
+    super(props);
+    this.state = {
+      isMounted: false,
```
hannesj: No need to have isMounted. Instead, you should do an import('...png').then(imageUrl => this.setState({imageUrl})). Also state can be initialised directly using a class property initializer.

sass/base/_elements.scss
```diff
@@ -195,16 +195,19 @@ $border-radius-rounded: 30px;
 .overflow-fade {
   overflow: hidden;
   mask-image: linear-gradient(to left, transparent, rgba(0,0,0,0.2) 5px, rgba(0,0,0,1) 25px, rgba(0,0,0,1));
+  -webkit-mask-image: linear-gradient(to left, transparent, rgba(0,0,0,0.2) 5px, rgba(0,0,0,1) 25px, rgba(0,0,0,1));
```
hannesj: This is not needed, as it is added by autoprefixer

optionsome:
Created another pull request (https://github.com/HSLdevcom/digitransit-ui/pull/2421) that includes suggested fixes

PR #2418, opened 2018-07-20
Increase walkBoardCost for Oulu

optionsome:
* small refactoring to allow different defaultSettings for configs
* increase walkBoardCost for Oulu


jolkkola:
Most of this functionality will need to be checked and possibly reimplemented once the new settings panel is in use. The handling and combining of different settings should also be unified, with the defaults declared in the configuration file.

PR #2443, opened 2018-08-13
DCW-29: As a user I want to be notified when I cannot take my bike to the selected public transport

lapim:
JIRA Link: https://digitransit.atlassian.net/browse/DCW-29


vesameskanen:
Route search Pasila-Puistola opens the error page.

PR #2469, opened 2018-09-05
Show arrival time of transit legs at intermediate stops

vesameskanen:
- Change the query to use place ( not stop ) type
- Render arrival time


pailakka:
Included in https://github.com/HSLdevcom/digitransit-ui/pull/2472

PR #2471, opened 2018-09-05
DCW-25: Promoting walking and cycling to the users

lapim:
JIRA Link: https://digitransit.atlassian.net/browse/DCW-25


jolkkola:
This is now included in #2486.

PR #2486, opened 2018-09-13
DCW-54: The promotion suggestion bar loads with a lag 

lapim:
JIRA Link: https://digitransit.atlassian.net/browse/DCW-54


jolkkola:
Will revisit this when there is support available on the OTP side. Currently the user experience remains somewhat lacking and, in many cases, out-of-sync.

PR #2491, opened 2018-09-20
DT-2735

vesameskanen:
Consider defined layer preferences of all kind of search items (old searches, favourites etc) in all cases. Intermix geocoder's rating with UI defined rating to achieve optimal sort order.


vesameskanen:
Summary of latest changes:
- Finer resolution in mixing geocoded and local items
- Deduplicate ordered result list
- Match rating considers scandic letters like the backend (vaari is not vääri)
- Don't store any favourite selections to old searches

PR #2502, opened 2018-09-29
French translations update

kalon33:
Update French translations to cover latest changes.


kalon33:
I can't run visual tests, so these fail with "BS_ACCESS_KEY not set"

PR #2517, opened 2018-10-11
DT-2289

vesameskanen:
- GeoJson interface. 
- New HSL zones as a geoJson layer


jolkkola:
Things to consider when further developing this feature:
- performance with large datasets
- per-language popup customization

PR #2520, opened 2018-10-12
Made default font sizes slightly bigger in near you table

optionsome:
Should make google font look slightly more similar to the HSL on in the near you table


optionsome:
Also tried changing font-weight but effects from changing it were too drastic

PR #2537, opened 2018-10-23
Update webpack dependencies

hannesj:
Makes installing easier on Node 10


optionsome:
This seems to cause slight increase in build times. On my computer it went up from ~110 s to ~200 s

PR #2556, opened 2018-11-03
Fix the clear-input action for autosuggest components

eemeli:
At the moment, clicking on the X icon at the right end of the autosuggest input box does not in fact clear the selection, but just focuses the input box. This is rather surprising. There is in fact no way of removing a selection once it's already been made. This is very frustrating especially when using the site as a homescreen app at least on Android, in which use case it never forgets a prior state.

This PR makes tapping the X icon actually clear the selection, by:
- Refactoring the behavious of clearItem() in DTSearchAutosuggest
- Supporting a `null` value for the item used in the DTEndpointAutosuggest, DTOldSearchSavingAutosuggest, and DTSearchAutosuggest
- Adding an optional `cleared` field to the location object, and using it where appropriate

As a related change, the IndexPage map will now show the destination location if the origin has been cleared but the destination is set. I've also added a unit test for the changes, for the DTEndpointAutosuggest component.

### DIFF
--- a/app/component/DTAutosuggestPanel.js
+++ b/app/component/DTAutosuggestPanel.js
@@ -167,9 +167,7 @@ class DTAutosuggestPanel extends React.Component {
 
   handleViaPointLocationSelected = (viaPointLocation, i) => {
     const { viaPoints } = this.state;
-    viaPoints[i] = {
-      ...viaPointLocation,
-    };
+    viaPoints[i] = viaPointLocation.cleared ? {} : { ...viaPointLocation };
     this.setState({ viaPoints }, () => this.updateViaPoints(viaPoints));
   };
 
@@ -307,7 +305,7 @@ class DTAutosuggestPanel extends React.Component {
             value={this.value(origin)}
             isFocused={this.isFocused}
             onLocationSelected={location => {
-              let newOrigin = { ...location, ready: true };
+              let newOrigin = { ...location, ready: !location.cleared };
               let { destination } = this.props;
               if (location.type === 'CurrentLocation') {
                 newOrigin = { ...location, gps: true, ready: !!location.lat };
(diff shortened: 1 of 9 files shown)


eemeli:
It looks like the CI visual tests are failing with this error:
```
$ test/visual.sh
BS_ACCESS_KEY not set, failing build on purpose
```
Presumably this isn't something that's due to my changes?

vesameskanen:
Thanks for PR which demonstrates an interesting new angle to the search tool. And yes, visual tests failure is indeed just a build configuration issue.

Currently the cross icon at the end of search is just a text editing tool which initiates input editing by first clearing the content. For this purpose, it works quite conveniently.

Your implementation where the cross immediately changes the URL seems to change UI behavior pretty dramatically. For example, in route summary page, attempt to change origin or destination with a click on cross jumps back to front page. Possible changes  in current routing settings (like departure time) would be lost.  Of course, users will learn the new meaning of the cross by time, but I still doubt this is  desirable.

Also, I see some regression in ease of use:
- Clearing the input would require more actions:  a double click and backspace/del key
- After clearing the search field, focus is lost. Another click/tap is needed for setting new content
- Also the magnifying glass icon got broken:  it no longer opens the default empty search list, with favorites and recent searches.

Setting up a homescreen app in Android probably happens seldom, so maybe actually editing the url text to get back to the initial state is a good enough solution. 

Anyways, these are my first impressions after quick testing. Further ideas and arguments are of course welcome.

eemeli:
The primary pain point I'm seeking to address here is that currently there's no way within the app of clearing a selection that's once been made. Perhaps a smaller change would be to address what happens when you first click on the X to clear the input field, and then tap elsewhere or otherwise blur the input. Currently, the input returns to the value it had before it was cleared, which in this case is unlikely to be what the user intended.

This would also allow for the button to keep its behaviour when showing the magnifying-glass icon.

vesameskanen:
I have passed your feedback to the designers. I will close this PR as it cannot be merged in its current form. Thanks for further information and your contribution!

PR #2561, opened 2018-11-07
DT-2817: update zone icons and area contours to the map layer

jolkkola:
The purpose of this pull request is to change the styling of the zone icons to match those used elsewhere in the software. In addition, the styles for the zone area contours were changed as requested.

Screenshot:
image


vesameskanen:
Please document multi style feature into docs/GeoJson.md .

PR #2568, opened 2018-11-11
Update react-leaflet

hannesj:
Remove usage of legacy context within map components


vesameskanen:
UI seems to throw errors in this branch: 

digitransit/ui/node_modules/react-leaflet/es/context.js:3
import _extends from "@babel/runtime/helpers/esm/extends";
^^^^^^

SyntaxError: Unexpected token import
    at createScript (vm.js:80:10)
    at Object.runInThisContext (vm.js:139:10)
    at Module._compile (module.js:616:28)

Any idea what breaks the UI?

PR #2576, opened 2018-11-15
DT-2822

vesameskanen:
Refactor realtime client code to support both mqtt and gtfs-rt. Hide mqtt implementation details from components.


app/configurations/config.tampere.js
```diff
@@ -38,12 +38,10 @@ export default configMerger(walttiConfig, {
   feedIds: ['tampere'],
 
   realTime: {
-    /*
     tampere: {
       gtfsRt: 'http://data.itsfactory.fi/journeys/api/1/gtfs-rt/vehicle-positions',
-      routeSelector: 'shortName',
+      routeSelector: (routePageProps) => routePageProps.route['shortName'],
```
jolkkola: This will not work in IE11 - at least previously no transpiling was done for the configuration files. I would suggest using the function notation.

app/configurations/config.default.js
```diff
@@ -60,6 +60,10 @@ export default {
     /* sources per feed Id */
     HSL: {
       mqtt: 'wss://mqtt.hsl.fi',
+      routeSelector: routePageProps => {
```
jolkkola: Same here, the arrow notation will not work in IE11.

PR #2582, opened 2018-11-20
Visual test tweak

HeikkiHei:
- Update gemini to support anti-aliasing tolerance.
- Add configs for tolerance, less retries.
- Change Chrome to run on Windows instead of OSX.
- Update pictures for comparison.


HeikkiHei:
Does not work yet, uses Travis resources in vain. Close.

PR #2631, opened 2019-01-10
Dt 2713

optionsome:
* Weekly route timetables
* New fetch util function that allows retrying fetch calls
* Upgrade to node 10


optionsome:
This PR is onhold until timetables are updated

optionsome:
Decided to disable hsl route timetable configuration from config files with comments so that they can be enabled in the future but this pr can be merged before there are too many conflicts

PR #2650, opened 2019-01-28
Update german translations

robbi5:
We've updated the digitransit-ui translations for german again :)


optionsome:
There are some lint errors. You can check those with `yarn run es-lint -- --quiet`, for example

robbi5:
moved variables around, linter is now happy. sorry :)

PR #2685, opened 2019-02-18
DT-2878: improved geojson support

jolkkola:
The purpose of this pull request is to enable native-like popups for POIs retrieved as geojson data. In addition, support was added for using an external geojson layer configuration.

JIRA link: https://digitransit.atlassian.net/browse/DT-2878

In addition, mocha's watch mode now works.

Screenshot:
image


vesameskanen:
Reviewed with respect!

PR #2716, opened 2019-03-19
Fix FilteredFavouritesPanel

### LINKED ISSUE #2714: shouldUpdate in FilteredFavouritesPanel doesn't work
Every single GPS movement triggers a new request and the favorite panels are almost useless because the route is shown only a very short time if at all.

Not sure why this occurs but I suspect this comparison:
```
nextProps.favouriteLocations !== props.favouriteLocations ||
nextProps.favouriteStops !== props.favouriteStops ||
```

I can't figure out any other reason why the component rerenders itself every time when the `origin` changes. The `FilteredNearbyRoutesPanel` component works fine because it compares only simple values in the same part.

ttsirkia:
## Proposed Changes

  - Compare contents instead of object references to avoid continuos updates which will make the favorite panel unusable as the information is not visible
  - Fixes #2714 

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/FavouritesPanel.js
+++ b/app/component/FavouritesPanel.js
@@ -3,6 +3,7 @@ import React from 'react';
 import Relay from 'react-relay/classic';
 import connectToStores from 'fluxible-addons-react/connectToStores';
 import shouldUpdate from 'recompose/shouldUpdate';
+import isEqual from 'lodash/isEqual';
 
 import FavouriteRouteListContainer from './FavouriteRouteListContainer';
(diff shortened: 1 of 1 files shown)


ttsirkia:
The judgement doesn't seem to be very fair. I don't confess breaking those...

pailakka:
Created new  PR (https://github.com/HSLdevcom/digitransit-ui/pull/2717) under HSLdevcom in order to enable environment variables in TravisCI

PR #2781, opened 2019-05-08
## Proposed Changes

obb12:
-Share button for android
## Tests
 wip

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,19 +1,3 @@
 ## Proposed Changes
 
-  -
-
-## Pull Request Check List
-
-  - A reasonable set of unit tests is included
-  - Console does not show new warnings/errors
-  - Changes are documented or they are self explanatory
-  - This pull request does not have any merge conflicts
-  - All existing tests pass in CI build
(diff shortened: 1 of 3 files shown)


obb12:
Do you need test to see if navigator.share is available?

obb12:
Pull-request updated, HEAD is now e0d6e14c174c70e670a71b73c2a9ffacfca073da

obb12:
BS_ACCESS_KEY not set, failing build on purpose

vesameskanen:
Thanks for the contribution! We will consider in the near future if the pull request can be merged as it is now, or do we perhaps need to make this a configurable option.

vesameskanen:
Hi! The share button idea was discussed thoroughly in a design session. Unfortunately we cannot use the PR as it is now. Namely, placing the share button into the context of a particular itinerary suggests that use of the shared url would somehow produce a similar route search result. Unfortunately this is not the case, because of constantly changing realtime predictions and user's own preferences. We do have plans to implement a feature for saving the actual static itinerary description, but that is not implemented yet.

Anyway, thank you for bringing up the interesting topic for discussion! Future will show how it evolves.

PR #2835, opened 2019-06-18
Add Virtual monitor link to stop page in Tampere 

Antiik91:
## Proposed Changes
An external link in stop page to virtual monitor. Configurable, and currently works in Tampere.


app/component/StopCardHeader.js
```diff
@@ -28,8 +29,34 @@ class StopCardHeader extends React.Component {
     return description;
   }
 
+  getExternalLink(code, isPopUp) {
+    // Check for popup from stopMarkerPopup, should the external link be visible
+    if (!code || isPopUp) {
```
optionsome: this.headerConfig.virtualMonitorBaseUrl check should probably be moved here as well and the later one can be removed-

app/component/CardHeader.js
```diff
@@ -84,6 +86,7 @@ CardHeader.propTypes = {
   name: PropTypes.string.isRequired,
   description: PropTypes.string.isRequired,
   code: PropTypes.string,
+  externalLink: PropTypes.object,
```
optionsome: I think this should maybe have object type node.

PR #2894, opened 2019-08-06
fix translation check

robbi5:
`node build/check_translations.js sv de` didn't work anymore, throws `SyntaxError: Unexpected token a in JSON at position ...`.

This happens, because translations.js misses quotes around object literal property names (translation keys without dash are not quoted anymore) and therefore cannot be read as json.

Simple fix for that is now using eval for parsing the array.


vesameskanen:
Merged in https://github.com/HSLdevcom/digitransit-ui/commit/d9e388ca949e644c6ba03ee7e5c4ee84848759bd

PR #3148, opened 2020-03-17
DT-3417

AleksiSiitari:
## Proposed Changes

  - add U prefix to U bus lines in the 

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen:
- Route page header layout breaks on 5 digit  route names (try searching 955A)
- Stop / station page layout breaks. 'Linja' column clips long route ids like U955A. Try ' hki linja-autoasema 1215'.  Same problem in Linja/Laituri view, too.
- Change mapRouting func name,  it is too generic. 'getRoutePrefix' might be better.
- Do we need to introduce config as ContextType in NextDeparturesList.js and some other components whic do not refer to it at all?

PR #3178, opened 2020-03-27
Dynamic searchinput ariacontent

MikkoPuustinen:
## Proposed Changes

  - First suggestion in the Autosuggest dropdown is now announced on NVDA every time it changes
  - NVDA now only announces the amount of found suggestions when it is relevant

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen:
- Change 'tämänhetkinen sijainti' to another expression, to avoid confusion with geolocation.
- The new dynamic aria message does not know how to read some suggestion types. That should be fixed because reading 'current selection'  without the related value is confusing. Note that when you browse suggestions with arrow keys, every suggestion will be read correctly. Do it the same way.

PR #3198, opened 2020-04-09
DT-3473: Show vehicles of the selected itinerary on the map

Kilbert76:
## Proposed Changes

  - Show vehicles of the selected itinerary on the map  

## Pull Request Check List

## Review

  - Merge the pull request


app/component/SummaryPage.js
```diff
@@ -93,6 +99,26 @@ export function reportError(error) {
   });
 }
 
+const startClient = context => {
+  const { realTime } = context.config;
+  let agency;
```
optionsome, line 104: Preferably call it feedId rather than agency because within a feed there can be multiple agencies and these realtime feeds cover data for multiple agencies within a feed.

app/component/SummaryPage.js
```diff
@@ -93,6 +99,26 @@ export function reportError(error) {
...
+
+  /* handle multiple feedid case */
+  context.config.feedIds.forEach(ag => {
```
optionsome, line 107: same as above

PR #3201, opened 2020-04-12
Added route pattern option translation for all core languages

BodoMinea:
Added route pattern option translation for Danish, Spanish, French, Norwegian and Romanian (which I added the core translation for) and also the corresponding initialization of resource bundles.

Prior to this, the aforementioned languages had a problem in the route view tab, pattern selection element where they were showing the translation tag instead of the string.


BodoMinea:
_This is failing on the BrowserStack tests because it lacks BS_ACCESS_KEY. I do not know if I can do anything about it._

PR #3226, opened 2020-04-30
Split translations

### LINKED ISSUE #3222: Split translations
Hi there!

I would like to ask your opinion about the `translations.js` file. Since more and more languages are there, it's getting huge and it is quite hard to work with. We'd like to update its contents with a script automatically and think that it would be easier to do so if the contents were sorted into different json files, one for each language.

If you think this is a working idea, we'll do the separation and raise a PR with it.

Beck-berry:
## Proposed Changes
  - split translations.js into JSON files

## Pull Request Check List

## Review
  - Merge the pull request

Closes  #3222


leonardehrenfried:
I think this could be combined with https://github.com/godaddy/eslint-plugin-i18n-json

optionsome:
We are currently using https://github.com/HSLdevcom/digitransit-ui/blob/master/test/unit/translations.test.js for checking that "Swedish" and "English" translations exist for each key defined for Finnish "translations", but perhaps that eslint plugin could do the same thing. Problem is that we cannot maintain all the translations while we are developing new features. Therefore, the check is now limited to just those two languages.

leonardehrenfried:
I think the `sort-keys` and `duplicate-keys` option might be enough for the beginning, because that is what I personally care about.

The syntax check and the others could be left for later.

leonardehrenfried:
But more importantly, are you generally in favour of this change?

optionsome:
Yes I think using that eslint plugin for those purposes could make sense. Also, splitting the translations file is a good change.

meromisi:
We switched from _eslint-plugin-i18n-json_ to _jsonlint_. We couldn't make the plugin work with json because they were parsed as Javascript files. I tested jsonlint and it seems to be working fine well.

Issue #3059, opened 2020-01-22
Departure view: grouping rows by route/direction instead of trips?


hbruch:
The departure view seems to group departures of identical trips in a row, indicating their next departures:
<img width="333" alt="image" src="https://user-images.githubusercontent.com/2187389/72893663-e1344f80-3d19-11ea-8e1e-16b0647a92f7.png">

However, some of our routes have trips which have identical headsigns, nearly identical trip patterns, only their quay differs for a few stops (the parent stop is currently not given in the GTFS but we could deduce it from the stop_id):

Extract from stop_times.txt:
```
"29.T0.10-2-j20-5.2.H","14:26:00","14:26:00","de:08116:1905:1:1","27","","0","0"
"30.T0.10-2-j20-5.7.H","14:56:00","14:56:00","de:08116:1905:1:2","27","","0","0"
```
Is there an option to group by route/direction (or headsign) instead of trips? And would the route list be able to handle this? 
Or, to just show next departures without grouping (and hence omitting over next departure)?

optionsome:
I think the current logic was built using data which did not have these types of trips and I don't think there is any configuration available for what is shown there as of now. I think there is room for improvement but I think we will get rid of this display at least in this current form at some point in the future when we are doing larger redesign of the UI. Therefore, I don't think we are doing any changes to it but if you happen to do some fix that can be enabled through a configuration parameter, you can create a pull request.

PR #3230, opened 2020-05-05
Add opportunity to disable "next" column in Route panel

Beck-berry:
## Proposed Changes
Based on MFDZ issue [#167](https://github.com/mfdz/digitransit-ui/issues/167): in our region there are small differences between routes which are handled as totally different ones, and so next departure is set to the next day (which is not true).
- added new config variable `displayNextDeparture` to config.default.js with the value `true`
- modified corresponding components of the Route panel in order to make show / display the "Next" column
- If true:
Screenshot_2020-05-05_11-13-54
- If false:
Screenshot_2020-05-05_11-13-09

We were a bit unsure about the config key - what do you think, should it be a subkey of something or would you prefer another name?

## Pull Request Check List

## Review
  - Merge the pull request

### DIFF
--- a/app/component/PatternStopsContainer.js
+++ b/app/component/PatternStopsContainer.js
@@ -25,6 +25,7 @@ class PatternStopsContainer extends React.PureComponent {
 
   static contextTypes = {
     router: routerShape.isRequired,
+    config: PropTypes.object.isRequired,
   };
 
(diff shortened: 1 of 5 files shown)


leonardehrenfried:
Note that this is meant to address the problem described in #3059 .

PR #3235, opened 2020-05-06
fix: handle uncaught relay parsing error globally

leonardehrenfried:
## Proposed Changes

When relay 1.6.0 encounters a GraphQL parsing error it doesn't seem to return a failed promise or at least a synchronous exception but it schedules an exception to be thrown later: https://github.com/facebook/relay/blob/v1.6.0/packages/react-relay/classic/util/throwFailedPromise.js

For example, you can provoke a parsing error by setting the triangle factors of the bike routing in the URL to add up to more than 1.0.

When the page is rendered with `react-isomorphic-relay-router` then this leads to a server crash as the exception is thrown completely outside the ordinary request/response cycle and there is no handler for it.

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/server.js
+++ b/app/server.js
@@ -371,7 +371,7 @@ export default function(req, res, next) {
       .catch(err => {
         // eslint-disable-next-line no-console
         console.log(err);
-        return ['', undefined];
+        return ['There was an error fetching your data.', undefined];
       });
 
     const [[content, relayData]] = await Promise.all([
--- a/server/server.js
+++ b/server/server.js
@@ -33,6 +33,16 @@ if (process.env.NODE_ENV === 'production' && process.env.SENTRY_SECRET_DSN) {
   });
 }
 
+process.on('uncaughtException', error => {
+  // perhaps this is a little too careful but i want to make sure that things like out of memory errors
+  // are still crashing the process as there is no point in trying to recover from these.
+  if (error.name === 'RRNLRequestError') {
+    console.log('Unhandled Error', error);
+  } else {
+    throw error;
+  }
+});
+
 /* ********* Server ********* */
 const express = require('express');
 const expressStaticGzip = require('express-static-gzip');


optionsome:
Perhaps in addition to this fix we could add missing validation for the parameter(s). There is already some relevant code in https://github.com/HSLdevcom/digitransit-ui/blob/master/server/reittiopasParameterMiddleware.js#L27

leonardehrenfried:
Is the validation a requirement to get this merged?

optionsome:
No. I think these situation are kinda rare and probably caused by the users themselves, so it's not necessarily worth the effort to add validation for all parameters. It just would prevent the mostly white page with untranslated error message from being shown but we change the parameters a bit in the new design anyway so I think we can add the missing validation there.

optionsome:
Also, we have upgraded relay for the new version of digitransit-ui and the validation is probably not even needed as these types of errors are hopefully handled by the libraries.

leonardehrenfried:
Yes, I find strange to throw these kind of delayed exceptions but I'm sure there was a reason for doing it that way. :man_shrugging:

leonardehrenfried:
On a related note, in several PRs i noticed that during the flow tests Chrome could not be installed with the following error message:

```
-1 [67.0 MB]

Fetched 67.0 MB in 1s (44.0 MB/s)

dpkg-deb: error: archive '/var/cache/apt/archives/google-chrome-stable_81.0.4044.138-1_amd64.deb' has premature member 'control.tar.xz' before 'control.tar.gz', giving up

dpkg: error processing archive /var/cache/apt/archives/google-chrome-stable_81.0.4044.138-1_amd64.deb (--unpack):

 subprocess dpkg-deb --control returned error exit status 2
```
This appears to be a bug in `apt`: https://askubuntu.com/questions/1065231/dpkg-deb-error-archive-has-premature-member-control-tar-xz-before-contr

Is there a reason you use Ubuntu 14.04 as the build machine?

optionsome:
I think something (maybe test unit coverage thingy) broke when we tried upgrading it once. Should probably look into it again at some point.

leonardehrenfried:
Seems that upgrading to Ubuntu 18.04 fixed it: https://github.com/HSLdevcom/digitransit-ui/commit/1037dd2ed58d31d7b177506713b53b3d1f27f81e

Thanks for merging this!

PR #3256, opened 2020-05-18
French translation update.

kalon33:
## Proposed Changes

  - Update French translation to be on par with latest en.json strings changes.


leonardehrenfried:
Hi @kalon33,

our team has contributed the separation of the translations into separate files. We've also added a linter that automatically formats these files but we have not added them to the pre-commit hooks yet.

If you want to reformat the French translation file you can run `yarn lint-translations` and it will order the keys alphabetically.

@vesameskanen Sadly, jsonlint doesn't fail if the keys are not sorted alphabetically, it just sorts them. Would you be ok with us adding this to the `static` npm step, like we did in our fork here: https://github.com/mfdz/digitransit-ui/pull/198/commits/fd0138c56fc99ba1d2beaee4281c88c98c49c1f9

PR #3281, opened 2020-06-01
DT-3533: New stop page styles

AleksiSiitari:
## Proposed Changes

  - New stop page styles according to design layout

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen:
A further note: stop page should have the same margins as the front page. Back arrow is located to the left from the content column, and should align with the header text. For some reason, the arrow is above the header text for stations and below it for normal stops.

PR #3282, opened 2020-06-01
Dt 3532: Route page header styling

MikkoPuustinen:
## Proposed Changes

  - Phase 1 layout changes to Route page

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen:
Looks good, but a couple of issues needs fixing:
- Narrower stop list sometimes break the layout. Try 8H and find the stop 'Kumpulan kampus' which also has 'Vain poistuminen' info text attached. Text wont fit.
- The favorite start at the top appears broken. The style won't change when the start button is clicked. It is supposed to make the route as a favorite one. Check from the design how the style is supposed to change.

vesameskanen:
On note more: vehicle icons on the map are now clearly larger than in zeplin layouts.  Shrink them to the defined size.

PR #3287, opened 2020-06-02
DT-3544: Search loses arriveBy parameter

Kilbert76:
## Proposed Changes

  - Fixing problem that using datetimepicker before setting origin/destination

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen:
Some issues need fixing:
- Change datetimePickerQuery to a better name. For example, itineraryParams. 'datetimePicker' is an UI element and irrelevant in describing the application state.
- Search components have nothing to do with itinerary parametrization. So, do not pass datetimePickerQuery to the shared component, that does not make any sense. Simply use this.props.itineraryParams in the 'onSelect' callbacks instead of bouncing the data back and forth.

PR #3311, opened 2020-06-09
DT-3398: Stop icons links to stop page

TarmoA:
## Proposed Changes

  - Clicking stop icons on map now links to stop page
  - Improve styling of popup when selecting from multiple stops

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen:
Some issues need fixing:
- Popup is too small. In itinerary page, where 'add middle point'  button is available, bottom row buttons get badly clipped.
- Stop link messes up page history. Getting back from stop page loses already selected itinerary search options and origin. After two stop page visits, back button no longer leads back to front page. Use router.push, not replace, to make page transformations.

PR #3340, opened 2020-06-15
feat: use different mqtt broker URL for testing new features

optionsome:
## Proposed Changes

  - Changed HSL mqtt URL so we can safely test new use cases of mqtt without risk of breaking production broker

## Pull Request Check List

## Review

  - Merge the pull request


optionsome:
Decided to do this with REALTIME_PATCH env instead as this change would have been reverted at some point

Issue #3670, opened 2020-11-04
[next] No way to use a custom map attribution


leonardehrenfried:
We are following the `next` branch to build up a digitransit instance for our client. We noticed that there is a config option `showOSMCopyright` but unfortunately we need a more complex attribution where we can define the text being shown.

The background for this change is that we use a few data providers and each of these must be credited.

Would you be interested in a PR that allows you to set the attribution text?

I can image the following way of specifying it in the config file:

```
map {
  attribution: false // no attribution
  attribution: "&copy; <a tabindex=&quot;-1&quot; href=&quot;http://osm.org/copyright&quot;>OpenStreetMap</a>" // default attribution
  attribution: "© <a tabindex=-1 href=http://osm.org/copyright>OpenStreetMap Mitwirkende</a>, <a tabindex=-1 href=https://www.nvbw.de/aufgaben/digitale-mobilitaet/open-data/>Datensätze der NVBW GmbH</a> und <a tabindex=-1 href=https://www.openvvs.de/dataset/gtfs-daten>VVS GmbH</a>" // what we need
}
```

What do you think?

vesameskanen:
Hi, Leonard

Sounds good to me, go ahead!

For your information, the focus in the next development has been strongly in one single 'HSL' instance, which will be releases in very near future.  I expect that after the release, the new code base will be developed towards more flexible and generic direction by  the digitransit team. Your contribution fits that well.

leonardehrenfried:
Closed by #3687

PR #3671, opened 2020-11-04
[next] Fix display of layer switcher button

leonardehrenfried:
## Proposed Changes

This fixes the display of the layer switcher so that it can be activated without the bubble moving beyond the edge of the visible area. It also moves the bubbles underneath each other.

## Before
Screenshot from 2020-11-04 15-57-27

## After
Screenshot from 2020-11-04 15-27-36

### Without layer switcher
Screenshot from 2020-11-04 15-52-33

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/bubble-dialog.scss
+++ b/app/component/bubble-dialog.scss
@@ -4,6 +4,8 @@ $bubble-dialog-font-size--fullscreen: 1.5rem;
 .bubble-dialog-component-container {
   position: relative;
   z-index: auto;
+  box-shadow: 0 2px 10px rgba(0, 0, 0, 0.2);
+  margin: 0 0 8px 8px;
 
   .bubble-dialog-container {
     bottom: 2.5em;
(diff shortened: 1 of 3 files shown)


vesameskanen:
Unfortunately this breaks our current next branch styles, whrere the layer switcher is not used:
image

leonardehrenfried:
Ok, I will look into it! Should not be difficult to fix.

leonardehrenfried:
I figured out what the problem was. As already discussed in #3670 we changed the map attribution in our fork but since our long text didn't look so nice while floating underneath the zoom buttons we used a more space-saving design reminiscent of DT 1.

Screenshot from 2020-11-05 09-46-09

(Also available here: https://next.stadtnavi.eu/)

This moved our zoom buttons a little closer to the bottom and therefore the absolute values for the Locate Me button and the layer switcher are slightly off.

Question: Would you be interested in also using the space saving design that we used? If yes, I can also include it in the PR.

vesameskanen:
Hi,

Sorry for delay in replying. Our designer wish to keep the current style, so no thanks.

PR #3747, opened 2020-11-23
Remove hardcoded paths

leonardehrenfried:
## Proposed Changes

The application contained a few places where there were hardcoded Finnish URL paths. This PR removes them and replaces them the constants or with function parameters.

Tests were also adjusted.

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/WithSearchContext.js
+++ b/app/component/WithSearchContext.js
@@ -12,6 +12,10 @@ import {
   navigateTo,
   PREFIX_NEARYOU,
   PREFIX_ITINERARY_SUMMARY,
+  PREFIX_BIKESTATIONS,
+  PREFIX_STOPS,
+  PREFIX_TERMINALS,
+  PREFIX_ROUTES,
 } from '../util/path';
 import searchContext from '../util/searchContext';
 import intializeSearchContext from '../util/DTSearchContextInitializer';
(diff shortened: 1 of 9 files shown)


vesameskanen:
Hi Leonard,

Thanks for the new contribution. I think there is  a slight problem: npm packaged components (digitransit-component, digitransit-search-util) must not depend on digitransit application code (more precisely app/util/path.js).

I do not see any other solution as declaring the required paths as optional new props, which default to old hard-coded values, so that current API will not break. 

Some of the  new path props can perhaps be hash maps where multiple paths are defined. For example:

static defaultProps = {
    prefixMap:  { prefixRoutes: 'linjat', prefixStops: 'pysakit' },
   ...

I hope you get the idea.

leonardehrenfried:
Yes, that sounds reasonable. However, my patch doesn't add a dependency on application code. Am I right in thinking that you would like the method signature to remain the same?

I have question though: the packages I touched aren't react components but plain functions. In order to make them backwards-compatible, I would add an optional `opts` parameter and set default values in the function itself. 

Is this what you imagined?

vesameskanen:
To clarify  further, I'd like to pinpoint one specially problematic line:

https://github.com/HSLdevcom/digitransit-ui/blob/868580e05699a1c2ab7a01630272bf7da792f469/digitransit-component/packages/digitransit-component-autosuggest/src/index.js#L22

The components are already being used in other web applications, and import paths referring outside the code of the npm module itself cannot be used.

Regarding method signature: I think you got my idea correctly. We do not wish to introduce breaking changes without a good reason, so backward compatibility should be maintained whenever possible.

Anyway, if this requirement seems to lead to complicated code and overly deep prop drilling, we should use common sense and change the interface. Please use your own consideration while implementing the changes.

leonardehrenfried:
> https://github.com/HSLdevcom/digitransit-ui/blob/868580e05699a1c2ab7a01630272bf7da792f469/digitransit-component/packages/digitransit-component-autosuggest/src/index.js#L22

Oh yes, you're totally right. I totally overlooked this!

I will fix it and implement your requests.

leonardehrenfried:
I have reworked this PR and it's ready for another review.

PR #3818, opened 2020-12-16
DT-4147: Indicator when walking/biking results are loaded

ahjyrkia:
## Proposed Changes

  - Add accessibility alert when biking and walking results are loaded

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen:
Works otherwise but if user clicks earlier departures, screen reader tells 'walk & bike updated' although actually public transit results were updated.

PR #3878, opened 2021-01-21
DT-4233: hsl.fi and reittiopas front page look different

MikkoPuustinen:
## Proposed Changes

  - Front page now matches design

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen:
Better now. Would it be easy to fix teh remaining differences:
- In desktop, reittiopas controls are some pixels wider
- In mobile, stops near you tile text is wider in reittiopas. Also the icons stretch to fill the row. It would be better to use fixed size and padding like in hsl.fi.

PR #3907, opened 2021-01-29
routepage link which works the same way in server and client routing

vesameskanen:


vesameskanen:
In the example picture, platform number font is some pixels smaller than departure time. Also the surrounding box is smaller.

PR #3913, opened 2021-02-02
DT-4212: User location reference is never passed to search

Antiik91:
## Proposed Changes

  - Bring focuspoint back to search

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/IndexPage.js
+++ b/app/component/IndexPage.js
@@ -430,9 +430,8 @@ const IndexPageWithStores = connectToStores(
     const newProps = {};
     const { query } = location;
     const { favouriteModalAction, fromMap } = query;
-    if (locationState) {
-      newProps.locationState = locationState;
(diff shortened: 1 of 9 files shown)


vesameskanen:
Location is now set correctly, but geocoding ingores it for some reason. That will be fixed in another ticket.

Adding refpoint in the middle of public executeSearch func breaks the api. I suggest moving the new parameter to the end of parameter list. Otherwise the major package version must be increased.

Antiik91:
> Location is now set correctly, but geocoding ingores it for some reason. That will be fixed in another ticket.
> 
> Adding refpoint in the middle of public executeSearch func breaks the api. I suggest moving the new parameter to the end of parameter list. Otherwise the major package version must be increased.

Good point. This is now refactored.

PR #3917, opened 2021-02-02
DT-4127: The Show more button works strangely on The closest ferry piers near you view

TarmoA:
## Proposed Changes

  - Hide stops with no upcoming traffic from stops near you map
  - This causes map to zoom farther away to see nearest stop, especially on ferry page
  - Also draw the icon for highlighted nearest stops on smaller zoom levels than before
  - At or below zoom level 12 icon is still hidden because map service does not return stops data at that level

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen:
For some reason, maps spins now very long time. An example: set location to kuninkaantammi, go to subway view. Next dev loads map in a couple of seconds, but this branch spins over 20 s. The view appears broken.

PR #3925, opened 2021-02-04
DT-4299: stop page new date and route picker

AleksiSiitari:
## Proposed Changes

  - New stop page date and route picker
  - desktop print buttons new location
  - remove now redundant component
  - improve timetable print styles

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen:
Clicking the calendar should open the date dropdown. Optimally, we should use here the same datepicker as in itinerary search settings.  However it is changing so lets wait until the final version.

PR #3928, opened 2021-02-08
DT-4242

ahjyrkia:
## Proposed Changes

  - Custom geojson popup contents and position fixed

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen:
Works with geoJson now, but there is a strange regression: bust stops and the terminal in Elielinaukio have stop/terminal page header empty. Most stops and terminals work fine, so there must be some special data associated with these stops. In next-dev, headers is fine.

Most likely the stop page uses the same component as the popup menu.

PR #3938, opened 2021-02-11
DT-4294: Nearby stops: Search location

MikkoPuustinen:
## Proposed Changes

  - Search field for updating the view with a new position

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen:
Works fine, but implementation need some refactoring:

Search is a general purpose component, which has some logic and renders  something. It is not a concern of it to know what kind of pages use it.  So, adding search to stops near you page should not change it in any way. It that happens, the search or the page which uses it is coded wrong.

- Add a prop to pass translated custom placeholder text
- Move all stops near you specific styles away from the search component. Add a wrapper if necessary.

PR #3960, opened 2021-03-01
DT-4142: Non-hsl links for stops near you

Kilbert76:
## Proposed Changes

  - Fixes as designed

## Pull Request Check List

## Review

  - Merge the pull request


app/configurations/config.waltti.js
```diff
@@ -41,38 +41,92 @@ export default {
   defaultLanguage: 'fi',
 
   transportModes: {
+    nearYouButton: {
```
derhuerst, line 44: This is *really confusing* while debugging in `modeUtils.js` (on `v3`), because it ends up being treated as a transport mode, and just *coincidentally* doesn't break things.
cc @vesameskanen @hbruch

vesameskanen: indeed, not only confusing but somewhat hilarious. Will be fixed.

vesameskanen: https://github.com/HSLdevcom/digitransit-ui/commit/896064bdbf3b840a0f784f57f47ffe53cb90214c

PR #3984, opened 2021-03-11
DT-4335: Redesign map popup

MikkoPuustinen:
## Proposed Changes

  - New design for location popups

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen:
Some views do not show origin/destination buttons.  In such cases, the popup should have more margin at the bottom.

Click the map e.g. in Stop view to see what I mean.

Another problem: geoJson objects seem to use the same location popup. There are some new errors, as the attached image from turku theme indicates: text overlaps close cross, buttons do not fit and the bottom tip is dislocated from the popup container.

image

PR #4006, opened 2021-03-18
DT-4179: Message bar for waltti instances

AleksiSiitari:
## Proposed Changes

  -

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen:
There are some issues/errors left:
- If message bar includes several messages and user first swipes between them, closing the message bar is no longer possible. Last message refuses to disappear.
- Geolocation error message which shows up while another message is already shown flashes text 'siirry sisältöön'  for a while. Maybe this is a dev server problem only, but check the  issue when the feature is in next-dev.
- Static messages from configuration (and obviously from yleisviesti kind of api which uses same format) are displayed wrong. 'Lue lisää' is automatically injected into message content, although the whole content is already visible. Link section no longer shows the actual link name but the url, which is supposed to be hidden.

vesameskanen:
Still one glitch left: if message is short (visible without 'näytä lisää' selection), the link section does not act as link.  It looks like text and clicking it does nothing.

Happens especially easily in desktop as messages usually fit.

PR #4153, opened 2021-05-17
Dynamic zooming of map items

vesameskanen:
- Move vehicle drawing to Map.js. Now one component does zoom monitoring.
- Remove strange resize detector as it seems to do nothing
- Clean up code more


vesameskanen:
Use CONFIG=jyvaskyla yarn dev to test vehicles on map. Tampere and Turku seem to have data problems.

PR #4286, opened 2021-06-30
Remove all vehicles that haven't been updated for 3 minutes

leonardehrenfried:
## Proposed Changes

We noticed that when you keep the screen with bus positions open for a long time, then finished busses accumulate on the screen.

image

For this reason we remove any vehicle that hasn't had an update in more than 3 minutes.

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/map/VehicleMarkerContainer.js
+++ b/app/component/map/VehicleMarkerContainer.js
@@ -2,6 +2,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import connectToStores from 'fluxible-addons-react/connectToStores';
 
+import moment from 'moment';
 import VehicleIcon from '../VehicleIcon';
 import IconMarker from './IconMarker';
 import { isBrowser } from '../../util/browser';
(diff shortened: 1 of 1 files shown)


app/component/map/VehicleMarkerContainer.js
```diff
@@ -127,9 +128,17 @@ const connectedComponent = connectToStores(
       );
       vehiclesFiltered = Object.fromEntries(filtered);
     }
+    // if you keep the UI open for a long time then trips that have finsished accumulated on the screen
+    // this removes anything that hasn't had an update in 3 minutes
+    const vehiclesWithRecentUpdates = Object.entries(vehiclesFiltered).filter(
+      ([, message]) => {
+        const threeMinutesAgo = moment().subtract(180, 'seconds');
+        return moment.unix(message.timestamp).isAfter(threeMinutesAgo);
```
optionsome, line 136: At least with the vehicle position feeds we use, the message.timestamp value comes from the feed instead of being determined in client. If either client or the feed clock is not in sync, vehicles might be filtered even though they've actually received recent updates.

leonardehrenfried: I'm not sure I know what the feed clock is. Are you saying that the timestamp doesn't refer to the time when the message was sent but some other one?

If the client clock is off, wouldn't that also cause a lot of other issues?

leonardehrenfried: One solution to this: adding a `receivedAt` timestamp for each message and then filter by that.

optionsome: Currently, the timestamp refers to the time the message was sent but I think receivedAt timestamp would fix this potential issue.

leonardehrenfried: In case I wasn't totally clear: I meant adding this in the client.

optionsome: The receivedAt timestamp should be generated at the client, if you mean that. Timestamp is generated at the origin of the data.

optionsome:
We included these changes in https://github.com/HSLdevcom/digitransit-ui/pull/4543

PR #4301, opened 2021-07-18
Use correct style guide url

simeonkerkola:
## Proposed Changes

  - Changed style guide url in readme to a correct one

## Pull Request Check List

## Review

  - Merge the pull request


optionsome:
I think the styleguide that was in the earlier link is different than in the updated link. The link was to the digitransit styleguide, the new link u gave is to HSL style guide which is a bit different. We are currently in process of removing the digitransit styleguide as it was a bit hard to maintain so I think the better option is to remove it from the documentation.

vesameskanen:
style guide is now removed

PR #4370, opened 2021-09-09
DT-4916: define citybike seasons in configuration

AleksiSiitari:
## Proposed Changes

  - Citybikes are displayed according to their season if it is defined, currently defined for Hsl, Lappeenranta and Turku

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen:
Some notes:
- Capacity should now  be a network property. Merging has incorrectly duplicated it.  Only default conf should have global capacity value for setting a default value.
- matka.fi configuration also includes citybikes. Maybe it could import other configs  and check if bikes are enabled there, and compose the networks array + feature enabling accordingly. Also, map drawing should filter drawn stations by existing networks.

UI server caches evaluated configurations, so citybikes do not activate unless UI server is restarted and user reloads the website. Add some comments  to document this somewhat unexpected behavior.

PR #4382, opened 2021-09-16
DT-4961: fix citybike network selector in itinerary search settings

AleksiSiitari:
## Proposed Changes

  - Network selector now shows up and works correctly

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen:
I found one more issue, which could be fixed in the scope of this  PR. If user selects vantaa network and disables smoove, and then vantaa season ends, network selector disappears and user has no way to reactivate the only existing  network. Citybikes  won't be available although the mode is set. Apparently network selection should not count if there is only one network active.

PR #4390, opened 2021-09-23
DT-4975: fix car routing settings

AleksiSiitari:
## Proposed Changes

  - The settings toggles work now

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen:
I believe we should not remove possibility to control car and PR separately. configuration should have a feature switch and a default value for each. 

HSL instance does not wish to show plain car routes, so the car routing option is hidden (feature switch). Out German friends want to have both. 

Maybe a small waltti city wants to offer just car routing, as park&ride does not make sense there.

PR #4397, opened 2021-10-04
DT-4988: Stops near you page refetches data when switching tabs

MikkoPuustinen:
## Proposed Changes

  - Query variables are static between tabs resulting in less refetches
  - Map doesn't get completely rerendered when changing tabs

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen:
Updates work now much more smoothly, very good. There is one regression thoug:
- Enter Stopsnearyoupage with any location/mode
- Navigate on the map until 'map location has changed' note appears
- Press 'Update xxx stops' button
Stop list updates but 'map has changed' note and update button remain

PR #4453, opened 2021-11-30
DT-5149: First iteration of embedded search

AleksiSiitari:
## Proposed Changes

  - a very basic version of the embedded search, supports pre-filling the origin/destination from url

## Pull Request Check List

## Review

  - Merge the pull request


AleksiSiitari:
You can ask me for an iframe test html file

vesameskanen:
A good start! Some observations for further development:
- path should be in finnish '/haku' and defined the same way as other paths, not hard coded
- select final url params carefully as those will be used by the rest of the world. olon and dlon are hard to understand. Maybe lat1, address1, .., lat2, address2 etc.
- How about trying all search sources, including history and favorites? Maybe localstore can be accessed from iframe. Worth trying.
- Selecting Current position does not work. Most likely withSearchContext needs some special handling so that Current Position is treated as a symbolic string, just like in hsl.fi (passed to reittiopas as 'POS' string)
- Mobile search most likely needs some tuning. It is a bit annoying that enter key does not trigger the selection. Also, ability to scroll the suggestion list is a bit hidden feature. When there is no scroll bar, user has to use arrow keys or mouse wheel.
 
BTW, language selection seems to work nicely like this: /en/embedded-search/

PR #4471, opened 2022-01-05
DT-3937: map related bugs in stops near you page

MikkoPuustinen:
## Proposed Changes

  - Hilighted stop changes correctly on the map when selecting favorites in stops near you page and while visiting different stops in stop page
  - Changing tabs in stops near you page changes the mode of the visible stops correctly

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/map/MapWithTracking.js
+++ b/app/component/map/MapWithTracking.js
@@ -82,7 +82,6 @@ class MapWithTrackingStateHandler extends React.Component {
     this.state = {
       mapTracking: props.mapTracking,
       settingsOpen: false,
-      forcedLayers: null,
     };
     this.naviProps = {};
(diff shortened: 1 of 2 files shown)


vesameskanen:
All listed bugs are now fixed, good work! 

There is a new regression, probably related to the complicated layer visibility forcing in non-hsl instances. Visibility forcing is now partly broken.  Run UI with tampere config and try this: 

In front page, turn bus stops map layer off. Search stop 'keskustori c ' to enter the stop page. Note that no bus stops are drawn on the map. Wait approximately 20 s - stops finally show up. That is too late, user has probably left the view by then.

If bus stops layer is enabled, stops show up immediately.

PR #4485, opened 2022-01-18
DT-4690: Accessibility of trip view

MikkoPuustinen:
## Proposed Changes

  - Vehicle icons in trip page lead to trip page instead of route page
  - Currently selected vehicle is reported to screen reader
  - The stops the vehicles are currently on are reported correctly in trip and route pages

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen:
I noticed one glitch on mobile devices (tested iPhone + voiceOver): when swiping over vehicle icons on stop chain, screen reader usually reads some huge number instead of the vehicle symbol. Tested with metro M1, sometimes M1 is read but more often an astronomical figure (trip id?).

PR #4491, opened 2022-01-31
DT-5135: Form to create embedded search component

AleksiSiitari:
## Proposed Changes

  -

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen (changes requested):
Prefilling of origin and destination seem to behave a bit inconsistently.
-  If selection is 'vapaasti valittavat', the search form will still prefill origin=current location. Both should be empty, right?
- If language is sv  or en, selecting prefilled current location will display 'nykyinen sijaintisi' ie finnish label (change lang first, then prefill current location to duplicate)
- Changing prefilling options will not update the preview at all.

PR #4497, opened 2022-02-03
fix: nvda reading messagebar alerts multiple times on firefox

AleksiSiitari:
## Proposed Changes

  - The uuid keys changed on every render resulting in screen reader believing the content had changed

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen:
Much better now. However, there is one minor glitch:
- start NVDA and then firefox
- move focus to origin search using tab. Set origin.

NVDA reads the banner content. This seems to happen only once (changing origin will not read it again).  This is no longer annoying but a bit confusing. Is this fixable?

PR #4500, opened 2022-02-08
DT-5265: general accessibility fixes 

AleksiSiitari:
## Proposed Changes

  - Accesibility fixes for problems found by axe
  - Fixes most of the problems listed in the ticket,  autosuggest is left untouched
  - Fixes done to following components: itinerary settings menu labels, swipeable tabs, timetable, icon, favourite suggestion list

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen:
One change looks strange:  in swipeable tab header, moving focus with tab key makes an aria label visible when focus is moved through swiping controls.

AleksiSiitari:
I would argue this is an accessibility improvement for non-screen reader keyboard users (similarly to visible "skip to content" links) but we can also hide the swipeable tab instruction text if its too ugly

PR #4504, opened 2022-02-15
DT-5272: favourite bar dropdown accessibility

AleksiSiitari:
## Proposed Changes

  - Changed favourite bar dropdown from listbox to more semantically correct toolbar, the buttons in the dropdown should be read and be navigable by all screen readers
  - Changed select logic by removing highlightedIndex which was causing problems on safari+VoiceOver

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen:
Works otherwise OK, but selection of topmost dropdown item is problematic.  In firefox + NVDA,  first item can't be selected. In Chrome, closing the dropdown (enter or space) also does selection of the first item.

PR #4511, opened 2022-03-01
DT-5301: improvements for embedded search

AleksiSiitari:
## Proposed Changes

  - Add cookie information link to HSL instance embedded search creation form
  - Fix invisible tab focus on mobile autosuggest
  - Fix embedded search component shrinking in certain situations
  - Properly clear values from memory when pressing "X" in autosuggest fields

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/EmbeddedSearch.js
+++ b/app/component/EmbeddedSearch.js
@@ -169,14 +169,6 @@ const EmbeddedSearch = (props, context) => {
   const sources = ['Favourite', 'History', 'Datasource'];
   const refPoint = getRefPoint(origin, destination, {});
 
-  const onClear = id => {
-    if (id === 'origin') {
-      setOrigin({});
-    } else {
-      setDestination({});
-    }
-  };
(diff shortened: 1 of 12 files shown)


vesameskanen:
All listed bugs are fixed and new features work fine. However, one totally new (actually most likely a very old bug) detected (tested with Chrome):
- On desktop, shrink browser window to get mobile layout. Set also a low window height so that the whole search will not fit.
- Start typing a search term to a search field, say origin = alek
- On the result list of the mobile search view, step down the result list with arrow key
- When selection highlight moves out from visible area, some kind of refresh happens. Note that input field now correctly shows the new lowest item which became visible. However, selection highlight is now somewhere in the middel of the list and does not match input value. Hitting enter will select the highlighted item.

Note that search works fine on real mobile devices and in desktop layout. Also, when running mobile layout on chrome debugger, the problem disappears.

PR #4531, opened 2022-03-29
DT-5334 - tune city bike seasons and settings for 2022 

vesameskanen:


app/configurations/config.tampere.js
```diff
@@ -250,7 +250,7 @@ export default configMerger(walttiConfig, {
         enabled: true,
         season: {
           // 1.4. - 31.10.
```
optionsome:
```suggestion
          // 15.4. - 31.10.
```

PR #4545, opened 2022-04-21
Docker & CI build: improve speed by leveraging caching

derhuerst:
Still missing:

- [ ] ~~Mount `.yarn/cache` into `node-modules` the Docker build.~~
- [x] Try caching the `node-modules` stage, and check if it is actually faster than building from scratch. – much faster, I didn't expect that
- [x] Check if the CI runs as intended.
- [ ] Investigate why `yarn build` is a lot slower than before. 🤔 ([ ~5m before](https://github.com/HSLdevcom/digitransit-ui/runs/6112614393?check_suite_focus=true#step:8:1), 10-20m after) – due to the upgraded `favicons-webpack-plugin`
- [x] My addition of `--immutable` to `yarn install` causes the build to fail, as currently both on `v3` and `improve-ci-speed`, the checked-in `yarn.lock` does not reflect the workspace packages' versions. cc @vesameskanen

## Proposed Changes

This PR changes the `Dockerfile` & GitHub Actions dev pipeline to be faster.

While an un-cached (different `yarn.lock` and/or different `digitransit-*/**` and/or different commit hash) build is *slightly* slower, a cached one is is much faster.

## Review

For an unknown reason, `favicons-webpack-plugin@0.0.9` crashed during the Docker build, and `favicons-webpack-plugin@^4` froze, so I used `favicons-webpack-plugin@^3`.

When looking at the changes, I recommend to review each commit individually, as I tried to make them as atomic and sensible as possible. Some of them also contain explanations.

Unfortunately, because [the GitHub Actions system seems to lack support](https://github.com/orgs/community/discussions/25545) for [YAML anchors](https://yaml.org/spec/1.2.2/#example-node-for-sammy-sosa-appears-twice-in-this-document), I added a lot of duplicated code in `.github/workflows/dev-pipeline.yml`. The benefit of the repeated "use cache if exists, otherwise rebuild" patterns is that each step is effectively idempotent – it works regardless of wether the cache is available and if a previous step populated it –, with the downside being very repetitive code. Once YAML anchors are supported, we could DRY it a lot.


.github/workflows/dev-pipeline.yml
```diff
@@ -3,6 +3,8 @@ on:
   push:
     branches:
       - v2
+      # todo: remove
```
derhuerst, line 12: This needs to be removed before merging!

.github/workflows/scripts/build_and_push_dev.sh
```diff
@@ -15,10 +15,13 @@ echo "Building digitransit-ui"
 echo -e "export const COMMIT_ID = \"${GITHUB_SHA}\";\nexport const BUILD_TIME = \""`date -Iminutes -u`"\";" > app/buildInfo.js
 docker build --tag=$DOCKER_IMAGE_TAG_LONG .
 
-docker login -u $DOCKER_USER -p $DOCKER_AUTH
-echo "Pushing $DOCKER_TAG image"
-docker push $DOCKER_IMAGE_TAG_LONG
-docker tag $DOCKER_IMAGE_TAG_LONG $DOCKER_IMAGE_TAG
-docker push $DOCKER_IMAGE_TAG
+# ---
+# ONLY re-add these lines after removing the `improve-ci-speed` branch opt-in the GH Actions workflow!!
```
derhuerst: This needs to be removed before merging!

.github/workflows/dev-pipeline.yml
```diff
@@ -225,14 +416,17 @@ jobs:
           STORAGESASTOKEN: ${{ secrets.STORAGESASTOKEN }}
           STORAGERESOURCENAME: "visualtestsnapshots"
         run: yarn test:e2e
+
   publish-npm:
-    if: github.ref == 'refs/heads/v2'
+    # todo: re-add this before merging!
+    # if: github.ref == 'refs/heads/v2'
```
derhuerst: This needs to be removed before merging!

.github/workflows/dev-pipeline.yml
```diff
@@ -262,24 +483,26 @@ jobs:
...
-        id: log-parse
-      - name: Send published packages to slack
-        if:  steps.log-parse.outputs.published-packages != ''
-        run: ./.github/workflows/scripts/slack_message.sh
-        env:
-          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
-          PUBLISHED_PACKAGES: ${{ steps.log-parse.outputs.published-packages }}
-        shell: bash
+      # todo: re-add this before merging!
+      # - name: Publish shared components to npm
```
derhuerst: This needs to be removed before merging!

.github/workflows/dev-pipeline.yml
```diff
@@ -262,24 +483,26 @@ jobs:
...
+      #   run: ./.github/workflows/scripts/slack_message.sh
+      #   env:
+      #     SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+      #     PUBLISHED_PACKAGES: ${{ steps.log-parse.outputs.published-packages }}
+      #   shell: bash
 
   docker-push:
-    if: github.ref == 'refs/heads/v2'
+    # todo: re-add this before merging!
+    # if: github.ref == 'refs/heads/v2'
```
derhuerst: This needs to be removed before merging!

vesameskanen:
It seems that the new favicon setup is broken. I built the UI server locally (yarn build), and run it with a config, for example:

CONFIG=tampere NODE_OPTS='--title=digitransit-ui' yarn start

Chrome debug tools indicate that no favicons are found (application/manifest). Broken paths  in the manifest are of type:
 
http://localhost:8080/icons/android-chrome-36x36.png

Correct icon path from v3 branch looks like:

http://localhost:8080/assets/icons-tampere-70020a06183cc7648f25613739dcb137/android-chrome-36x36.png

vesameskanen:
A further note: all required favicons are generated, only the paths are wrong.

hbruch:
Upgrading `webpack-assets-manifest` to `^4.0.0` (with `html-webpack-plugin` resolved to `4.5.2`) recreated the correct icon paths in the `manifest.json` (see https://github.com/embarkokc/digitransit-ui/pull/26)

derhuerst:
> It seems that the new favicon setup is broken. […] Chrome debug tools indicate that no favicons are found (application/manifest). […]

It seems like the HTTP request for the favicon is redirected wrongly:

```shell
curl 'http://localhost:8080/icons/android-chrome-36x36.png' -L -I
# HTTP/1.1 302 Found
# X-Powered-By: Express
# Set-Cookie: lang=fi; Path=/
# Location: /reitti/icons/android-chrome-36x36.png
# Vary: Accept
# Content-Type: text/plain; charset=utf-8
# Content-Length: 60
# Date: Wed, 29 Mar 2023 12:45:18 GMT
# Connection: keep-alive
# Keep-Alive: timeout=5

# HTTP/1.1 200 OK
# X-Powered-By: Express
# Set-Cookie: lang=fi; Path=/
# content-type: text/html; charset=utf-8
# Date: Wed, 29 Mar 2023 12:45:18 GMT
# Connection: keep-alive
# Keep-Alive: timeout=5
```

vesameskanen:
Right. Correct path would be `assets/icons-<theme>-<hash>/<imagename>`, not` icons/<imagename>`. 

Yarn build creates all required icons into the correct path, but somehow the build ends up configuring  wrong path.

derhuerst:
In 6db1b69f1, I had changed the `yarn install` to use [a Docker cache mount (`RUN --mount=type=cache`)](https://docs.docker.com/engine/reference/builder/#run---mounttypecache), but it doesn't seem to work with GitHub Actions, failling with `rm -rf node_modules/.cache: resource busy`. I have removed it from this PR.

PR #4547, opened 2022-04-21
DT-5633 : Use long name of route when short name is unavailable

nurmAV:
## Proposed Changes

  - Use  the long name of a route in a route page if the short name is unavailable

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen (commented):
Did youactually  test what happens after this change? 

Logic seems to be reversed. shortName is not shown if it exists.

PR #4583, opened 2022-05-23
DT-5400: Search does not show network specific citybike icon

MikkoPuustinen:
## Proposed Changes

  - Autosuggest takes a function as a prop which can be used for determining the icon and color displayed for each bike rental station
  - Two icon options exist: one with a dark background and a colored bicycle and one with a colored background and a dark bicycle

## Pull Request Check List

## Review

  - Merge the pull request


app/configurations/config.hsl.js
```diff
@@ -100,6 +100,17 @@ export default {
...
   },
+  getAutoSuggestIcons: {
+    citybikes: station => {
+      if (station.properties.source === 'citybikessmoove') {
+        return ['citybike-stop-default', '#f2b62d'];
+      }
+      if (station.properties.source === 'citybikesvantaa') {
+        return ['citybike-stop-default-secondary', '#f2b62d'];
+      }
+      return ['citybike-stop-default', '#f2b62d'];
```
vesameskanen, line 108: One 'if .. else return default' would be enough.

vesameskanen (commented):
New structured prop for  layer specific callbacks is a good idea. Maybe we someday refactor the crazy if-else jungle into a clean mapping where all layers systematically use a callback.

Some observations:
- Opening Favorite dropdown crashes immediately
- Document the new prop like other props

joonaojapalo:
> New structured prop for layer specific callbacks is a good idea. Maybe we someday refactor the crazy if-else jungle into a clean mapping where all layers systematically use a callback.

Sounds good! 👍 Just wondering if in the long run defining a configuration directive as an executable code will lead to logic leaking from application code to configuration files.

PR #4601, opened 2022-07-06
DT-5399: Configure varely (seutuplus) otp and stop & station map service

joonaojapalo:
## Proposed Changes

  - Fix "varely" (reittiopas.seutuplus.fi) configuration to use specific OTP v2 instance and respective vector map layer for stops 

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen (commented):
Remove process.env.OTP_URL  option to force varely otp routing. OTP_URL points to waltti OTP because varely UI is run by waltti UI server instance.

app/configurations/config.varely.js
```diff
@@ -54,78 +64,78 @@ export default configMerger(walttiConfig, {
 
   // Varsinais-Suomi (acquired from: https://github.com/varmais/maakunnat/blob/master/maakunnat.geojson?short_path=81dbd20#L184)
   areaPolygon: [
```
joonaojapalo, line 66: Polygon vertex precision dropped as requested. @optionsome

PR #4611, opened 2022-08-04
DT-5344 - Add i18n to OTP2 requests

Antiik91:
## Proposed Changes

  -

## Pull Request Check List

## Review

  - Merge the pull request


app/configurations/config.hsl.js
```diff
@@ -20,7 +20,7 @@ export default {
 
   URL: {
     OTP: process.env.OTP_URL || `${API_URL}/routing/v1/routers/hsl/`,
-    STOP_MAP: `${MAP_URL}/map/${MAP_VERSION}/hsl-stop-map/`,
+    STOP_MAP: `${process.env.OTP_URL}vectorTiles/stops,stations/`,
```
joonaojapalo, line 23: This is very optional comment/idea.

Just thinking if, instead of string concatenation, we should build urls in such a manner that no assumption of trailing slash would be necessary. In existing codebase, there's a utility function https://github.com/HSLdevcom/digitransit-ui/blob/v2/app/util/path.js#L36 with reasonable test coverage and could be utilized as:

```javascript
import { pathJoin } = require('../util/path.js');

// ...
{
  //...
  STOP_MAP: pathJoin([process.env.OTP_URL, "vectorTiles/stops,stations"])
}
``` 

Preferably all url's should be created as  native javascript class `URL` (see: https://developer.mozilla.org/en-US/docs/Web/API/URL). For this instance, it might require unfeasible refactoring but maybe consider using `URL` in upcoming development.

app/util/fetchUtils.js
```diff
@@ -34,4 +34,19 @@ export const retryFetch = (URL, options = {}, retryCount, retryDelay) =>
...
+    }
+    return res;
+  });
+};
+
+export const fetchWithLanguage = (URL, language, addHeaders = {}) => {
+  const headers = { 'Accept-Language': language, ...addHeaders };
+  return fetchWithHeaders(URL, headers);
+};
+export default { retryFetch, fetchWithHeaders };
```
joonaojapalo, line 52: Default export this should be removed as module fetchUtil doesn't sound like it provides any natural export primitive and there's no default import statement to this in the codebase - yet.

joonaojapalo (changes requested):

optionsome:
This was implemented in two separate prs https://github.com/HSLdevcom/digitransit-ui/pull/4666 and https://github.com/HSLdevcom/digitransit-ui/pull/4665

PR #4612, opened 2022-08-05
Update GraphQL schema to latest OTP2

joonaojapalo:
## Proposed Changes

  - `schema.json` re-generated against https://dev-api.digitransit.fi/routing/v2/ (latest version of OpenTripPlanner's LegacyGraphQLApi)
  - new fields enables eg. usage of detailed error messages via`Plan.routingErrors`

## Review

  - Merge the pull request


app/component/map/tile-layer/bikeParks.js
```diff
@@ -2,11 +2,11 @@ import { graphql } from 'react-relay';
 
 export default graphql`
   query bikeParksQuery($ids: [String!]!) {
```
optionsome: Is this used somewhere? I tried to do a search in the code base for bikeparks but I found nothing.

optionsome (commented):
The schema exists also in the @digitransit-search-util/digitransit-search-util-query-utils package. I think we should decide how we will go about with the versioning of the packages in this otp2 branch. However, I think we could maybe delay that decision for a week or so, and I don't think we necessarily need to bump up the package version now if update the schema package. I think it should be taken into use in this branch regardless if we bump up the version or not, as no one else currently needs the version of the package with the new schema.

app/component/map/tile-layer/bikeParks.js
joonaojapalo: Should I remove this as unused code? Indeed, there seems to be not a single reference to query.

joonaojapalo:
> I think we should decide how we will go about with the versioning of the packages in this otp2 branch.

Maybe bumping the major version could work here? But as said, this becomes relevant when drafting a production-like release.

PR #4615, opened 2022-08-16
Render OTP2 routing error messages (DT-5405)

joonaojapalo:
## Proposed Changes

  - Rendering of OTP2 error messages based on the design (https://www.figma.com/file/dXPBr0Ow9BcOEi4hXqeca5/Summary?node-id=1401%3A3745) 
  - Refactoring of `ItinerarySmmaryList` to subcomponents in new subdirectory.
  - Unit tests for error message matching login.
  - Improved package.json scripts to enable more efficient test driven approach. Eg. `yarn test-unit:app -g 'ComponentName' -- --watch` to watch and run of single test case.

Related issue: DT-5405

## Pull Request Check List

## Review

  - Merge the pull request


joonaojapalo:
All required translations are included.

README.md
```diff
@@ -1,39 +1,69 @@
...
 
 Digitransit-ui is tested to work on the latest and the second latest major versions of Firefox, Chromium and Safari.
+
 - Continuous Integration: [https://github.com/HSLdevcom/digitransit-ui/actions](https://github.com/HSLdevcom/digitransit-ui/actions)
 
 Visual tests are run with jest and playwright.
 
 More information about [testing](docs/Tests.md).
 
+### Unit testing
```
optionsome: There is similar text already in Tests.md file. Should that be updated instead/in addition to this?

app/component/ItinerarySummaryListContainer/ItinerarySummaryListContainer.js
```diff
@@ -2,22 +2,27 @@ import PropTypes from 'prop-types';
 import React, { useState } from 'react';
```
optionsome, line 2: Is there some best practice for putting these components into directories? This structure hasn't been used before in this project (or only once in some recent update). I like that the files are grouped together. Only questions I have: is it necessary/good to have the index.js in each directory? Why is there a nested component structure? Is it because a component uses some other component which then uses some components? Some of the files in the components directory are more like util classes rather than components, should there be some different directory for those?

app/component/ItinerarySummaryListContainer/components/components/Link.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { FormattedMessage } from 'react-intl';
+import cx from 'classnames';
+
+const Link = ({ msgId, href }) => (
```
optionsome: We already use the Link component from found in a lot of places which is meant for links within this site. Should we name this something like ExternalLink so the difference between this and the found Link is more obvious. I was quite surprised there wasn't already a component like this somewhere. I would assume something like this would also be needed in other places. Is putting this component inside this ItinerarySummaryListContainer directory the right place? Is this generic enough so it could also be used elsewhere?

app/component/ItinerarySummaryListContainer/components/errorCardProperties.js
```diff
@@ -0,0 +1,205 @@
...
+  {
+    id: 'no-stops-in-range-3',
+    props: {
+      titleId: 'router-unable',
+      bodyId: 'router-stops-in-range-3',
+      ...info,
+    },
+  },
+  {
+    id: 'walk-bike-itinerary-4',
```
optionsome: why is 4 before 1 in this list?

app/translations.js
```diff
@@ -1120,7 +1120,7 @@ const translations = {
     'itinerary-hide-stops': 'Hide stops',
     'itinerary-in-the-past':
       'The route search falls within a period that is in the past.',
-    'itinerary-in-the-past-link': 'I want to travel now ›',
+    'itinerary-in-the-past-link': 'Depart now ›',
```
optionsome, line 1123: Why does this have the arrow in the translation while some others don't even though it should also be shown with "Change departure time", for example?

optionsome (commented):
Haven't tested this properly yet, should read through the code. I like that the big components are split into smaller pieces. I merged the current v3 into this and pushed it.

app/component/ItinerarySummaryListContainer/ItinerarySummaryListContainer.js
joonaojapalo: Good points!

In this PR, `index.js` file keeps the import statement untouched.

Splitting big component to smaller components with more strict responsibilities yields many components and I think it's natural to store these hierarchially rather than as a flat list and let the directory layout help with code-base navigation. Eg. SummaryPage, apporx. 2400 LOC probably could be splitted to dozens of child components. In hsl.fi project there's a good example of using nesting: [CityBikesRentalsBlock](https://dev.azure.com/hslfi/hslfi/_git/hslfi-frontend?path=%2FFrontend%2Fsrc%2Fblocks%2FCityBikesRentalsBlock&version=GBdevelop&_a=contents) (they seem to use one index.ts file per directory). This nested hierarchy definitely needs some rules simplify code-base navigation rather than making it more complex. Natural basic rule is that components should only import components from *subdirectory or from parent directories*, but *not from any parent's subdirectory*. This way generic components would live higher in the directory hierarchy. Also it would enable extracting a child component without need to worry about adding noise/complexity to directory listings.

I think that classification simple and pure utility components from more fat container like orchestrator components boils down to a debated question of how to differentiate [presentational vs container components](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0). It seems that the current practice is not to make this separation and just treat all components equally.

Just wondering should we run a single automated batch operation to convert all current `app/components/Xxx.js` to `app/component/Xxx/{Xxx.js,index.js}` structure to make project layout homogenous?

app/component/ItinerarySummaryListContainer/components/errorCardProperties.js
joonaojapalo: This directly from priority definition: https://github.com/HSLdevcom/digitransit-ui/blob/v2/app/component/ItinerarySummaryListContainer.js#L263

app/component/ItinerarySummaryListContainer/components/components/Link.js
joonaojapalo: Actually there already exists the component https://github.com/HSLdevcom/digitransit-ui/blob/v2/app/component/ExternalLink.js but it has a different DOM structure and CSS definitions which affect positioning. There are 18 external anchor tags in the code base (`<a>`) but almost all of them have individual styling.  I think it's better to substitute this back to `ChangeDepartureTimeLink` and `PastLink`.

app/translations.js
joonaojapalo: Carets are indeed missing from some translation strings - now added.

optionsome:
Seems like some translations need to be updated. At least for NO_TRANSIT_CONNECTION the swedish translations were incorrect

app/component/ItinerarySummaryListContainer/ItinerarySummaryListContainer.js
optionsome: I'm not sure if the file `findErrorMessageIds.js` is either presentational or container component as it just contains utility functions but no react.

I think we first try use this directory structure for a bit and see if we find any issues with it, if not, then we could do the refactoring to change all the components to be in directories.

joonaojapalo: Indeed, I think `utils` subdirectory will work fine. 👍  I read the comment too quickly and not properly checked the exact files in question, excuse my misunderstanding.

optionsome (commented):
NO_TRANSIT_CONNETION and NO_TRANSIT_CONNECTION_IN_SEARCH_WINDOW swedish translation had wrong word in title: ruttförslag -> ruttalternativ.

NO_TRANSIT_CONNETION and NO_TRANSIT_CONNECTION_IN_SEARCH_WINDOW finnish title had "." at the end of the title even though it shouldn't.

Compared to production reittiopas, the OUTSIDE_BOUNDS type errors are now rarely returned. I think it makes more sense to return those when we are outside of the service area than the LOCATION_NOT_FOUND type errors.

I think we could display the system error message when routing returns 5xx error or GraphQL error that starts with "Exception while fetching data" (if there are no itineraries returned).

joonaojapalo:
> NO_TRANSIT_CONNETION and NO_TRANSIT_CONNECTION_IN_SEARCH_WINDOW swedish translation had wrong word in title: ruttförslag -> ruttalternativ.
>
> NO_TRANSIT_CONNETION and NO_TRANSIT_CONNECTION_IN_SEARCH_WINDOW finnish title had "." at the end of the title even though it shouldn't.

Fixed!

> Compared to production reittiopas, the OUTSIDE_BOUNDS type errors are now rarely returned. I think it makes more sense to return those when we are outside of the service area than the LOCATION_NOT_FOUND type errors.

Fixed! Added origin and destination point client-side validation against `config.areaPolygon` property.

> I think we could display the system error message when routing returns 5xx error or GraphQL error that starts with "Exception while fetching data" (if there are no itineraries returned).

Verified that SYSTEM_ERROR is given on HTTP error status. However, it seems that in case GraphQL server responds with partially sensible payload but contains an error, `react-relay` doesn't propagate payload to wrapped component even though [specification should allow this](https://spec.graphql.org/June2018/#sec-Errors). I think this behavior could be tolerated.

optionsome:
Noticed a few thing while retesting this.
* At least the system error was displayed even though some itineraries were returned. I have some odd memory that maybe the props were not passed to the to components if there was an error but maybe it was just when the request was sent as part of a found `Route`. Hopefully we can render the result even if there are errors. I tested this running otp locally where I had edited https://github.com/opentripplanner/OpenTripPlanner/blob/dev-2.x/src/ext/java/org/opentripplanner/ext/legacygraphqlapi/datafetchers/LegacyGraphQLItineraryImpl.java#L73-L76 to be like:
```@Override
  public DataFetcher<Long> startTime() {
    return environment -> {
      if (getSource(environment).startTime().toInstant().toEpochMilli() > 0) {
        throw new RuntimeException();
      }
      return getSource(environment).startTime().toInstant().toEpochMilli();
    };
  }
```
* At least the OUTSIDE_SERVICE_PERIOD causes some rendering issues in mobile. It just renders an empty error box.
* We could change the links to opas.matka.fi include the user's language like opas.matka.fi/sv/ or opas.matka.fi/en/

app/configurations/config.hsl.js
```diff
@@ -117,7 +117,20 @@ export default {
...
 
-  nationalServiceLink: { name: 'matka.fi', href: 'https://opas.matka.fi/' },
+  nationalServiceLink: {
+    fi: {
+      name: 'matka.fi',
+      href: 'https://opas.matka.fi/',
+    },
+    sv: {
+      name: 'matka.fi',
+      href: 'https://opas.matka.fi/?locale=sv',
```
optionsome:
```suggestion
      href: 'https://opas.matka.fi/sv/',
```

app/configurations/config.hsl.js
```diff
@@ -117,7 +117,20 @@ export default {
...
+    },
+    en: {
+      name: 'matka.fi',
+      href: 'https://opas.matka.fi/?locale=en',
```
optionsome:
```suggestion
      href: 'https://opas.matka.fi/en/',
```

app/configurations/config.waltti.js
```diff
@@ -126,7 +126,20 @@ export default {
...
 
-  nationalServiceLink: { name: 'matka.fi', href: 'https://opas.matka.fi/' },
+  nationalServiceLink: {
+    fi: {
+      name: 'matka.fi',
+      href: 'https://opas.matka.fi/',
+    },
+    sv: {
+      name: 'matka.fi',
+      href: 'https://opas.matka.fi/?locale=sv',
```
optionsome:
```suggestion
      href: 'https://opas.matka.fi/sv/',
```

app/configurations/config.waltti.js
```diff
@@ -126,7 +126,20 @@ export default {
...
+    },
+    en: {
+      name: 'matka.fi',
+      href: 'https://opas.matka.fi/?locale=en',
```
optionsome:
```suggestion
      href: 'https://opas.matka.fi/en/',
```

test/unit/component/ItinerarySummaryMessage.test.js
```diff
@@ -0,0 +1,322 @@
...
+  areaPolygon: HelsinkiRegionBounds,
+  minDistanceBetweenFromAndTo: 100.0,
+  nationalServiceLink: {
+    fi: {
+      name: 'matka.fi',
+      href: 'https://opas.matka.fi/',
+    },
+    sv: {
+      name: 'matka.fi',
+      href: 'https://opas.matka.fi/?locale=sv',
```
optionsome:
```suggestion
      href: 'https://opas.matka.fi/sv/',
```

test/unit/component/ItinerarySummaryMessage.test.js
```diff
@@ -0,0 +1,322 @@
...
+    },
+    en: {
+      name: 'matka.fi',
+      href: 'https://opas.matka.fi/?locale=en',
```
optionsome:
```suggestion
      href: 'https://opas.matka.fi/en/',
```

optionsome (commented):
If I remember correctly, there was some slight difference between linking with locale param vs the `/<lang>/` url path and I think the path was better options. Tested and everything seems to work nicely now.

PR #4618, opened 2022-08-22
Change HSL mqtt broker in dev for testing purposes

vesameskanen:


vesameskanen:
Broker can be changed using ENV var called REALTIME_PATCH.

PR #4619, opened 2022-08-22
Typescript

joonaojapalo:
## Proposed Changes

  -

## Pull Request Check List

## Review

  - Merge the pull request


optionsome (commented):
I think we should enable tests for pull requests against this otp2 branch. However, I think we maybe should rename the otp2 branch to be v3 instead so it follows normal versioning (although it's not as clear what the changes are intended for in the branch) before we change the github actions to run against pull requests.

I noticed a couple of things, the .nycrc.json file should be updated so it takes typescript files into account. Also, I think the relay does not work for typescript files currently. I think the support for typescript has improved in the react-relay version 13. We could try updating the relay related dependencies instead of doing some extra configuration/adding dependencies that will no longer be needed later on. We should also look into what would be the most optimal way for us to use the graphql types for typescript (maybe it can be done directly through relay?). I'm not sure if the relay/graphql related changes should be done in this pr or in a separate pr as for files that don't use relay this basic setup seems to work.

joonaojapalo:
> However, I think we maybe should rename the otp2 branch to be v3 instead so it follows normal versioning (although it's not as clear what the changes are intended for in the branch) before we change the github actions to run against pull requests.

@optionsome please see https://github.com/HSLdevcom/digitransit-ui/pull/4620. I already renamed the branch otp2 to v3 in Github UI (by accident), but didn't revert as this seems to be our desired state.

PR #4620, opened 2022-08-24
Rename workflow branch and Docker image from otp2 to v3

joonaojapalo:
## Proposed Changes

  - Rename workflow branch to v3
  - Rename pushed Docker image to v3

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/.github/workflows/v3-dev-pipeline.yml
+++ b/.github/workflows/v3-dev-pipeline.yml
@@ -1,11 +1,11 @@
-name: Deploy to otp2-dev
+name: Deploy v3-dev
 on:
   push:
     branches:
-      - otp2
+      - v3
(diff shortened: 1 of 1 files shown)


optionsome (commented):
I changed this pull request to be against v3 branch because I think that was the intended branch. I think we should enable (some?) tests for the workflow and run it against pull requests also. Also, the way the docker image is built, does not create a 100% working image as it's missing https://github.com/HSLdevcom/digitransit-ui/blob/v2/.github/workflows/scripts/build_and_push_dev.sh#L15 .

Also, should we now create a workflow for releases against v3 branch because I think we might need to do it soon for opas.waltti.fi?

PR #4636, opened 2022-09-21
fix: disable viapoints with OTP2 DT-5517

tekoiv:
Disable viapoint routing in default config


vesameskanen (changes requested):
Configuration change is not enough. The viapoint switch has one leak which needs to be fixed:
- click a stop on map
- on stop page, open 'reitti pysäkille' dropdown. It allows setting a via point.

Stop page dropdown should obey viapoint switch and hide the  intermediate point option.

PR #4643, opened 2022-09-23
fix: reallow itinerary direction switch DT-5528

tekoiv:
Viapoint specific code commented out in case viapoints become supported by OTP2 in the future.


vesameskanen (changes requested):
layout of origination-destination fields seem to have changed. There is no space between them and grey vertical connecting bar at the left is missing.

PR #4647, opened 2022-09-29
Re-fetch more itineraries (part 1)

joonaojapalo:
Related issue: DT-5464

## Proposed Changes

  - Re-fetch more itineraries on using initial SummaryPage graphql query, if initial query responds with less than 5.
  - Re-fetch more itineraries when user request more earlier or later and response contains less than 3.
  - Refactor: SummaryPage.js to subdirectory
  - Refactor: graphql refetch logic move from SummaryPage component to higher-order-components.
  - Relay bundle middleware removed as currently providing only marginal performance improvement and som issues with pageCursor queries.
  - Partial result rendering implementation will be splitted to another PR

## Pull Request Check List

## Review

  - Merge the pull request


app/component/SummaryPage/hoc/withFetchMoreItineraries.js
```diff
@@ -0,0 +1,312 @@
...
+
+      if (state.arriveBy) {
+        setLaterItineraries([...laterItineraries, ...itineraries]);
+      } else {
+        // Reverse the results so that route suggestions are in ascending order
+        const reversedItineraries = itineraries
+          .slice() // Need to copy because result is readonly
+          .reverse()
+          .filter(
+            itinerary => !itinerary.legs.every(leg => leg.mode === 'WALK'),
```
optionsome: I think we could actually leave the WALK mode out of the public transportation requests because OTP2 returns walk legs for public transportation even without it. Then we wouldn't need to do this filtering in UI which could simplify the logic a bit.

app/component/SummaryPage/hooks/useAutofetchQuery.js
```diff
@@ -0,0 +1,238 @@
...
+
+  const ACTION_DATA = 'data';
+  const ACTION_START = 'start';
+  const ACTION_ERROR = 'error';
+  const ACTION_DATA_COMPLETE = 'data-complete';
+
+  const STATUS_PROCESSING = 'PROCESSING';
+  const STATUS_COMPLETE = 'COMPLETE';
+  const STATUS_FETCHING = 'FETCHING';
+  const STATUS_READY = 'READY';
```
optionsome: Should we change these to be in the enum object format and maybe in some different file we can, for example, check that the status is something like `status === AutofetchQueryStatuses.Complete`?

app/component/SummaryPage/hooks/useAutofetchRelay.js
```diff
@@ -0,0 +1,213 @@
...
+import createCancelable from '../utils/createCancelable';
+
+const ACTION_MOUNTED = 'MOUNTED';
+const ACTION_DATA_UPDATE = 'DATA_UPDATE';
+const ACTION_QUERY_UPDATE = 'QUERY_UPDATE';
+const ACTION_ERROR = 'ERROR';
+
+const STATUS_FETCHING = 'FETCHING';
+const STATUS_REFETCHING = 'REFETCHING';
+const STATUS_COMPLETE = 'COMPLETE';
```
optionsome: Same here as above.

optionsome (commented):
Tested this quickly and found a couple of problems. Changing something from settings causes no new normal public transportation plan query to be sent. If the search finds no itineraries because no itineraries can be found within the search window, the next page query seems to be canceled.

app/component/SummaryPage/hoc/withFetchMoreItineraries.js
optionsome: We won't do this because the WALK only itineraries are now sometimes rendered.

PR #4648, opened 2022-09-30
feat: matka.fi separate park and ride options for car and bike

tekoiv:


app/configurations/config.matka.js
```diff
@@ -288,7 +288,9 @@ export default {
 
   includeCarSuggestions: true,
   includeParkAndRideSuggestions: true,
-
+  // DT-5469
+  includeBikeParkSuggestions: true,
```
vesameskanen: one switch is enough. Use existing showBikeAndParkItineraries.

vesameskanen (changes requested):
- Car parkride switch has disappeared from HSL planner
- Matka planner requests bikeparkride itinerariers even when the respective switch is turned off. This is a computation intensive feature so bikeparkplan should be included only when user asks them.

vesameskanen (approved):

app/component/customizesearch/StreetModeSelectorPanel.js
```diff
@@ -62,43 +66,75 @@ const StreetModeSelectorPanel = (
...
                     }
                   />
-                </label>
+                </span>
+              </label>
+            </div>
+          ) : null}
+        </div>
+        {config.includeParkAndRideSuggestions &&
+          !config.separatedParkAndRideSwitch && (
```
derhuerst, line 109: I'm not sure I understand the logic, but shouldn't this boolean flag work in the exact opposite way? As in: If the "separated switch" config is `true`, it should show the switch in the UI?
We stumbled upon this while merging changes into the [bbnavi code](https://github.com/bbnavi/digitransit-ui).

vesameskanen: In reittiopas.hsl.fi, car symbol means park and ride. There is no plain car routing available.
in other planners ( say tampere.digitransit.fi ), car symbol means car routing and there is additional switch for park & ride.

Configuration for e.g. tampere says: 
  // Park and ride and car suggestions separated into two switches                                                                                                                                         
  separatedParkAndRideSwitch: true,

So the switch naming logic seems to be right. The rendering code might be strange as and identical element is used for two different purposes.

Such inconsistency is unfortunate but comes from customer's design.

derhuerst: Oh, we missed the part below, thanks for pointing this out. In any case, I'd strongly advocate for DRYing the two (quite similar but subtly different) code sections.

PR #4659, opened 2022-10-18
Turku funicular DT-5423

tekoiv:
## Proposed Changes

  - Feature: UI Turku funicular

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen (commented):
Some quick comments:
1. It seems that all workspace package versions have been changed. This is not necessary. First you bump up versions of libraries you actually change.  Next you have to update libraries which depend on those changed libs. Then you change libs which depend on libs changed by dependency update, an so on. For example,  get-json search utility surely has nothing to do with funicular.
2.  svg-sprite.default.svg has invalid format. g blocks in funicular icon do not match.
3. This feature was intended for current production. i.e. UI v2 branch. Rewriting it on top of v2 might be so laborious task that maybe we wait until otp2 is released.

PR #4660, opened 2022-10-20
fix: fixing rendering error when fetching favourites DT-5562

tekoiv:
## Proposed Changes

  - Check when favorites have been fetched, pass props accordingly. Using state is not needed so let's not

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/StopsNearYouPage.js
+++ b/app/component/StopsNearYouPage.js
@@ -46,6 +46,7 @@ import {
   getNearYouModes,
   useCitybikes,
 } from '../util/modeUtils';
+import FavouriteStore from '../store/FavouriteStore';
 
 // component initialization phases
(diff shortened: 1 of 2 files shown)


app/component/StopsNearYouPage.js
```diff
@@ -1005,6 +991,7 @@ const PositioningWrapper = connectToStores(
       favouriteStopIds,
       favouriteBikeStationIds,
       favouriteStationIds,
+      favouritesFetched: status === FavouriteStore.STATUS_HAS_DATA,
```
vesameskanen: If favourite fetch fails, the view will render a spinner forever

vesameskanen (changes requested):

PR #4665, opened 2022-10-31
DT-5475 & DT-5565 Use map layers from OTP instead of map server

optionsome:
## Proposed Changes

  - Use OTP2 map layers instead of layers from hsl-map-server
  - Changes bike park map to behave the same way as car park
  - Updates to bike rental stations are now fetched through vector tiles directly and every 60 seconds instead of 30 seconds
     - There is a small edge case where realtime information is still fetched for a highlighted station when zoomed out
  - Adds accept-language header to vector tile requests so we get translations
  - Removes unused code

## Pull Request Check List

## Review

  - Merge the pull request


app/component/map/tile-layer/BikeRentalStations.js
```diff
@@ -0,0 +1,207 @@
...
+      (!this.timeOfLastFetch ||
+        currentTime - this.timeOfLastFetch > REALTIME_REFETCH_FREQUENCY)
+    ) {
+      this.fetchAndDraw(lang);
+    }
+  };
+
+  shouldShowStation = (id, network) =>
+    (this.tile.stopsToShow && this.tile.stopsToShow.includes(id)) ||
+    showCitybikeNetwork(this.config.cityBike.networks[network]);
```
vesameskanen, line 202: This logic is now wrong. Favorite stops near you page shows alwasy all bike stations.

vesameskanen (commented):
Language support has bugs.  Not all marker popups update their  language after switching the language. Tested with bike parks -  popup opened with previous language , not the current one.

PR #4671, opened 2022-11-04
fix: fix faulty suggestion sorting DT-5593

tekoiv:
## Proposed Changes

  - Previously sorting has worked when the amount of search results is less than 100 because of the following reason: The process of getting search results is first getting all of the results, then filtering the best 100, then sorting them again by LayerType. The first filtering has been a normal filter which means that numbers are always favored before letters. When the result has more than 100 entries, the faulty first filter removes correct suggestions. Let's replace this ordering with a working one.
  

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/digitransit-search-util/packages/digitransit-search-util-query-utils/index.js
+++ b/digitransit-search-util/packages/digitransit-search-util-query-utils/index.js
@@ -10,11 +10,9 @@ import {
   mapRoute,
   isStop,
   getLayerRank,
-  match,
-  LayerType,
 } from '@digitransit-search-util/digitransit-search-util-helpers';
 import filterMatchingToInput from '@digitransit-search-util/digitransit-search-util-filter-matching-to-input';
-import { isString, orderBy } from 'lodash';
+import { orderBy } from 'lodash';
 
 let relayEnvironment = null;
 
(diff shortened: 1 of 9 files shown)


digitransit-search-util/packages/digitransit-search-util-query-utils/index.js
```diff
@@ -474,14 +478,52 @@ export function getRoutesQuery(input, feedIds, transportMode, pathOpts) {
...
+      results,
+      [
+        result => {
+          const { confidence, layer, source } = result.properties;
+          if (normalizedTerm.length === 0) {
+            // Search with an empty string
+            return getLayerRank(layer, source);
+          }
+
+          if (!confidence) {
```
vesameskanen: this query is for OTP, not geocoder, so confidence is always undefined. All logic after this block can be deleted.

vesameskanen (changes requested):

digitransit-search-util/packages/digitransit-search-util-query-utils/index.js
```diff
@@ -482,42 +480,12 @@ export function getRoutesQuery(input, feedIds, transportMode, pathOpts) {
...
-          if (normalizedTerm.length === 0) {
-            // Search with an empty string
-            return getLayerRank(layer, source);
-          }
-
-          if (!confidence) {
-            // not from geocoder
-            return (
-              getLayerRank(layer, source) +
-              match(normalizedTerm, result.properties)
```
vesameskanen: match(normalizedTerm, result.properties) is the only relevant sort action.

PR #4687, opened 2022-11-25
fix: fine tuning funicular and ferry DT-5423

tekoiv:
## Proposed Changes

  - Static constant operation timetable messages in trip view, stop view and route views
  - Redirect to stop page instead of trip view if the route is constant operation
  - Change Turku config: constant operation configs and tuning funicular and ferry with modeWeight
  - Update single test
  - Improve logic behind near you button styles

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/configurations/config.turku.js
+++ b/app/configurations/config.turku.js
@@ -245,11 +245,10 @@ export default configMerger(walttiConfig, {
   },
   strongWalkReluctance: {
     weights: {
-      FUNICULAR: 0.25,
+      FUNICULAR: 0.1,
     },
-    walkReluctance: 15,
   },
   customWeights: {
-    FERRY: 0.8,
+    FERRY: 0.6,
   },
 });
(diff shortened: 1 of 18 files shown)


app/configurations/config.turku.js
```diff
@@ -203,4 +233,23 @@ export default configMerger(walttiConfig, {
...
+  },
+  constantOperationRoutes: {
+    'FOLI:180': CONSTANT_OPERATION_PARAGRAPHS.fori,
+    'FUNI:2': CONSTANT_OPERATION_PARAGRAPHS.funi,
+  },
+  strongWalkReluctance: {
+    weights: {
+      FUNICULAR: 0.25,
+    },
+    walkReluctance: 15,
```
optionsome: I think having this high walkReluctance value will slow down the routing by quite a bit. Could the same effect be achieved by lowering the funicular weight even more?

app/util/planParamUtil.js
```diff
@@ -259,6 +259,12 @@ export const preparePlanParams = (config, useDefaultModes) => (
   );
 
   const cookies = new Cookies();
+
+  let modeWeight;
```
vesameskanen: variable and if  block not needed, use customWeights directly

PR #4700, opened 2022-12-19
fix: fix ticket type bug DT-5643

tekoiv:
## Proposed Changes

  - All modes query is made on page load and the result is compared to the result made with selected settings. If they don't match and no itineraries are found, alternative itineraries are shown. If they don't match and some itineraries are found, an info box informing that the selected settings are limiting the results is rendered below the list of itineraries. We need to treat ticket types as modes when querying with default settings, i.e. don't include ticket type in all modes query. 

## Pull Request Check List

## Review

  - Merge the pull request


optionsome:
The "all modes" query also gets send when user has only changed walk reluctance or transfer reluctance but the query uses the user's settings instead of default settings for those values. Therefore, sending the requests when user has changed one of those settings but no others, doesn't make sense.

app/util/planParamUtil.js
```diff
@@ -255,6 +255,17 @@ export const preparePlanParams = (config, useDefaultModes) => (
     intermediatePlaceLocations,
   );
 
+  // Use defaults or user given settings
+  const ticketTypes = useDefaultModes
```
derhuerst, line 259: This change effectively changes the meaning of the `useDefaultModes` variable to "use default modes and some other selected (default) routing parameters". This subtle change confused us quite a bit.

---

The DX aside, what is the intended UX? The `useDefaultModes = true` query may return itineraries which the user did not expect?

> We need to treat ticket types as modes when querying with default settings, i.e. don't include ticket type in all modes query.

Why is that? (I don't understand the reasoning here, probably because I don't have some context knowledge.)

Maybe `useRelaxedRoutingPreferences` would be a more appropriate naming?

cc @vesameskanen

vesameskanen: Exactly. Your suggestion describes the purpose well, whereas useDefaultModes is misleading.

The context is that user may have set some settings which limit the search and UI tries to give a hint about that.

vesameskanen: I updated the name in v3.

PR #4702, opened 2023-01-02
DT-5654 Fix SummaryPage regressions

tekoiv:
## Proposed Changes

  - Don't show conflicting info messages (no showing alternative plans and "only walking routes were found" at the same time)
  - Don't make second allmodesquery if selected plan only consists of a walking itinerary
  - Fix isWalkingFastest(). There was a bug in determining the duration of park and ride. Also, consider park and ride as driving so the info message displays correctly. Example: previously if walking and park and ride were found, park and ride wasn't considered as driving --> "only walking itineraries found" was rendered, which is not true
  - isLoading() to determine when to render suggestions. If public itineraries are fetched but only walking itineraries are found, we need to wait to load alternative suggestions
  - Fix "only walking component" rendering logic
  - Make sure public itineraries are displayed right away IF not only walking itineraries are found

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/SummaryPage.js
+++ b/app/component/SummaryPage.js
@@ -355,6 +355,7 @@ class SummaryPage extends React.Component {
     this.origin = undefined;
     this.destination = undefined;
     this.expandMap = 0;
+    this.allModesQueryDone = false;
 
     if (props.error) {
       reportError(props.error);
@@ -1108,6 +1109,7 @@ class SummaryPage extends React.Component {
           this.setLoading(false);
           this.isFetching = false;
           this.setParamsAndQuery();
+          this.allModesQueryDone = true;
         },
       );
     });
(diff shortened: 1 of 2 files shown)


app/component/SummaryPage.js
```diff
@@ -1519,7 +1521,13 @@ class SummaryPage extends React.Component {
           alternativePlan: undefined,
         },
         () => {
-          if (relevantRoutingSettingsChanged(this.context.config)) {
+          const hasNonWalkingItinerary = this.selectedPlan?.itineraries?.some(
+            itinerary => !itinerary.legs.every(leg => leg.mode === 'WALK'),
+          );
+          if (
+            relevantRoutingSettingsChanged(this.context.config) &&
+            hasNonWalkingItinerary
```
derhuerst, line 1529: Is this really what's intended here? If we *guess* what's intended, shouldn't we check the user-configured modes (`getSettings().modes`) if there's *only* walking enabled? Is it really the right approach to inspect the returned itineraries?

We stumbled upon this when merging from HSL's repo into our bbnavi fork, and have trouble understanding why this change has been made, despite the PR's description. 😕 – Maybe we're missing some context here.

We think a good way forward would be to document such intricate logic – which effectively make up significant parts of digitransit-ui's (intended and unintended) behaviour – in a more permanent way, e.g. in `docs/Routing.md`.

cc @vesameskanen

PR #4747, opened 2023-02-14
chore: add new hameenlinna zone lines DT-5669

tekoiv:
## Proposed Changes

  - New zone lines geojson and updated config

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen (changes requested):
Would it be possible to inject manually same kind of double styling of zone lines as in hsl and some other zone lines?  That is, wide opaque line and narrow less opaque line in the middlle.  HSL zone lines grow when you zoom in, which is nice.

Check anyway that double styled result looks  OK; if the data contains overlapping zone edges, opaque styling may actually look ugly.

PR #4774, opened 2023-03-23
DT-5795 - Add routing feedback button

nurmAV:
## Proposed Changes

  -Adds a new routing feedback prompt component and places it between the already visible itineraries and newer itineraries once the earlier or later departure buttons are pressed

## Pull Request Check List

## Review

  - Merge the pull request


app/component/RoutingFeedbackPrompt.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import { FormattedMessage } from 'react-intl';
+import PropTypes from 'prop-types';
+
+const RoutingFeedbackPrompt = (props, { config }) => {
+  return config.useRoutingFeedbackPrompt ? (
+    <div className="routing-feedback-container">
+      <div style={{ padding: '15px' }}>
```
vesameskanen: put styles to one place, don't mix inline and styles defined in css file.

app/component/itinerary.scss
```diff
@@ -2389,3 +2389,22 @@ $font-print-decrease: 0.7;
...
+  align-items: center;
+  flex-direction: column;
+}
+
+.routing-feedback-link {
+  background: white;
+  width: 60%;
+  border-radius: 50vh;
+  padding: 5px;
+  color: #007ac9;
```
vesameskanen: use sass variables whenever possible, instead of hardcoding exact values. $hsl-blue = #007ac9

vesameskanen (changes requested):
It seems that mobile view renders feedback button always.

PR #4777, opened 2023-03-24
switch Yarn v3

derhuerst:
This PR includes the changes of and depends on PR #4776; Please review & merge that one first.

## Proposed Changes

- upgrade to Yarn v3.4.1 (and the version plugin to v3.1.3 because the old version isn't compatible)
- in the CI workflow spec, add `-v3` to the cache keys, in order to prevent incompatible caches from being restored
- since Yarn v3 seems to have changed the algorithm, **update all checksums in `yarn.lock`** 💥 (done via running just `yarn`)

## Pull Request Check List

  - [ ] Changes are documented or they are self explanatory
  - [ ] Code coverage does not decrease (unless measured incorrectly)

## Review

  - Merge the pull request

---

(This is the upstream equivalent of https://github.com/embarkokc/digitransit-ui/pull/22.)


vesameskanen:
This PR got already merged into V3 as a part of https://github.com/HSLdevcom/digitransit-ui/pull/4838 .

PR #4781, opened 2023-04-04
fix: don't override theme colors with values from hsl theme

langbein-daniel:
I noticed two things:

## 1)  Some theme colors get lost

The import of `@import "../../sass/themes/hsl/theme";` in `itinerary.scss` results in some theme colors being overridden by the HSL theme values.

With the above import line, notice how some text and icons are in the `$hsl-blue: #007ac9` color:

image

Without the above import, the green of the current theme is being used:

image

This can be resolved by removing (or changing) the import line.

## 2) HSL theme specific variable added

A HSL theme specific style variable is being used inside `.routing-feedback-link`.

```scss
  color:  $hsl-blue;
```

I think it is better to use theme independent style variables.

As the HSL theme sets the `$primary-color` to `$hsl-blue`, we can also use `$primary-color` and get the same color.

### DIFF
--- a/app/component/itinerary.scss
+++ b/app/component/itinerary.scss
@@ -1,5 +1,4 @@
 @charset "UTF-8";
-@import "../../sass/themes/hsl/theme";
 
 $topbar-height: 40px;
 $itinerary-tab-switch-height: 48px;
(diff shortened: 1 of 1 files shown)


langbein-daniel:
I think those were changes by @nurmAV, can you have a look please? Thanks!

optionsome:
Hi I can take a look at this. Thank you for fixing this issue, we noticed it as well today.

PR #4783, opened 2023-04-12
DT-5648 WalttiOpas realtime fixed, changing stop in near you map shows vehicles for the new stop

sharhio:
## Proposed Changes

  - When there are multiple feedIDs, the selection is done based on the feedID of the first route on the list of routes. In WalttiOpas it is possible to change between stops in areas far from each other where the feedIDs can be different. 
  - Updating the mqtt client is done whenever the place changes as different stops are related to different vehicles. This fixes an issue where changing the stop moves the map to a new location but still shows the vehicles of the previous stop. 
  - Minor fixes on units and variable types based on the prints in the console.
  - Installation instructions updated with versions that are known to work. Also, removed yarn version 1.22.0 from the recommendations as it is now known not to work correctly (issues with tests crashing).  

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/map/StopsNearYouMap.js
+++ b/app/component/map/StopsNearYouMap.js
@@ -57,13 +57,10 @@ const handleStopsAndStations = edges => {
 
 const getRealTimeSettings = (routes, context) => {
   const { realTime } = context.config;
-  let agency;
   /* handle multiple feedid case */
-  context.config.feedIds.forEach(ag => {
-    if (!agency && realTime[ag]) {
-      agency = ag;
-    }
-  });
+  const agency = context.config.feedIds.find(
+    ag => realTime[ag] && routes[0].feedId === ag,
+  );
   const source = agency && realTime[agency];
   if (source && source.active && routes.length > 0) {
     return {
@@ -166,7 +163,7 @@ function StopsNearYouMap(
     isFetching: false,
     stop: null,
   });
-  const prevMode = useRef();
+  const prevPlace = useRef();
   const { mode } = match.params;
   const isTransitMode = mode !== 'CITYBIKE';
   const walkRoutingThreshold =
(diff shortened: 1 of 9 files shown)


app/component/map/StopsNearYouMap.js
```diff
@@ -166,7 +163,7 @@ function StopsNearYouMap(
     isFetching: false,
     stop: null,
   });
-  const prevMode = useRef();
+  const prevPlace = useRef();
```
vesameskanen, line 166: Now place change seems to update vehicles on map, but it you switch tab e.g. in Tampere journey planner from bus to tram, new vehicles will not show up.

digitransit-component/packages/digitransit-component-autosuggest/src/index.js
```diff
@@ -269,7 +269,7 @@ class DTAutosuggest extends React.Component {
     mobileLabel: PropTypes.string,
     lock: PropTypes.func.isRequired,
     unlock: PropTypes.func.isRequired,
-    refPoint: PropTypes.object,
+    refPoint: PropTypes.string,
```
vesameskanen, line 272: One of the shared npm libraries changes here. In this situation, you must do the following:
- increase minor version in package.json of changed library (SEMVER numbering)
- update the dependency of every package.json referring to the changed library
- increase the version number referring to changed library in digitransit-component/packages/digitransit-component/package.json

In this case, cascading changes stop at autosuggest-panel library level. If you change a very low level library,  need to update dependencies above it can spread very wide. Every time  you touch a library in any way, version numbers must be increased. Unfortunately we do not have any automation for this laborious process.

vesameskanen (changes requested):

vesameskanen (approved):

PR #4796, opened 2023-05-02
DT-5372 Add city bike stations to itinerary map

sharhio:
## Proposed Changes

  - Show bike rental station map layer to citybike itinerary map. 

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/SummaryPage.js
+++ b/app/component/SummaryPage.js
@@ -507,16 +507,14 @@ class SummaryPage extends React.Component {
     });
   };
 
-  getListOfHiddenStops = (
-    itineraryContainsBikeRentalStation,
-    activeItinerary,
-  ) => {
+  getHiddenObjects = (itineraryContainsBikeRentalStation, activeItinerary) => {
+    const hiddenObjects = { vehicleRentalStations: [] };
     if (itineraryContainsBikeRentalStation) {
-      return activeItinerary?.legs
+      hiddenObjects.vehicleRentalStations = activeItinerary?.legs
         ?.filter(leg => leg.from.bikeRentalStation)
-        ?.map(station => station.from.name);
+        ?.map(station => station.from.bikeRentalStation.stationId);
     }
-    return [];
+    return hiddenObjects;
   };
 
   planHasNoItineraries = () =>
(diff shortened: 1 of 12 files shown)


app/component/SummaryPage.js
```diff
@@ -500,6 +500,25 @@ class SummaryPage extends React.Component {
...
+    });
+  };
+
+  getListOfHiddenStops = (
+    itineraryContainsBikeRentalStation,
+    activeItinerary,
+  ) => {
+    if (itineraryContainsBikeRentalStation) {
+      return activeItinerary?.legs
+        ?.filter(leg => leg.from.bikeRentalStation)
```
optionsome: I wonder if we should, just to be safe, also include leg.to.bikeRentalStation in case an itinerary ends at a bike rental station (I don't know if it can actually end there currently)

app/component/SummaryPage.js
```diff
@@ -500,6 +500,25 @@ class SummaryPage extends React.Component {
...
+        ?.map(station => station.from.name);
```
optionsome: You should use `stationId` instead of name here.

app/component/SummaryPage.js
```diff
@@ -1856,6 +1875,18 @@ class SummaryPage extends React.Component {
     }
     const onlyHasWalkingItineraries = this.onlyHasWalkingItineraries();
 
+    const itineraryContainsBikeRentalStation = filteredItineraries[
+      activeIndex
+    ]?.legs.some(leg => leg.from.bikeRentalStation);
```
optionsome: Same leg.to dilemma as above.

app/component/SummaryPage.js
```diff
@@ -1856,6 +1875,18 @@ class SummaryPage extends React.Component {
...
+
+    const mapLayerOptions = itineraryContainsBikeRentalStation
+      ? this.addBikeStationMapForRentalBikeItineraries(filteredItineraries)
+      : this.props.mapLayerOptions;
+
+    const stopsToHide = this.getListOfHiddenStops(
```
optionsome: Maybe instead of calling them stops we could be a bit more generic and call them `objectsToHide`. Then it should probably be of structure:
```
{
    vehicleRentalStations: [<list of ids>]
}
```
or alternatively just call it `vehicleRentalStationsToHide` but I'm not sure if it's better to have one prop with a mystery structure or multiple props :D.

app/component/map/tile-layer/BikeRentalStations.js
```diff
@@ -89,6 +89,7 @@ class BikeRentalStations {
               !this.features.some(feature =>
                 this.shouldShowStation(
                   feature.properties.id,
+                  feature.properties.name,
```
optionsome: This is not needed if we use id instead of name (just be sure that the id is actually in the same format as the one you pass as sometimes it's feedscoped and sometimes isnt.

PR #4804, opened 2023-05-09
DT-5746 and DT-3461: refactor alerts and fix random bugs

optionsome:
## Proposed Changes

  - Refactors alert handling so it's more up to date with OTP2's alerts and uses OTP's alert model instead of an internal mode that caused extra complexity
  - At least temporarily remove support for alerts targeted on Trips and Patterns and the old handling caused more confusion than had benefits and they are not currently used by any of our data producers (as far as I know)
  - Fix accessibility of route and stop links in the disruption listing
  - Try to get rid of all sorts of bugs where it was indicated that a stop or route had alerts but the disruption tab was empty or the other way around
  - Fix some issues with alerts not being shown on stops/routes when they should have been
  - Create a more permanent fix for the disruption list found inside the hamburger menu previously fixed with a quick fix in https://github.com/HSLdevcom/digitransit-ui/pull/4765
  - Fix color of routes in alternative leg suggestions if a route has a special color
  - Remove merging alerts generated from cancelations as it had issues with important information missing out

Hopefully the code is a bit cleaner now and we can add support for new alert entity selectors in the future and maybe create a better global view of alerts than the one currently found inside the hamburger menu.

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen (commented):
Stop view's alert tab crashes on some (not all) routes having canceled departures. Crashed e.g on train E and bus 549.

warning: Failed prop type: Invalid prop `cancelations[0].entities[0].__typename` of type `object` supplied to `AlertList`, expected `string`.
 
warning: Failed prop type: Invalid prop `entities[0].__typename` of type `object` supplied to `AlertRow`, expected `string`.
    in AlertRow (at AlertList.js:81)
 
warning: Failed prop type: Invalid prop `mode` of type `boolean` supplied to `RouteNumber`, expected `string`.
 
pops.mode.toLowerCase is not a function
    at RouteNumber (RouteNumber.js:24:25)

vesameskanen:
Trip view shows alert banner which is not visible in alert tab of the route page. This is somewhat surprising. Maybe data issue.

PR #4815, opened 2023-05-23
fix: dt-5833 allModesQuery made when expected

sharhio:
## Proposed Changes

The spinner no longer spins forever. 

- The spinner is _not_ shown when the relevant settings ("mode") has changed, as alternative routes are not fetched in that case (or if the all modes query has finished).
- Alternative query is also made when it is supposed to be made - when no itineraries or only walking itineraries are found (the boolean variable was inverted). 
- Renamed variables for clarity.
- Edited comment to reflect the change.

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/SummaryPage.js
+++ b/app/component/SummaryPage.js
@@ -349,7 +349,7 @@ class SummaryPage extends React.Component {
   constructor(props, context) {
     super(props, context);
     this.isFetching = false;
-    this.secondQuerySent = false;
+    this.walkBikeQuerySent = false;
     this.setParamsAndQuery();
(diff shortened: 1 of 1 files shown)


vesameskanen:
Infinite spinning is fixed. There is one regression, though:
- With HSL theme, make search steissi-suomenlinna
- Disable ferry from settings. No itineraries are found, but proper explanation message is shown.
-  Open time selector ->  itinerary search gets triggered immediately and alternative routes are suggested.

The error is not very serious, this is much better than infinite spin. Anyway, check out if the unexpected state change could be fixed.

PR #4868, opened 2023-10-02
DT-5977 Varely config

sharhio:
## Proposed Changes

  - Varely real-time data
  - Rauma data to Varely
  - Varely datat to Matka

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/configurations/config.matka.js
+++ b/app/configurations/config.matka.js
@@ -511,6 +511,11 @@ export default {
       sv: 'Raseborg',
       en: 'Raasepori',
     },
+    VARELY: {
+      fi: 'Seutu+',
+      sv: 'Seutu+',
+      en: 'Seutu+',
+    },
   },
   stopCard: {
     header: {
--- a/app/configurations/config.matka.js
+++ b/app/configurations/config.matka.js
@@ -92,6 +92,7 @@ export default {
     'Salo',
     'Pori',
     'Raasepori',
+    'VARELY',
   ],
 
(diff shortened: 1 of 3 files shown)


app/configurations/config.matka.js
```diff
@@ -510,6 +511,11 @@ export default {
       sv: 'Raseborg',
       en: 'Raasepori',
     },
+    VARELY: {
+      fi: 'Seutu+',
+      sv: 'Seutu+',
+      en: 'Seutu+',
+    },
```
optionsome: Not quite sure what to call this but probably Varsinais-Suomi. Seutu+ is more than this data set (and it sounds slightly weird if you are not aware of what it is and you are interested in alerts around Rovaniemi region for example :D).

app/configurations/config.varely.js
```diff
@@ -139,6 +139,19 @@ export default configMerger(walttiConfig, {
     ],
   },
 
+  sourceForAlertsAndDisruptions: {
+    VARELY: {
+      fi: 'Seutu+',
+      sv: 'Seutu+',
+      en: 'Seutu+',
+    },
```
optionsome: Same as above.

optionsome (changes requested):
Btw I wonder if there is any reason for us not to include Turku's realtime now as well now in seutuplus?

sharhio:
There is no reason to not add Turku, in fact, it should be added.

PR #4870, opened 2023-10-03
fix: show more nearby stops that are not part of station

nurmAV:
## Proposed Changes

  - Improves the nearby stops view by listing more stops that are not part of any station
  - Related OTP PR: https://github.com/opentripplanner/OpenTripPlanner/pull/5390

## Pull Request Check List

## Review

  - Merge the pull request


app/component/StopNearYou.js
```diff
@@ -28,7 +28,8 @@ const StopNearYou = (
     }
   }, [currentTime, currentMode]);
   const description = desc || stop.desc;
-  const isStation = !!stop.parentStation || !!stopId;
+  const isStation =
+    stop.locationType === 'STATION' || !!stop.parentStation || !!stopId;
```
optionsome: I think we can probably just simply check `stop.locationType === 'STATION'` here and if the locationType is missing from query, we can add it to there.

app/component/StopsNearYouPage.js
```diff
@@ -213,7 +213,7 @@ class StopsNearYouPage extends React.Component {
 
   getQueryVariables = mode => {
     const { searchPosition } = this.state;
-    let placeTypes = 'STOP';
+    let placeTypes = ['STOP', 'STOP_OR_STATION'];
```
optionsome: I think we probably should only have the 'STOP_OR_STATION' filter here.

app/component/StopNearYouContainer.js
```diff
@@ -126,6 +127,11 @@ const containerComponent = createRefetchContainer(
         ...StopNearYouContainer_stop
         @arguments(startTime: $startTime, omitNonPickups: $omitNonPickups)
       }
+
+      station(id: $stopId) {
```
optionsome, line 85: I didn't test this yet but does the code somehow magically already handle that sometimes stations is now returned?

optionsome (changes requested):

app/component/StopNearYouContainer.js
nurmAV: Yes, it seems that the code already handles this. All the stops and stations seem to work like before without changes to code

optionsome (approved):

PR #4871, opened 2023-10-04
feat: realtime traveller capacity

tekoiv:
## Proposed Changes

  - Add realtime capacity information to stops near you -view
  - Add realtime capacity information to route suggestion -view
  - Note for the reviewer: I'd suggest using line 79 to test, it has the most consistent data even though it sometimes fails to deliver
  
## Pull Request Check List

## Review

  - Merge the pull request


app/component/CapacityModal.js
```diff
@@ -0,0 +1,137 @@
...
+        <FormattedMessage
+          id="capacity-modal.subheading"
+          defaultMessage="Real-time capacity information is available for some vehicles"
+        />
+      </p>
+      <span className="explanations-heading">
+        <FormattedMessage id="capacity-modal.legend" defaultMessage="Legend" />
+      </span>
+      <div className="capacity-info-row">
+        <div className="icon" style={{ color: '#007AC9' }}>
```
optionsome: It's probably better to give these divs style in sass instead of here. Probably it would be a good idea to create own sass file for this.

app/component/DepartureRow.js
```diff
@@ -16,14 +17,23 @@ import Icon from './Icon';
...
 
 const DepartureRow = (
-  { departure, departureTime, showPlatformCode, canceled, ...props },
+  {
+    departure,
+    departureTime,
+    showPlatformCode,
+    canceled,
+    onCapacityClick,
+    capacity,
```
optionsome: This prop seems to be unused.

app/component/RouteNumber.js
```diff
@@ -194,6 +210,9 @@ RouteNumber.propTypes = {
   withBicycle: PropTypes.bool,
   card: PropTypes.bool,
   appendClass: PropTypes.string,
+  hasOneTransitLeg: PropTypes.bool,
+  occupancyStatus: PropTypes.string,
+  leg: PropTypes.object,
```
optionsome: I think we should avoid passing the leg or having any props that refer to leg in this component as this is meant to be reusable for many things so I think we should be more generic.

app/component/RouteNumber.js
```diff
@@ -153,6 +154,21 @@ function RouteNumber(props, context) {
             </div>
           )}
       </span>
+      {context.config?.useRealtimeTravellerCapacities &&
+        props.hasOneTransitLeg &&
+        props.occupancyStatus &&
+        props.occupancyStatus !== 'NO_DATA_AVAILABLE' &&
+        isDepartureWithinTenMinutes(props.leg.startTime) && (
+          <span
+            style={{ color: 'white', marginLeft: 'auto', paddingRight: '3px' }}
```
optionsome: Styles could in a sass file again.

app/component/RouteNumberContainer.js
```diff
@@ -37,6 +43,9 @@ RouteNumberContainer.propTypes = {
   className: PropTypes.string,
   fadeLong: PropTypes.bool,
   withBicycle: PropTypes.bool,
+  hasOneTransitLeg: PropTypes.bool,
+  occupancyStatus: PropTypes.string,
+  leg: PropTypes.object,
```
optionsome: This is again a component that is used in many contexts so I think we should be generic with the props.

app/component/StopNearYou.js
```diff
@@ -98,6 +102,16 @@ const StopNearYou = (
           </>
         )}
       </div>
+      <Modal
+        appElement="#app"
+        contentLabel="Testi"
```
optionsome: This probably should be something else :D.

app/component/StopNearYou.js
```diff
@@ -98,6 +102,16 @@ const StopNearYou = (
...
+        closeButtonLabel="Sulje"
```
optionsome: I think we should use english as default but should this come through translations?

app/component/RouteNumber.js
optionsome: We could probably just pass the occupancyStatus here when it should be rendered and then render it if it is defined. We can move the logic that decides if the occupancy where the route number (and its container) are rendered.

app/util/occupancyUtil.js
```diff
@@ -0,0 +1,18 @@
...
+
+export function mapStatus(status) {
+  if (status === 'EMPTY') {
+    return 'MANY_SEATS_AVAILABLE';
+  }
+  if (status === 'NOT_ACCEPTING_PASSENGERS') {
+    return 'FULL';
+  }
+  return status;
+}
```
optionsome: I think we should have a switch case here that return handles all the possible values and a default which is probably NO_DATA_AVAILABLE in case at some point the OTP's enum gets more values so we should have some default handling.

app/util/occupancyUtil.js
```diff
@@ -0,0 +1,18 @@
...
+
+export function isDepartureWithinTenMinutes(legStartTime) {
+  return (
+    moment(legStartTime).diff(moment(), 'minutes') <= 10 &&
+    moment(legStartTime).diff(moment(), 'minutes') > -1
+  );
+}
```
optionsome: Maybe this is where the logic should belong that decides if the occupancy should be rendered or not. This could be renamed and expanded so that it contains most checks and then some component specific checks can exist in the component (at least the limit on the number of legs in the summary page). Also, the parameter should be probably just called `departureTime` or `stopDepartureTime`.

digitransit-search-util/packages/digitransit-search-util-query-utils/schema/schema.graphql
optionsome, line 1: Since we update the schema here (which I think we should to keep the different schema files consistent), I think we should bump the package versions.

app/configurations/config.hsl.js
```diff
@@ -537,6 +537,8 @@ export default {
 
   showSimilarRoutesOnRouteDropDown: true,
 
+  useRealtimeTravellerCapacities: true,
```
optionsome, line 540: Maybe we should add this parameter to the default config as false to improve the visibility of this parameter's existence. Also, should we already include the default icons or are the icons we were given somehow HSL specific? I wonder if Jyväskylä also would want this to be enabled as they had the occupancy information in their data if I don't remember wrong.

optionsome (changes requested):
I quickly tested this and noticed a couple of issues:
* Stop (and probably also station) page crashes due to some issue with departure row component
* According the design in epic, the occupancy icon should be clickable in the route suggestions as well but it doesn't seem to be currently

app/util/occupancyUtil.js
```diff
@@ -1,18 +1,40 @@
...
+export function isDepartureWithinTenMinutes(departureTime) {
   return (
-    moment(legStartTime).diff(moment(), 'minutes') <= 10 &&
-    moment(legStartTime).diff(moment(), 'minutes') > -1
+    moment(departureTime).diff(moment(), 'minutes') <= 10 &&
+    moment(departureTime).diff(moment(), 'minutes') > -1
   );
 }
+
+export function getCapacity(config, leg) {
```
optionsome: Maybe we should call this getCapacityForLeg (or maybe getOccupancyStatusForLeg) and then have a more generic function with parameters (config, occupancyStatus, departureTime) which can be used from DepartureRow as well. Also, I think we should add some documentation on utility methods in general.

optionsome (changes requested):
* Seems like there isn't enough padding on the right in occupancy icons in summary page
* In stops near you page the occupancy icon (sometimes?) goes too much to the right
* Occupancy popup has no close button but a x on the top right
* Seems like it's not possible to navigate to the occupancy in stops near you view with tab
* The occupancy legend in popup is slightly difficult to understand with screen reader, not sure how to better link the descriptions and details together
* Nothing about occupancy seems to be read for screen readers in neither stops near you view or itineraries

PR #4881, opened 2023-11-07
DT-5904 (DT-5325) CO2 emissions

sharhio:
## Proposed Changes

  - Show CO2 emissions 
  - Configured for HSL and Matka

## Pull Request Check List

## Review

  - Merge the pull request


app/component/ItineraryTab.js
```diff
@@ -354,6 +385,52 @@ class ItineraryTab extends React.Component {
...
+                              defaultMessage="CO₂ emissions for this route"
+                              values={{
+                                co2value,
+                                carCo2Value,
+                              }}
+                            />
+                          }
+                          {config.URL.EMISSIONSINFO && (
+                            <div className='co2link'>
+                              <Link style={{ textDecoration: 'none', fontWeight: '450' }}
```
optionsome: Link from found library should only be used for links inside this app. Just use `<a>` tag here. The link doesn't work currently, I'm not sure if it is just because of this issue.

app/component/ItineraryTab.js
```diff
@@ -80,13 +84,16 @@ class ItineraryTab extends React.Component {
...
+    lang: PropTypes.string.isRequired,
     hideTitle: PropTypes.bool,
+    carItinerary: PropTypes.array,
     currentLanguage: PropTypes.string,
     changeHash: PropTypes.func,
   };
 
   static defaultProps = {
     hideTitle: false,
+    carItinerary: [],
```
optionsome: The name is carItinerary but it's an array?

app/component/ItineraryTab.js
```diff
@@ -217,8 +223,13 @@ class ItineraryTab extends React.Component {
     const suggestionIndex = this.context.match.params.secondHash
       ? Number(this.context.match.params.secondHash) + 1
       : Number(this.context.match.params.hash) + 1;
+
+    const co2value = typeof itinerary.emissionsPerPerson?.co2 === 'number' && itinerary.emissionsPerPerson?.co2 >= 0 ? Math.round(itinerary.emissionsPerPerson?.co2) : -1;
```
optionsome: Maybe use null instead of -1.

app/component/ItineraryTab.js
```diff
@@ -217,8 +223,13 @@ class ItineraryTab extends React.Component {
...
+    const carCo2Value = this.props.carItinerary && this.props.carItinerary.length > 0 ? Math.round(this.props.carItinerary[0].emissionsPerPerson?.co2) : 0;
```
optionsome: Here we use 0 as a fallback, maybe null here as well.

app/component/ItineraryTab.js
```diff
@@ -217,8 +223,13 @@ class ItineraryTab extends React.Component {
...
+    const co2SimpleDesc =  carCo2Value === 0;
```
optionsome: this is not the description itself, maybe rename to be `useCo2SimpleDesc`, there also seems to be extra space after the `=`, weird that a linter isn't complaining about this(?).

app/component/ItineraryTab.js
```diff
@@ -354,6 +385,52 @@ class ItineraryTab extends React.Component {
...
+                <div className="itinerary-co2-comparison">
+                  <div className="itinerary-co2-line">
+                    <div className={cx('divider-top')} />
+                    <div className="co2-container">
+                      <div className="co2-description-container">
+                        <Icon img="icon-icon_co2_leaf" className="co2-leaf" />
+                        <span className={cx("itinerary-co2-description", { simple: co2SimpleDesc })}>
+                          {co2SimpleDesc ?
+                            <FormattedMessage id="itinerary-co2.description-simple"
+                              defaultMessage="CO₂ emissions for this route2"
```
optionsome:
```suggestion
                              defaultMessage="CO₂ emissions for this route"
```

app/component/ItineraryTab.js
```diff
@@ -354,6 +385,52 @@ class ItineraryTab extends React.Component {
...
+                              id="itinerary-co2.description"
+                              defaultMessage="CO₂ emissions for this route"
+                              values={{
+                                co2value,
+                                carCo2Value,
+                              }}
+                            />
+                          }
+                          {config.URL.EMISSIONSINFO && (
+                            <div className='co2link'>
```
optionsome: Maybe for consistency, the name of the class should be `co2-link`.

app/component/SummaryPage.js
```diff
@@ -2501,6 +2513,7 @@ class SummaryPage extends React.Component {
                   focusToPoint={this.focusToPoint}
                   focusToLeg={this.focusToLeg}
                   isMobile={false}
+                  carItinerary={carPlan?.itineraries}
```
optionsome: Maybe here we should just fetch the first itinerary (there will always be just one or none), so the carItinerary can be singular.

app/component/SummaryPage.js
```diff
@@ -2732,6 +2745,7 @@ class SummaryPage extends React.Component {
           serviceTimeRange={this.props.serviceTimeRange}
           focusToLeg={this.focusToLeg}
           onSwipe={this.changeHash}
+          carItinerary={carPlan?.itineraries}
```
optionsome: Same as above.

app/component/SummaryRow.js
```diff
@@ -259,7 +260,11 @@ const SummaryRow = (
   const startTime = moment(data.startTime);
   const endTime = moment(data.endTime);
   const duration = endTime.diff(startTime);
-
+  const co2value =
+    typeof data.emissionsPerPerson?.co2 === 'number' &&
+    data.emissionsPerPerson?.co2 >= 0
+      ? Math.round(data.emissionsPerPerson?.co2)
+      : -1;
```
optionsome: This code is repeated in a couple of places, maybe add an util for it somewhere, not sure if there is some fitting file somewhere already or do we need a new one.

app/component/itinerary.scss
```diff
@@ -191,6 +191,158 @@ $itinerary-tab-switch-height: 48px;
...
+        font-style: normal;
+        font-weight: 325;
+        font-size: 15px;
+        line-height: 18px;
+        /* identical to box height */
+        display: flex;
+        align-items: center;
+        letter-spacing: -0.03em;
+
+        /* Grayscale/HSL Black 75 */
```
optionsome: Seeing something HSL specific outside of configuration makes me wonder if something should belong to configuration. On the other hand, seems like we have used this color in a lot of places and not just for HSL. Maybe just remove the HSL reference from the comment or the whole comment :D.

app/component/itinerary.scss
```diff
@@ -191,6 +191,158 @@ $itinerary-tab-switch-height: 48px;
...
+      overflow: visible;
+
+      background: rgba(100, 190, 30, 0.15);
+      border-radius: 4px;
+
+      .itinerary-co2-value {
+        min-width: 29px;
+        height: 16px;
+
+        font-family: $font-family; //'Gotham Rounded';
```
optionsome: This is not always gotham rounded, depends on the configuration.

app/configurations/config.default.js
```diff
@@ -865,4 +865,7 @@ export default {
   },
 
   showAlternativeLegs: true,
+
+  // Notice! Turning on this setting forces the search for car routes.
```
optionsome: When I first read this comment, I thought this magically turned on the car itinerary suggestions, which this doesn't do. Make it more explicit that this causes a car search used only for the co2 comparison purposes.

app/configurations/config.hsl.js
```diff
@@ -502,6 +507,9 @@ export default {
   showBikeAndPublicItineraries: true,
   showBikeAndParkItineraries: true,
 
+  // DT-5325 Notice! Turning on this setting forces the search for car routes.
```
optionsome: Same as above.

app/configurations/config.jyvaskyla.js
```diff
@@ -163,4 +163,6 @@ export default configMerger(walttiConfig, {
     stops: true,
     itinerary: true,
   },
+  // DT-5325 Notice! Turning on this setting forces the search for car routes.
```
optionsome: Same as above.

app/configurations/config.matka.js
```diff
@@ -524,4 +524,6 @@ export default {
       virtualMonitorBaseUrl: 'https://matkamonitori.digitransit.fi/',
     },
   },
+  // Notice! Turning on this setting forces the search for car routes.
```
optionsome: Same as above.

optionsome: Also, since the kela configuration is now built on top of the matka configuration (which was a mistake we should fix at some point), set this to false in the kela configuration.

optionsome (changes requested):
On mobile, in the view with the itinerary suggestion open, the emissions on the top of the view should be moved slightly to the right. I didn't fully test yet because there were some questions about the spec and issues with testing as the otp data configuration was accidentally removed in a merge conflict.

app/component/ItineraryTab.js
sharhio: Link wasn't working because of the lang variable. Changed Link to <a> tag also.

app/component/ItineraryTab.js
sharhio: Changed the naming

app/component/ItineraryTab.js
sharhio: Changed to null.

app/component/itinerary.scss
sharhio: Comment was left by mistake, removed it.

app/configurations/config.default.js
sharhio: Yes, comment was meant to indicate the query is run, not that the car option is shown on ui. Added a clarification.

app/configurations/config.jyvaskyla.js
sharhio: Added a clarification.

app/configurations/config.matka.js
sharhio: Added a clarification for matka and set this option to false for kela.

app/configurations/config.hsl.js
sharhio: Added a clarification.

app/component/ItineraryTab.js
sharhio: Fixed the name. I don't know either why linter didn't catch the gap...

app/component/ItineraryTab.js
sharhio: Fixed the typo.

app/component/ItineraryTab.js
```diff
@@ -354,6 +390,50 @@ class ItineraryTab extends React.Component {
...
+                            </div>
+                          )}
+                        </span>
+                      </div>
+                    </div>
+                    <div className={cx('divider-bottom')} />
+                  </div>
+                </div>
+              ) :
+                (null)}
```
hbruch: `ItineraryTab` is already huge. I'd suggest to create a proper component for co2information.

app/component/SummaryPage.js
sharhio: Changed to send only the first one on.

app/component/SummaryPage.js
sharhio: Fixed

app/component/itinerary.scss
sharhio: This comment as left by mistake, deleted it.

app/component/ItineraryTab.js
optionsome: Yes that is a good point, I think we should do it

app/configurations/config.hsl.js
```diff
@@ -47,6 +47,11 @@ export default {
     BANNERS: BANNER_URL,
     HSL_FI_SUGGESTIONS: 'https://content.hsl.fi/api/v1/search/suggestions',
     EMBEDDED_SEARCH_GENERATION: '/reittiopas-elementti',
+    EMISSIONSINFO: {
```
optionsome: I think this should be `EMISSIONS_INFO`

sharhio: Fixed

app/component/ItineraryTab.js
sharhio: Done, also separated the smaller co2 info earlier in the code.

app/component/SummaryRow.js
sharhio: Done

app/component/ItineraryTab.js
sharhio: The className isn't actually used for anything so I removed it.

optionsome (approved):

PR #4886, opened 2023-11-16
DT-6092 - Move to new fare API

nurmAV:
## Proposed Changes

  - Fare information for an itinerary is now computed from fareProducts fields returned in each leg instead of an itinerary level fares object
  - Updates GraphQL schema
  - Updates utilities functions and tests related to fares computation

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen (changes requested):
Itinerary view crashes if some legs have unknows fares. Try helsinki-riihimäki.

app/component/MobileTicketPurchaseInformation.js
```diff
@@ -21,7 +21,7 @@ export default function MobileTicketPurchaseInformation(
     !fare.isUnknown,
     config.availableTickets,
   );
-  const price = `${(fare.cents / 100).toFixed(2)} €`.replace('.', ',');
+  const price = `${fare.price.toFixed(2)} €`.replace('.', ',');
```
vesameskanen, line 24: Do we need toFixed?

app/component/TicketInformation.js
```diff
@@ -81,7 +83,11 @@ export default function TicketInformation(
               <a href={config.ticketLink}>
                 <div className="ticket-identifier">
                   {config.useTicketIcons
-                    ? renderZoneTicket(fare.ticketName, alternativeFares)
+                    ? renderZoneTicket(
+                        // fare.product.id.split(':')[1],
```
vesameskanen: what is the purpose of this comment?

app/component/TicketInformation.js
```diff
@@ -95,12 +101,16 @@ export default function TicketInformation(
               <div className="fare-container">
                 <div className="ticket-identifier">
                   {config.useTicketIcons
-                    ? renderZoneTicket(fare.ticketName, alternativeFares)
+                    ? renderZoneTicket(
+                        // fare.product.id.split(':')[1],
```
vesameskanen: Another strange code comment

app/util/fareUtils.js
```diff
@@ -1,57 +1,49 @@
...
-    ...fare,
-    agency:
-      (Array.isArray(fare.routes) &&
-        fare.routes.length > 0 &&
-        fare.routes[0].agency) ||
-      undefined,
+  const knownFareLegs = uniqBy(
+    filteredLegs
+      .filter(l => l.fareProducts.length > 0)
+      .map(leg => ({
```
vesameskanen: isn't fareProduct filtering  already done above?

app/util/fareUtils.js
```diff
@@ -1,57 +1,49 @@
...
+        fareProducts: leg.fareProducts.filter(fp =>
+          availableTickets.includes(fp.product.id),
+        ),
+        route: leg.route,
+        agency: leg.route.agency,
+      })),
+    'fareProducts[0].id',
+  ).map(fp => ({
+    fareProducts: fp.fareProducts,
```
vesameskanen: why do you map 2 times? Is it possible to define final object in one round?

app/util/fareUtils.js
```diff
@@ -1,57 +1,49 @@
...
-      Array.isArray(routes) &&
-      routes) ||
-    []
-  )
-    .filter(route => !routesWithFares.includes(route.gtfsId))
+  // Legs that have empty fares but still have a route, i.e. transit legs
+  const unknownFareLegs = filteredLegs
+    .filter(l => l.fareProducts.length === 0 && l.route)
+    .map(leg => leg.route)
     .map(route => ({
```
vesameskanen: Again two mapping rounds for no obvious reason.

vesameskanen (commented):
Code could be improved a bit.

vesameskanen:
Kuopio planner shows always ticket price NaN

app/component/MobileTicketPurchaseInformation.js
nurmAV: It seems that without it, the UI will display e.g. 2.7 € instead of the 2.70 € that has been previously shown. OTP returns the price with only one digit.

app/util/fareUtils.js
nurmAV: The first filtering only removes unavailable ticket types (for some reason OTP seems to return a fareProduct with id HSL:regular as well as the actual ticket for a leg with available tickets). The two subsequent filterings then separately process the legs that have available tickets and those that do not.

app/util/fareUtils.js
nurmAV: These extra mapping rounds seem just to be leftovers from some experimentation during the development. Reduced them to one round.

app/component/TicketInformation.js
nurmAV: This is again probably the result of forgetting to clean up after doing some experimentation during development. Removed these comments as unnecessary.

PR #4891, opened 2023-11-21
chore: configure digitraffic vehicle realtime

tekoiv:
## Proposed Changes

  - Configured Digitraffic vehicle realtime
  - Fixed a bug where intercity journeys resulted in realtime topics with wrong feed ids

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/util/mqttClient.js
+++ b/app/util/mqttClient.js
@@ -135,14 +135,7 @@ export function changeTopics(settings, actionContext) {
 
 export function startMqttClient(settings, actionContext) {
   const options = settings.options || [{}];
-  let topics = options.map(option => getTopic(option, settings));
-  // If a general topic is returned, at least limit it to instances feed ids
-  if (topics[0] === '/gtfsrt/vp/+/+/+/+/+/+/+/+/+/+/+/+/+/+/+/#') {
-    const { feedIds } = actionContext.config;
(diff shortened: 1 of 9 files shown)


app/util/mqttClient.js
```diff
@@ -134,7 +135,14 @@ export function changeTopics(settings, actionContext) {
...
   const options = settings.options || [{}];
-  const topics = options.map(option => getTopic(option, settings));
+  let topics = options.map(option => getTopic(option, settings));
+  // If a general topic is returned, at least limit it to instances feed ids
+  if (topics[0] === '/gtfsrt/vp/+/+/+/+/+/+/+/+/+/+/+/+/+/+/+/#') {
+    const { feedIds } = actionContext.config;
+    topics = feedIds.map(
+      feedId => `/gtfsrt/vp/${feedId}/+/+/+/+/+/+/+/+/+/+/+/+/+/+/#`,
+    );
+  }
```
optionsome: This topic code is a bit too specific to this one MQTT broker. Couldn't it be possible to somehow define in the Map.js startClient that we should subscribe to these topics?

optionsome:
I don't know if these are somehow already disabled for kela but I think we should disable vehicle showing there if it already isn't.

PR #4896, opened 2023-11-27
DT-6060: Check config

Antiik91:
## Proposed Changes

  -

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/util/analyticsUtils.js
+++ b/app/util/analyticsUtils.js
@@ -58,5 +58,12 @@ const handleChange = () => {
  */
 export function initAnalyticsClientSide() {
   window.dataLayer = window.dataLayer || [];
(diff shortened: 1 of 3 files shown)


app/util/analyticsUtils.js
```diff
@@ -58,5 +58,12 @@ const handleChange = () => {
  */
 export function initAnalyticsClientSide() {
   window.dataLayer = window.dataLayer || [];
-  window.addEventListener('CookieInformationConsentGiven', handleChange, false);
+  const config = window.state?.context?.plugins['extra-context-plugin'].config;
```
optionsome: You could just add config as a param to this function.

Antiik91: Done.

PR #4900, opened 2023-12-01
DT-5926: Datetimepicker for generated components

Dvun:
## Proposed Changes

  - Datetimepicker for generated components

## Pull Request Check List

## Review

  - Merge the pull request


app/component/EmbeddedSearch/EmbeddedSearch.js
```diff
@@ -298,15 +314,82 @@ const EmbeddedSearch = (props, context) => {
...
+    setState({ ...state, open: true });
+  };
+
   return (
     <div
       className={`embedded-seach-container ${
         bikeOnly ? 'bike' : walkOnly ? 'walk' : ''
       }`}
       id={appElement}
+      style={{ height: isTimepickerSelected ? '380px' : '250px' }}
```
optionsome: This could be moved sass file by providing a class name here if the timepicker is selected.

app/component/EmbeddedSearch/EmbeddedSearch.js
```diff
@@ -298,15 +314,82 @@ const EmbeddedSearch = (props, context) => {
...
     >
       <div className="background-container">{drawBackgroundIcon()}</div>
-      <div className="control-panel-container">
+      <div className="control-panel-container" style={{ position: 'relative' }}>
```
optionsome: This style could be potentially moved to sass as well.

app/component/EmbeddedSearch/EmbeddedSearch.js
```diff
@@ -321,7 +404,37 @@ const EmbeddedSearch = (props, context) => {
...
+                onClose={onClose}
+                openPicker={state.open}
+                isHideCloseButton={state.isHideCloseButton}
+              />
+            </div>
+          )}
+
+          <div
+            className="embedded-search-button-container"
+            style={{ margin: '10px 0 0 0' }}
```
optionsome: This style can be moved to sass.

app/component/EmbeddedSearch/EmbeddedSearch.js
```diff
@@ -321,7 +404,37 @@ const EmbeddedSearch = (props, context) => {
...
+                onDateChange={onDateChange}
+                onNowClick={onNowClick}
+                onDepartureClick={onDepartureClick}
+                onArrivalClick={onArrivalClick}
+                embedWhenClosed={null}
+                embedWhenOpen={null}
+                lang={lang}
+                color={colors.primary}
+                timeZone={config.timezoneData.split('|')[0]}
+                serviceTimeRange={context.config.itinerary.serviceTimeRange}
```
optionsome, line 410:
```suggestion
                serviceTimeRange={config.itinerary.serviceTimeRange}
```

app/component/EmbeddedSearchGenerator.js
```diff
@@ -411,6 +431,34 @@ const EmbeddedSearchGenerator = (props, context) => {
             )}
           </fieldset>
 
+          <fieldset id="timePicker">
```
optionsome: I think we should use `time-picker` style naming instead of `timePicker` for ids and class names.

app/component/EmbeddedSearch/EmbeddedSearch.js
```diff
@@ -321,7 +404,37 @@ const EmbeddedSearch = (props, context) => {
...
+                fontWeights={config.fontWeights}
+                onOpen={onOpen}
+                onClose={onClose}
+                openPicker={state.open}
+                isHideCloseButton={state.isHideCloseButton}
```
optionsome: I think better name for this parameter could be `isAlwaysOpen`, this is used in multiple files so rename it everywhere. The design images showed that the grey background should be removed when the timepicker is rendered from the embedded search. You should use this same prop to remove the grey background if this is true.

app/component/EmbeddedSearch/EmbeddedSearch.js
```diff
@@ -298,15 +314,82 @@ const EmbeddedSearch = (props, context) => {
...
+    });
+  };
+
+  const onClose = () => {
+    setState({ ...state, open: false });
+  };
+
+  const onOpen = () => {
+    setState({ ...state, open: true });
+  };
```
optionsome: I think these can be removed if we force the timepicker to be always open in this view and `open` can be removed from state.

app/component/EmbeddedSearch/EmbeddedSearch.js
```diff
@@ -321,7 +404,37 @@ const EmbeddedSearch = (props, context) => {
...
+                embedWhenClosed={null}
+                embedWhenOpen={null}
+                lang={lang}
+                color={colors.primary}
+                timeZone={config.timezoneData.split('|')[0]}
+                serviceTimeRange={context.config.itinerary.serviceTimeRange}
+                fontWeights={config.fontWeights}
+                onOpen={onOpen}
+                onClose={onClose}
+                openPicker={state.open}
```
optionsome: This can be hard-coded as true.

app/component/EmbeddedSearch/EmbeddedSearch.js
```diff
@@ -87,6 +88,13 @@ const EmbeddedSearch = (props, context) => {
     });
   });
 
+  const [state, setState] = useState({
+    open: true,
+    isHideCloseButton: true,
+    time: undefined,
+    arriveBy: false,
+    keepPickerOpen: false,
```
optionsome: I think this `keepPickerOpen` can be removed as well from everywhere here.

app/component/embeddedSearchGenerator.scss
```diff
@@ -82,6 +82,16 @@
...
                 }
             }
+
+            .datetimepicker-container {
+                margin-top: 1rem;
+
+                & fieldset {
+                    margin: 0 !important;
+                    padding: 0 !important;
+                    max-width: 399px !important;
```
optionsome: We should try to avoid using these !important style definitions. I think the issue here is probably that the styles from here leak to the timepicker. We should either make those generic style definitions here that affect all html elements be more specific and only affect certain classes or make some wrapper divs for the other elements in the generator and make the generic styles under those divs so they don't affect the preview.

optionsome (changes requested):

PR #4921, opened 2024-01-09
DT-6175 part1: code cleanup

vesameskanen:
- Rename components to meaningful names:  SummaryPage -> ItineraryPage etc.
- First split of huge ItineraryPage into its queries and static util functions
- Standardise unusual fragmented ItinerarySummaryListContainer sub component  folder
- Use optional chaining where possible
- Rename some functions in ItineraryPage

At this point, functionality does not change and everything works as before.


app/component/AltTravelBar.js
```diff
@@ -7,7 +7,7 @@ import { matchShape, routerShape } from 'found';
 import RightOffcanvasToggle from './RightOffcanvasToggle';
 import DatetimepickerContainer from './DatetimepickerContainer';
 
-class QuickSettingsPanel extends React.Component {
+class AltTravelBar extends React.Component {
   // TODO decide if these are needed for new datepicker
```
optionsome: Have we decided :D?

app/component/AltTravelBar.js
```diff
@@ -59,4 +59,4 @@ class QuickSettingsPanel extends React.Component {
   }
 }
 
-export default QuickSettingsPanel;
+export default AltTravelBar;
```
optionsome: Does this actually contain anything else than the Datetime picker? If not, I would prefer to call this something related to the picker because this new name doesn't really tell me what this is any better than the old name. This file also uses a "quicksettings-container" class, maybe rename that too.

app/component/Itinerary.js
```diff
@@ -237,7 +237,7 @@ const hasOneTransitLeg = data => {
   return data.legs.filter(leg => leg.transitLeg).length === 1;
 };
 
-const SummaryRow = (
+const Itinerary = (
```
optionsome, line 240: Maybe call this ItinerarySummary to differentiate that this isn't the details page? Again, there are some classes etc that use the old naming.

app/component/ItineraryDetailView.js
```diff
@@ -72,7 +72,7 @@ const ItineraryShape = PropTypes.shape({
 });
 
 /* eslint-disable prettier/prettier */
-class ItineraryTab extends React.Component {
+class ItineraryDetailView extends React.Component {
```
optionsome: What is the difference between view and page? Also, not sure if it should be Details or Detail. The old name sort of implied this is a swipeable tab (maybe by accident), not sure if that is relevant.

app/component/ItineraryPage.js
```diff
@@ -607,361 +378,33 @@ class SummaryPage extends React.Component {
 
   paramsOrQueryHaveChanged = () => {
     return (
-      !isEqual(this.params, this.context.match.params) ||
-      !isEqual(this.query, this.context.match.location.query)
+      !isEqual(this.params, this.props.match.params) ||
```
optionsome, line 381: match is usually loaded through context, I don't remember anymore why we sometimes load it from props.

app/component/AltTravelBar.js
vesameskanen: Dead props and context parts removed

app/component/AltTravelBar.js
vesameskanen: True, this component actually rendered time picker and settings button. Renamed as SearchSettings instead of QuickSettings. Also renamed the styles,.

app/component/Itinerary.js
vesameskanen: There is already another ItinerarySummary component, where name makes sense. It is shown on top of itinerary details.

app/component/ItineraryDetailView.js
vesameskanen: Component does not know (and should not know) that it is rendered into a 'tab'.  So Tab should be removed.

I renamed the component as ItineraryDetails.

app/component/ItineraryPage.js
vesameskanen: I think there was more props.match accees than context.match so I took the more popular. Match is available in props because router automatically puts it there. I verified that match.params are the same in both cases.

vesameskanen:
Mqtt works again - it broke because switch was called showOnSummaryPage and I replaced all SummaryPage occurences in ItineraryPage.

app/component/Itinerary.js
vesameskanen: And the actual legs are rendered by a component called ItineraryLegs which is pretty OK.

PR #4923, opened 2024-01-09
DT-6199: Time picker dismiss user's click when user has typed something in the picker.

Antiik91:
## Proposed Changes

  - New logic for time picker: when user completes the time by typing, time picker now shows the typed time as an option.
image of time picker showing 12:32 as an option
  - This fixes issues when the onChange function can't decide which value to use: `displayValue` that user has typed or the `option`  that is provided for it.

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen (commented):
Some issues:
- After selecting 'now', the current time does not appear in the dropdown. When user clicks time field, time dropdown opens showing time 00:00 selected and it is also set with enter key.  Dropdown is not scrolled to the current value.
- After typing the first minute digit, it would be great if the dropdown could show those ten minute values. Otherwise user's feedback may be ignored at enter key (even hour xx:00 gets selected)

digitransit-component/packages/digitransit-component-datetimepicker/src/helpers/Datetimepicker.js
```diff
@@ -89,7 +89,14 @@ function Datetimepicker({
...
   const inputRef = useRef();
   const alertRef = useRef();
+  const [invalidInput, setinvalidInput] = useState(false);
 
+  useEffect(() => {
+    document.body.style.setProperty(
+      '--input-color',
+      invalidInput ? '#ff0000' : ' #000000',
+    );
+  }, [invalidInput]);
```
vesameskanen: This styling looks very unusual. Normally we set styling by combining suitable classes in render method. Is there a reason for this solution?  

Also, it is strange that invalid property is sent from actual timepicker to this parent component with a callback, and the given value is passed back to the sender.  Why not to detect invalidity in timepicker and render an additional style which gives red input color there?

digitransit-component/packages/digitransit-component-datetimepicker/src/helpers/Datetimepicker.js
```diff
@@ -240,6 +251,74 @@ function Datetimepicker({
     }
   }
 
+  const validateInputTime = inputValue => {
+    const validateClock = (hours, minutes) => {
```
vesameskanen: Input validation is now confusing, for many reasons:
1. There are two validation functions. Code comments do not even try to explain their differences . Names should be clearly different and indicate purposes.
2. Validation function names change in crazy way when passed to child components. In desktop rendering, 'validateTime' is passed as 'validate' and 'validateInputTime' is passed as 'validateTime'. Confusion maximised :)
3. Related to above, there is no reason to pass static code as a prop. Move the validation functions to util and import them from there.

vesameskanen (commented):
Functionality works fine, but code quality needs improvements.

PR #4931, opened 2024-01-16
DT-6175 part 2 itinerary page and its sub component refactoring

vesameskanen:
- Fix all prop type warnings related to wrong prop type definitions, missing props etc.
- Add valid key to all components in lists (fixes debugger warnings)
- Validate prop values before passing them
- Code simplification and cleanup 

After these fixes, debugger gives only relay related warnings from the itinerary page.


app/component/ItineraryDetails.js
```diff
@@ -256,88 +261,75 @@ class ItineraryDetails extends React.Component {
...
+              isMobile={isMobile}
+              hideBottomDivider={isMobile && shouldShowFarePurchaseInfo(
+                config,
+                breakpoint,
+                fares,
+              )}
+            />,
             showRentalBikeDurationWarning && (
               <VehicleRentalDurationInfo
+		key="rentaldurationinfo"
```
optionsome, line 308: I wonder what case we should actually use for keys, seems like we use like 5 different variations.

app/component/ItineraryPage.js
```diff
@@ -889,35 +798,56 @@ class ItineraryPage extends React.Component {
...
+    } else if (!isEmpty(state.itineraryTopics)) {
+      this.stopClientAndUpdateTopics();
     }
-  }
+    if (
+      hash === 'parkAndRide' &&
+      !state.isFetchingWalkAndBike &&
+      !Array.isArray(state.parkRidePlan?.itineraries)
+    ) {
+      this.toggleStreetMode(''); // go back to showing normal itineraries
```
optionsome: I know this isn't introduced here but empty string always feels a bit wrong.

app/component/ItineraryPage.js
```diff
@@ -1012,19 +923,24 @@ class ItineraryPage extends React.Component {
...
+                  !Number.isNaN(windSpeed) &&
+                  !Number.isNaN(iconIndex)
+                ) {
+                  weatherData = {
+                    weatherHash,
+                    time,
+                    temperature,
+                    windSpeed,
+                    // Icon spec: www.ilmatieteenlaitos.fi/latauspalvelun-pikaohje -> Sääsymbolien selitykset ennusteissa
+                    iconId: checkDayNight(iconIndex, timem, from.lat, from.lon),
```
optionsome: timem?

optionsome (changes requested):
Timepicker in front page is broken

app/component/ItineraryDetails.js
vesameskanen: Converted all to lower case

app/component/ItineraryPage.js
vesameskanen: I agree. Cleanup done.

optionsome (approved):

PR #4936, opened 2024-01-24
DT-6224  render stops only from configured feeds

vesameskanen:
Add filtering by config feeds


optionsome:
I think we should discuss this change a bit more. I have heard some feedback previously that some cities have liked that the trains stations show up on the map.

PR #4949, opened 2024-02-08
digitransit-component cleanup

vesameskanen:
- object and array prop type usage fixed in reasonable cases
- missing default prop warnings fixed 
- missing react component key errors fixed for component lists


app/component/VehicleRentalDurationInfo.js
```diff
@@ -99,7 +99,9 @@ function VehicleRentalDurationInfo(props) {
 }
 
 VehicleRentalDurationInfo.propTypes = {
-  networks: PropTypes.array.isRequired,
+  networks: PropTypes.arrayOf(
+    PropTypes.oneOfType([PropTypes.string, PropTypes.array]),
```
sharhio, line 103: Should "array" be "arrayOf" here as well?

vesameskanen: Yes. Now - all - array props are converted into arrayOf.

PR #4961, opened 2024-02-23
DT-6242: Citybike configurations from the citybike configuration

Antiik91:
## Proposed Changes

  - UI:s that utilize citybikes will fetch season configuration from external source instead of old way that the configuration resided in config files
  - In order to use this new functionality, new env variables need to be set: `CITYBIKE_DB_CONN_STRING `and `CITYBIKE_DATABASE`
  - Configurations that can be changed from the external source are
    - enabled: should we show the citybikes in our UI at all
    - season. An object containing dates for `preSeason`, `start` and `end`.
    
    
   - Please note that `showCitybieNetwork` function in `modeUtils.js` contains `isDevelopmentEnvironment(config)` check. 
      -  Citybikes will show up in your UI regardles of the season settings if this is not modified, or the data is verified in somewhere else

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -163,40 +163,21 @@ export function getNamedConfiguration(configName) {
     }
   }
   if (conf.cityBike && allCitybikes?.length) {
-    const network = allCitybikes.find(
-      cb => conf.cityBike[cb.networkName] !== null,
-    );
-    let confCitybike;
     if (configName === 'matka') {
       allCitybikes.forEach(cb => {
-        let obj;
-        if (cb.smoove) {
-          obj = cb.smoove;
-        } else if (cb.vantaa) {
-          obj = cb.vantaa;
-        } else {
-          obj = cb;
-        }
-        const net = conf.cityBike.networks[obj.networkName];
-        if (net) {
-          net.enabled = obj.enabled;
-          net.season = obj.season;
+        const confCityBike = conf.cityBike.networks[cb.networkName];
+        if (confCityBike) {
+          confCityBike.enabled = cb.enabled;
+          confCityBike.season = cb.season;
         }
       });
-    } else if (configName === 'hsl') {
-      // eslint-disable-next-line no-restricted-syntax
-      for (const [key, value] of Object.entries(network)) {
-        confCitybike = conf.cityBike.networks[key];
-        confCitybike.enabled = value.enabled;
-        confCitybike.season = value.season;
-      }
     } else {
-      confCitybike = conf.cityBike.networks[network.networkName];
-
-      if (confCitybike) {
-        confCitybike.enabled = network.enabled;
-        confCitybike.season = network.season;
-      }
+      const networks = allCitybikes.filter(cb => configName === cb.config);
+      networks.forEach(network => {
+        const confCityBike = conf.cityBike.networks[network.networkName];
+        confCityBike.enabled = network.enabled;
+        confCityBike.season = network.season;
+      });
     }
   }
   if (!process.env.OIDC_CLIENT_ID && conf.allowLogin) {
--- a/app/config.js
+++ b/app/config.js
@@ -179,7 +179,7 @@ export function getNamedConfiguration(configName) {
       });
     } else {
       const seasonDefinitions = citybikeSeasonDefinitions.filter(
-        seasonDef => configName === seasonDef.config,
+        seasonDef => configName === seasonDef.configName,
       );
       seasonDefinitions.forEach(seasonDef => {
         const confCitybike = conf.cityBike.networks[seasonDef.networkName];
--- a/app/config.js
+++ b/app/config.js
@@ -19,6 +19,11 @@ if (defaultConfig.themeMap) {
   });
 }
 
+let allCitybikes;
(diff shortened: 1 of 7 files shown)


app/config.js
```diff
@@ -157,7 +162,24 @@ export function getNamedConfiguration(configName) {
       }
     }
   }
-
+  if (conf.cityBike && allCitybikes?.length) {
```
vesameskanen: This code is run every time new request arrives to UI server. Maybe we could store a flag 'seasonSet' to config so that   seasons get processed only once per theme.

app/config.js
```diff
@@ -157,7 +162,24 @@ export function getNamedConfiguration(configName) {
...
+    if (configName === 'matka') {
+      allCitybikes.forEach(cb => {
```
vesameskanen: Remove 'matka' from code. Instead, set a new property 'useAllSeasons=true' in matka's citybike config section.

app/config.js
```diff
@@ -157,7 +162,24 @@ export function getNamedConfiguration(configName) {
...
+        const confCitybike = conf.cityBike.networks[cb.networkName];
+        if (confCitybike) {
+          confCitybike.enabled = cb.enabled;
+          confCitybike.season = cb.season;
+        }
+      });
+    } else {
+      const networks = allCitybikes.filter(cb => configName === cb.config);
+      networks.forEach(network => {
+        const confCitybike = conf.cityBike.networks[network.networkName];
```
vesameskanen: I dislike 'cb 'name - it is a standard way to name callbacks. Maybe 'seasonDef' ? Also 'const networks' is confusing because filtered array does not contain networks. Call it e.g. seasonDefinitions.

Same naming should be used in server/server.js.

server/server.js
```diff
@@ -370,7 +371,107 @@ function startServer() {
...
+
+  return new Promise(mainResolve => {
+    const promises = [];
+
+    fetchCitybikeSeasons().then(r => {
+      const schedules = [];
+      r.forEach(cb => schedules.push(...cb.schedules));
+      configFiles.forEach(file => {
+        if (file.startsWith('config')) {
+          // eslint-disable-next-line import/no-dynamic-require
```
vesameskanen: file.startsWith('config') is repeated here. Also zone handling needs it. Add filtering to the place where file list is created.

server/server.js
```diff
@@ -370,7 +371,107 @@ function startServer() {
...
+          const conf = require(`${configsDir}/${file}`);
+          const configName = conf.default.CONFIG;
+          const { cityBike } = conf.default;
+          if (cityBike && Object.keys(cityBike).length > 0) {
+            promises.push(
+              new Promise(resolve => {
+                resolve(handleCitybikes(schedules, configName));
```
vesameskanen: Promise is currently not needed to execute synchronous code. Anyway, as we later may start fetching seasons one by one, this solution will become useful then.  Let it be!

server/server.js
```diff
@@ -370,7 +371,107 @@ function startServer() {
...
+  };
+}
+
+function handleCitybikes(schedules, configName) {
+  const bikes = schedules.filter(
+    cb => cb.region.toLowerCase() === configName || cb.config === configName,
+  );
+  const cbConfigurations = [];
+  bikes.forEach(bike => {
+    // This check is due to naming scheme in HSL citybike networks (Smoove and Vantaa)
```
vesameskanen: better name please. These are not bikes.

server/server.js
```diff
@@ -370,7 +371,107 @@ function startServer() {
...
+    if (bike.config && bike.config === 'hsl') {
```
vesameskanen: can bike.config be falsy?

server/server.js
```diff
@@ -370,7 +371,107 @@ function startServer() {
...
+      if (bike.region === 'Vantaa') {
+        cbConfigurations.push(
```
vesameskanen: Is Vantaa the only use case for region? Do we need region at all? The hard coded config names hsl and vantaa are really bad, so could this be like:

const networkName = seasonDef.networkName || operator.concat('_', configName);

This oneliner seems to replace 10 + lines of if else blocks.

vesameskanen (changes requested):

server/server.js
Antiik91: Region is defined in every other citybike cities name: For example there is multiple cities that uses same operator (For example donkey). And in waltti configurations they are named operator_city For example donkey_kotka. Only vantaa configuration is called vantaa, and hsl is called smoove, wihtout any regions. Hence the exceptions.

But You are right about the naming. Will fix them.

app/config.js
Antiik91: Done.

app/config.js
Antiik91: Done.

server/server.js
Antiik91: Yes, currently the config field is extra information for convenience in HSL citybikes. Real question is should it be falsy.

Let's hit with a call tomorrow and discuss this in more detail. I would like some insight on this one.

PR #4970, opened 2024-03-12
DT-5452 Show closed stops on map using badge icons

sharhio:
## Proposed Changes

  - Draw badges on stop icons 
  - Stop is permanently closed when there are no NO_SERVICE alerts and there are no scheduled services in the future. 
  - Stop is temporarily closed when there are no scheduled or real-time services on current day.
  - Test added for realtime stop layer

- Relies on otp PR https://github.com/opentripplanner/OpenTripPlanner/pull/5743 and https://github.com/HSLdevcom/OpenTripPlanner-data-container/pull/389 and https://github.com/opentripplanner/OpenTripPlanner/pull/6715#event-18425663839

## Pull Request Check List

## Review

  - Merge the pull request


app/util/mapIconUtils.js
```diff
@@ -350,6 +363,38 @@ function getSmallStopIcon(type, radius, color) {
...
+  x,
+  y,
+  iconWidth,
+  stopOutOfService,
+  noRoutesForStop,
+) {
+  const badgeSize = iconWidth * 0.75; // badge size is 75% of the icon size
+  const badgeImageId = stopOutOfService
+    ? `icon-icon_stop-temporarily-closed-badge`
+    : `icon-icon_stop-closed-badge`;
```
hbruch: Which mode does this stop icon represent (the sprite files does not show up in git)?
Another option to handle badges I'm currently exploring with rental stations, is having badges included in the original svg, and activate/color them via css vars. This would decouple the code from knowing where to place the badge.

sharhio: These are universal badges for any mode. Grey clock symbol / cross on red background.

app/component/map/IndexPageMap.js
```diff
@@ -53,11 +53,23 @@ function IndexPageMap(
...
 
   if (origin.lat) {
-    leafletObjs.push(<LocationMarker position={origin} type="from" />);
+    leafletObjs.push(
+      <LocationMarker
+        key={`${origin.lat},${origin.lon}`}
+        position={origin}
+        type="from"
+      />,
+    );
```
VillePihlava, line 62: Are these two `LocationMarker` changes just general improvements to make the key more unique?

app/component/map/tile-layer/Stops.js
```diff
@@ -123,8 +159,13 @@ class Stops {
   }
 
   getPromise(lang) {
+    const zoom = this.tile.coords.z + (this.tile.props.zoomOffset || 0);
```
VillePihlava: There exists a variable called `zoom` in this file that refers to `this.tile.coords.z` alone. Therefore, I think this should be called something like `zoomWithOffset.`

VillePihlava: Also you should replace all mentions of `this.tile.coords.z + (this.tile.props.zoomOffset || 0)` with this variable, I can find two.

app/component/map/tile-layer/Stops.js
```diff
@@ -93,21 +110,40 @@ class Stops {
...
-          ? feature.properties.platform
-          : false,
-        isHilighted,
-        !!(
-          feature.properties.type === 'FERRY' &&
-          !isNull(feature.properties.code)
-        ),
-        this.config.colors.iconColors,
-      );
+      if (isHilighted && zoom <= minZoom) {
```
VillePihlava: For some reason, we use both `hilighted` and `highlighted` in this repository (`highlighted` is correct). I think it would be nice to change all of these to `highlighted`, but maybe this is a separate PR

app/component/map/tile-layer/Stops.js
```diff
@@ -305,6 +346,46 @@ class Stops {
       );
     });
   }
+
+  drawHighlighted = (
```
VillePihlava, line 354: Why is the `stopAlertsQuery` only run for the highlighted stop?

digitransit-search-util/packages/digitransit-search-util-query-utils/schema/schema.graphql
```diff
@@ -2040,6 +2040,7 @@ type Stop implements Node & PlaceInterface {
   gtfsId: String!
   "Global object ID provided by Relay. This value can be used to refetch this object using **node** query."
   id: ID!
+
```
VillePihlava: Unnecessary newline

VillePihlava (changes requested):

app/component/map/IndexPageMap.js
sharhio: This removes a warning about missing keys (which do need to be unique)

app/component/map/tile-layer/Stops.js
sharhio: Because most of the time the alert info comes from the vectorTile but the realtimeLayer, which contains the alerts, is used only when the zoom level is close enough. The highlighted stops are shown also when the map is zoomed out so the alerts have to be fetched separately when the stop is highlighted.

app/component/map/tile-layer/Stops.js
VillePihlava: I created a PR for this

VillePihlava: https://github.com/HSLdevcom/digitransit-ui/pull/5436

VillePihlava (approved):
LGTM!

PR #4979, opened 2024-03-27
DT-5464 part 2: use paging 

vesameskanen:
- Iterate queries until enough itineraries has been found
- Use paging for earlier and later searches


app/component/ItineraryList.js
```diff
@@ -77,8 +75,8 @@ function ItineraryList(
     );
   }
   if (hash === streetHash.bikeAndVehicle) {
-    // bikeAndParkItineraryCount tells how many first itineraries use bike parking
-    if (bikeAndParkItineraryCount > 0) {
+    // bikeParkItineraryCount tells how many first itineraries use bike parking
```
optionsome: first itineraries?

app/component/ItineraryListContainer.js
```diff
@@ -235,21 +195,10 @@ class ItineraryListContainer extends React.Component {
...
-          searchTime={searchTime}
-          bikeAndParkItineraryCount={bikeAndParkItineraryCount}
-          showRelaxedPlanNotifier={showRelaxedPlanNotifier}
-          separatorPosition={separatorPosition}
-          loadingMore={loadingMore}
-          routingFeedbackPosition={routingFeedbackPosition}
-        >
-          {this.props.children}
-        </ItineraryList>
+          {...this.props}
```
optionsome: Are all props necessary? Passing all props to a child component always feels a bit wrong but since this is just a container, it might be that these share the same props.

app/component/ItineraryPage.js
```diff
@@ -121,14 +124,20 @@ export default function ItineraryPage(props, context) {
...
-    return {
-      settingsChanged: settingsState.settingsChanged,
-      from: props.match.params.from,
-      to: props.match.params.to,
-      time: props.match.location.query.time,
-      arriveBy: props.match.location.query.arriveBy,
-    };
+  function altLoading() {
+    return (
+      Object.keys(altStates).find(
```
optionsome: I think we can use Object.values and `some` instead of `find`.

app/component/ItineraryPage.js
```diff
@@ -121,14 +124,20 @@ export default function ItineraryPage(props, context) {
...
+        key => altStates[key][0].loading === ALT_STATE.LOADING,
+      ) !== undefined
+    );
+  }
+
+  function altLoadingDone() {
+    return (
+      !Object.keys(altStates).find(
```
optionsome: Same as above.

app/component/ItineraryPage.js
```diff
@@ -274,285 +272,257 @@ export default function ItineraryPage(props, context) {
...
+      if (!plan) {
+        plan = result.plan;
+      } else if (arriveBy) {
+        plan = {
+          ...plan,
+          pageInfo: {
+            ...plan.pageInfo,
+            startCursor: result.plan.pageInfo.startCursor,
+          },
+          edges: plan.edges.concat(result.plan.edges),
```
optionsome, line 267: I think there might still be an edge case where in some instances, same itinerary is returned twice while paging. I could ask in some OTP dev meeting if that is still the case. We might have to filter out duplicates somewhere on some criteria (I'm not sure if you are already doing it).

app/component/PlanConnection.js
```diff
@@ -0,0 +1,114 @@
...
+              gtfsId
+            }
+            trip {
+              gtfsId
+              directionId
+              stoptimesForDate {
+                scheduledDeparture
+              }
+              pattern {
+                ...RouteLine_pattern
```
optionsome: Do we actually render RouteLine in the itinerary view? Fragments from components which we don't render shouldn't be included.

app/configurations/config.hsl.js
```diff
@@ -499,7 +498,7 @@ export default {
   showBikeAndParkItineraries: true,
 
   // Notice! Turning on this setting forces the search for car routes (for the CO2 comparison only).
-  showCO2InItinerarySummary: true,
+  showCO2InItinerarySummary: false,
```
optionsome: Why is this changed?

app/configurations/config.default.js
```diff
@@ -214,9 +216,6 @@ export default {
   // if you enable car suggestions but the linear distance between all points is less than this, then a car route will
   // not be computed
   suggestCarMinDistance: 2000,
-  minTransferTime: 90,
-  optimize: 'GREENWAYS',
-  transferPenalty: 0,
```
optionsome, line 219: Was this unused?

optionsome (commented):
I didn't test this yet as https://github.com/opentripplanner/OpenTripPlanner/pull/5185/commits/50ff507fb433dfb79566f58d3effdb7831ade39c breaks some functionality. Instead of passing empty arrays, null should be used instead.

app/component/ItineraryListContainer.js
vesameskanen: Class converted to function component which uses `...rest` notification to pass props through.

app/component/ItineraryPage.js
vesameskanen: done

app/component/ItineraryPage.js
vesameskanen: fixed

app/configurations/config.hsl.js
vesameskanen: Some strange mistake, restored now.

app/configurations/config.default.js
vesameskanen: No, but those 3 settings are now moved to defaultSettings block so that they do not require special handling but work  the same way as all other routing settings.

optionsome: I had just missed that the transferPenalty was already defined in the defaultSettings, that's why it didn't need to be added there.

app/component/PlanConnection.js
vesameskanen: Good catch, that was some old stuff which I did not notice. Removed.

vesameskanen:
arriveBy earlier button still does not work, otherwise pretty OK.

PR #4993, opened 2024-04-19
DT-6297 use latest graphql queries everywhere

vesameskanen:
This PR removes all references to deprecated graphQL fields. Especially:
- Unified vehicleParking is now used instead of separate carPark and bikePark. Lots of duplicate code removed.
- New time types for itineraries adopted

Changes also include component refactoring and misc bug fixes. Lots of repeating code removed.

Itinerary page related components no longer use moment library and the component tree is now collected into a single 'itinerary' folder.

After this PR, linter gives one warning : UI is already using the new PlanConnection.


app/util/legUtils.js
```diff
@@ -3,6 +3,14 @@ import get from 'lodash/get';
 import { BIKEAVL_UNKNOWN } from './vehicleRentalUtils';
 import { getRouteMode } from './modeUtils';
 
+/**
+ * Get time as  milliseconds since the Unix Epoch
+ */
+export function legTime(lt) {
+  const t = lt.estimated?.time || lt.scheduledTime;
+  return Date.parse(t);
+}
```
optionsome, line 12: Quite often we use this for moment. I wonder if it makes any difference to create a method that directly parses the ISO 8601 time to moment.

app/component/AirportCollectLuggageLeg.js
```diff
@@ -18,7 +18,7 @@ function AirportCollectLuggageLeg(props, { config }) {
     <div className="row itinerary-row">
       <div className="small-2 columns itinerary-time-column">
         <div className="itinerary-time-column-time">
-          {moment(props.leg.endTime).format('HH:mm')}
+          {moment(props.leg.end).format('HH:mm')}
```
optionsome: I think you need to use the legTime method here to parse the actual time.

app/util/legUtils.js
vesameskanen: I added new util which returns an actual time string in the  form UI frequently uses.

app/component/AirportCollectLuggageLeg.js
vesameskanen: Right - I grepped 'moment(props.leg', found many of those and fixed them all.

app/component/ParkOrStationHeader.js
```diff
@@ -43,8 +46,8 @@ const ParkOrBikeStationHeader = ({ parkOrStation, breakpoint }, { config }) => {
     });
   }, []);
 
-  const { name, bikeParkId, stationId } = parkOrStation;
-  const parkHeaderId = bikeParkId ? 'bike-park' : 'car_park';
+  const { name, stationId } = parkOrStation;
+  const parkHeaderId = parkType === 'bike' ? 'bike-park' : 'car_park';
```
optionsome: This is doesn't change this but `-` vs `_` difference is a bit odd.

app/component/itinerary/Itinerary.js
```diff
@@ -687,35 +690,25 @@ const Itinerary = (
...
                 values={{
                   vehicle: vehicleNames[0],
-                  departureTime: firstDeparture ? (
-                    <LocalTime time={firstDeparture.startTime} />
-                  ) : (
-                    'ggh'
-                  ),
+                  departureTime: firstDeparture
+                    ? legTimeStr(firstDeparture.start)
+                    : 'ggh',
```
optionsome: I wonder what this `ggh` is.

app/component/itinerary/Legs.js
```diff
@@ -151,144 +152,42 @@ export default class ItineraryLegs extends React.Component {
...
-            focusAction={this.focus(leg.from)}
-            changeHash={this.props.changeHash}
-            tabIndex={this.props.tabIndex}
-          />,
-        );
+        legs.push(<SubwayLeg {...transitLegProps} />);
       } else if (leg.mode === 'AIRPLANE') {
         legs.push(
           <AirportCheckInLeg
+            index={j - 0.5}
```
optionsome, line 181: What do these fractional indices mean?

app/component/itinerary/Legs.js
```diff
@@ -151,144 +152,42 @@ export default class ItineraryLegs extends React.Component {
...
-        legs.push(
-          <ViaLeg
-            index={j}
-            leg={leg}
-            arrivalTime={startTime}
-            focusAction={this.focus(leg.from)}
-            focusToLeg={this.focusToLeg(leg)}
-          />,
-        );
+        legs.push(<ViaLeg {...legProps} arrival={startTime} />);
```
optionsome, line 155: This is a bit confusing that the arrival=startTime but it was like this before as well.

app/component/itinerary/Legs.js
```diff
@@ -312,40 +211,23 @@ export default class ItineraryLegs extends React.Component {
           }
           bicycleWalkLeg = {
             duration: 0,
-            startTime: 0,
-            endTime: 0,
+            start: leg.start,
+            end: leg.start,
```
optionsome, line 215: Why are these the same?

app/component/itinerary/TransitLeg.js
```diff
@@ -53,8 +57,7 @@ const filterNextLegs = leg => {
     return [];
   }
   return leg.nextLegs.filter(
-    nextLeg =>
-      moment(nextLeg.startTime).diff(moment(leg.startTime), 'hours') < 12,
+    nextLeg => legTime(nextLeg.start) - legTime(leg.start) < 12 * 3600 * 1000, // 12 hours
```
optionsome, line 60: Can we use the leg duration here or are these legs not the legs returned from OTP (or are there additional legs generated by ui)?

optionsome (changes requested):
We need to decide what to do with backward compatibility of parking pages as this changes which parking ids are accepted (previously non-feedscoped, now feedscoped).

I think there might be an issue with showing occupancy information at least on itinerary page (the icons don't show up when they do in dev).

Vehicle rental station availability (number of bicycles) is not shown on map.

On parking page, the selected parking is not highlighted on map.

On bicycle parking page, some capacity/availability number is displayed in the content as 0, it wasn't shown at all previously.

Stops near you buttons on frontpage don't show an alert which is visible in the view.

Vehicles are not shown on map in itinerary views.

Duration of a leg is shown as `NaN min` min with interlining trips (can be tested with for example kuopio config with /reitti/Vaahteratie%2024%2C%20Kuopio%3A%3A62.93671883702396%2C27.654132843017578/Korpitie%203%2C%20Kuopio%3A%3A62.85357911448784%2C27.66220092773438/?setTime=true&time=1713886213

app/component/itinerary/Legs.js
vesameskanen: Just two valid values. Prop must be given, used to be 0 for both.

app/component/itinerary/TransitLeg.js
vesameskanen: There is no precomputed duraton as difference is computed from two different legs.

app/component/itinerary/Legs.js
vesameskanen: Yes, I guess that is somehow correct.

app/component/itinerary/Legs.js
vesameskanen: Leg needs a number index prop (it is used as react key). Computation does that.

app/component/itinerary/Itinerary.js
vesameskanen: Me too. I removed it.

app/component/ParkOrStationHeader.js
vesameskanen: Right, translation ids are inconsistent. I fixed that one and removes some unused translations at the same time.

vesameskanen:
Listed bugs and some more fixed.

vesameskanen:
Added also logic for stopping slow searches. 

There is still a question if search iteration should be stopped if response contains an error message. For example, short walk search takes a long time because walking filters away all transit itineraries. Query is repeated 4 times, every time with poor results. Error code is WALKING_BETTER_THAN_TRANSIT.

optionsome:
We still need to do something about the backwards compatibility of parking pages and the issue with alert not showing on stops near you buttons seems to persists (can be tested with matka config, although might also be that somehow I'm using wrong package version). Otherwise looks like the issues were fixed and found no new issues.

vesameskanen:
I seem to get same alert triangles in stopsnearyou buttons in opas.matka.fi and this branch.
Backward compatibility of park links: it is not difficult but requires a bit ugly hsl specific configuration. I wonder if such links exist in reality, and we should take the effort.

optionsome (approved):

PR #4999, opened 2024-05-15
Put routepage into a sub folder

vesameskanen:
- Route page and its sub components moved to a sub folder 'routepage'
- Some components renamed, dead ones removed
- Some runtime proptype warnings fixed
- Moment library use reduced
- Timestore exposes unix time instead of cloned moment object, which changes at every access


app/component/routepage/TripStopListContainer.js
```diff
@@ -48,7 +47,7 @@ class TripStopListContainer extends React.PureComponent {
 
     const vehicles = groupBy(
       values(propVehicles).filter(
-        vehicle => currentTime - vehicle.timestamp * 1000 < 5 * 60 * 1000,
+        vehicle => currentTime - vehicle.timestamp < 5 * 60,
```
optionsome, line 50: In some places the timestamp is still multiplied by 1000 and in some places it isn't. Is the timestamp itself edited in this pr or the currentTime?

vesameskanen: Vehicle timestamps are unix time i.e. seconds after epoch. currentTime used to be a moment object but this PR indeed changes it to unix time.

PR #5016, opened 2024-06-14
DT-6182 rental scooters

sharhio:
## Proposed Changes

  - Scooters added to routes and map. 
  - Scooters configured for HSL and Matka (Helsinki only)

When testing 
1. Ensure that otp is configured properly (see data-container pr)
2. UI map layer configuration uses the local otp instead of dev. 

Relies on:
https://github.com/HSLdevcom/OpenTripPlanner-data-container/pull/403
https://github.com/opentripplanner/OpenTripPlanner/pull/5909

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen:
This is a mighty PR! Review will take days.

To get started, fix all new linter warnings. It is not hard - you have already added rentalVehicleShape. Just use the shapes.

app/configurations/config.hsl.js
```diff
@@ -127,6 +133,8 @@ export default {
       'mode-citybike': '#f2b62d',
       'mode-citybike-secondary': '#333333',
       'mode-speedtram': '#007E79',
+      'mode-scooter': '#BABABA',
+      'mode-scooter-secondary': '#333333',
     },
```
vesameskanen: secondary is probably not needed. All scooters use same icon.
Also, scooter color is already defined in base config (config.default.js)

app/configurations/config.hsl.js
```diff
@@ -31,6 +31,12 @@ export default {
     RENTAL_STATION_MAP: {
       default: `${POI_MAP_PREFIX}/fi/rentalStations/`,
     },
+    RENTAL_VEHICLE_MAP: {
+      default: `${POI_MAP_PREFIX}/fi/rentalVehicles/`,
+    },
+    REALTIME_RENTAL_VEHICLE_MAP: {
+      default: `${POI_MAP_PREFIX}/fi/realtimeRentalVehicles/`,
+    },
```
vesameskanen, line 42: No need to define URLs again if they are same as in default config.

app/server.js
```diff
@@ -327,6 +327,9 @@ export default async function serve(req, res, next) {
     res.write('<!doctype html>\n');
     res.write(`<html lang="${locale}">\n`);
     res.write('<head>\n');
+    /*    res.write(
+      '<script src="https://cdn.jsdelivr.net/gh/c-kick/mobileConsole/hnl.mobileconsole.min.js"></script>\n',
+    ); */
```
vesameskanen: what is this?

app/store/MapLayerStore.js
```diff
@@ -36,12 +37,23 @@ class MapLayerStore extends Store {
...
-    this.mapLayers.citybike = showCityBikes(config.cityBike?.networks, config);
-
+    this.mapLayers.citybike = showRentalVehiclesOfType(
+      config.cityBike?.networks,
+      config,
+      TransportMode.Citybike,
+    );
+    this.mapLayers.scooter =
+      config.transportModes.scooter.hideUnlessSelectedForRouting &&
+      showRentalVehiclesOfType(
```
vesameskanen, line 47: Please use conditional:
 config.transportModes.scooter?.hideUnlessSelectedForRouting
External UI users might not have scooter included in their conf.

app/util/planParamUtil.js
```diff
@@ -186,6 +202,13 @@ export function planQueryNeeded(
         settings.includeBikeSuggestions
       );
 
+    case PLANTYPE.SCOOTERTRANSIT:
+      return (
+        transitModes.length > 0 &&
+        !wheelchair &&
+        settings.allowedScooterRentalNetworks > 0
+      );
+
```
vesameskanen: Is it by design that plain scooter routes are not searched?

vesameskanen (changes requested):

app/component/itinerary/ItineraryPage.js
```diff
@@ -352,6 +368,81 @@ export default function ItineraryPage(props, context) {
...
+      return;
+    }
+
+    setRelaxRentalState({ loading: LOADSTATE.LOADING });
+    const allScooterNetworks = getAllNetworksOfType(
+      context.config,
+      TransportMode.Scooter,
+    );
+    getPlanParams(config, match, PLANTYPE.TRANSIT, true);
+
```
vesameskanen, line 400: I guess this line should be removed?

app/component/itinerary/ItineraryPage.js
```diff
@@ -352,6 +368,81 @@ export default function ItineraryPage(props, context) {
...
+    const planParams = getPlanParams(
+      config,
+      match,
+      PLANTYPE.SCOOTERTRANSIT,
+      true, // force relaxed settings
+      //   true, // force scooter query
+    );
```
vesameskanen, line 406: force what?

app/server.js
sharhio: An accident from debugging something...

app/util/planParamUtil.js
sharhio: Yes, scooters are only supposed to be shown together with some type of public transportation.

app/component/itinerary/ItineraryPage.js
sharhio: Accidental comment left from an earlier iteration.

app/configurations/config.hsl.js
vesameskanen: hmm my mistake, URL is actually not the same (hsl has own map path). I restored the change.

app/util/planParamUtil.js
```diff
@@ -296,6 +323,10 @@ export function getPlanParams(
     case PLANTYPE.TRANSIT:
       direct = access;
       break;
+    case PLANTYPE.SCOOTERTRANSIT:
+      access = ['WALK', 'SCOOTER_RENTAL'];
+      egress = access;
+      break;
```
vesameskanen, line 329: One more change: set numItineraries to 1 if searching scooter.

app/component/itinerary/ItineraryPage.js
```diff
@@ -352,6 +368,79 @@ export default function ItineraryPage(props, context) {
...
+        ...scooterState,
+        plan: {},
+        loading: LOADSTATE.UNSET,
+        error,
+      });
+    }
+  }
+
+  async function makeRelaxedScooterQuery() {
+    if (!planQueryNeeded(config, match, PLANTYPE.TRANSIT)) {
```
vesameskanen: I am not able to see relaxed scooter itineraries. Query is made but standard relaxed itinerary set is shown.

app/component/itinerary/ItineraryPage.js
```diff
@@ -717,6 +818,18 @@ export default function ItineraryPage(props, context) {
...
+    if (
+      state.loading === LOADSTATE.DONE &&
+      scooterState.loading === LOADSTATE.DONE &&
+      scooterState?.plan?.edges?.length > 0
+    ) {
+      const plan = mergeScooterTransitPlan(scooterState.plan, state.plan);
+      setCombinedRentalState({ plan });
+    }
+  }, [scooterState.plan, state.plan]);
+
```
vesameskanen, line 802: There is a minor issue: when user hits earlier button, the new set is merged with standard itineraries, not with scooter-combined itineraries. Same with later button if arrive by option is set.

app/component/RentalVehicleContent.js
```diff
@@ -0,0 +1,178 @@
+import PropTypes from 'prop-types';
```
vesameskanen, line 1: Scooter page layout seems broken - e.g.  back arrow is huge and overlaps other content. I guess it should look similar to citybike page.
image

vesameskanen (changes requested):
PlanParamUtil should set numItineraries to 1 if searching scooters.

app/component/itinerary/ItineraryPage.js
sharhio: The relaxed scooter itineraries are only supposed to show up if no standard relaxed itineraries are found. This is a rather rare situation and I have only been able to test it by breaking the search temporarily by disabling transport modes when the standard relaxed search is made.

app/util/planParamUtil.js
sharhio: The search also returns plain scooter itineraries, which are not shown on the UI. Searching for more than one itinerary gives a better chance to receive an itinerary with another mode of transport. It's not very efficient but I don't think it's possible to force the query to not return plain scooter journeys.

vesameskanen: Maybe transitOnly=true option will help?

app/component/itinerary/ItineraryPage.js
vesameskanen: ok!

app/component/itinerary/ItineraryPage.js
vesameskanen: Canot duplicate. Most likely 'user error'.

app/store/MapLayerStore.js
```diff
@@ -36,12 +37,23 @@ class MapLayerStore extends Store {
...
-    this.mapLayers.citybike = showCityBikes(config.cityBike?.networks, config);
-
+    this.mapLayers.citybike = showRentalVehiclesOfType(
+      config.cityBike?.networks,
+      config,
+      TransportMode.Citybike,
+    );
+    this.mapLayers.scooter =
+      config.transportModes.scooter?.hideUnlessSelectedForRouting &&
+      showRentalVehiclesOfType(
```
vesameskanen, line 47: This is a counter intuitive: I changed hideUnless... attribute to false in hsl conf and scooters disappeared. Hide name suggests  it should work the opposite way.

app/component/itinerary/BicycleLeg.js
```diff
@@ -244,6 +275,51 @@ export default function BicycleLeg(
             </div>
           </div>
         )}
+        {isScooter && !scooterSettingsOn && (
+          <div>
+            <div className={cx('itinerary-leg-action', 'scooter')}>
+              <div className="itinerary-leg-action-content">
```
vesameskanen: I think these divs could be combined. New div is not needed just for adding one more style class.

app/component/itinerary/CustomizeSearch.js
```diff
@@ -127,7 +151,69 @@ class CustomizeSearch extends React.Component {
...
+                      defaultMessage={intl.formatMessage({
+                        id: 'e-scooters',
+                        defaultMessage: 'Scooters',
+                      })}
+                    />
+                  </legend>
+                  <div className="transport-modes-container">
+                    {this.state.showEScooterDisclaimer && (
+                      <div className="e-scooter-disclaimer-container">
+                        <div className="e-scooter-disclaimer">
```
vesameskanen: Can we merge disclaimer-conteiner and dislaimer, as well as respective css classes, into one? Usually the simpler the better.

app/component/itinerary/Itinerary.js
```diff
@@ -1019,6 +1035,21 @@ const containerComponent = createFragmentContainer(ItineraryWithBreakpoint, {
...
+            name
+            lat
+            lon
+            network
+            rentalUris {
+              android
+              ios
+              web
+            }
+            vehicleRentalSystem {
```
vesameskanen: Are these new request fields used? Request shoudl be always minimal for each component level.

vesameskanen (changes requested):

app/util/planParamUtil.js
sharhio: Actually, now that I checked again, the problem was that we can't force the itinerary to contain a scooter journey so it's better to search more so that at least one of the itineraries contains a scooter leg as well if it's feasible.

app/store/MapLayerStore.js
sharhio: Ah, yes, apparently my brain works backwards, I reversed it to showIf... instead.

app/util/planParamUtil.js
vesameskanen: Right, that indeed makes sense.

app/component/itinerary/Itinerary.js
sharhio: Good point! Forgot to tidy this up.

app/component/RentalVehicleContent.js
sharhio: Yeah, not sure how this was missed but I adjusted it. There are no exact design pictures for the desktop scooters, only mobile version. I also tried to make sure it closely resembles the design pictures (there was a line missing).

vesameskanen (approved):

vesameskanen:
I added some minor fixes and refactored the code. There might be still a need for renaming more citybike things to a  generic form.

PR #5018, opened 2024-06-20
Add Polish translation

mkljczk:
## Proposed Changes

  - Added Polish translation, mostly done
  - Allowed EmbeddedSearch to support more languages, depending on `availableLanguages` config


vesameskanen:
Thank you for another contributon!

The PR is  otherwise fine, but library versions must be always bumped whenever they change in any way (by code or dependencies) . I took care of that.

PR #5019, opened 2024-06-20
MainMenuLinks should close MainMenu on click

mkljczk:
## Proposed Changes

  - MainMenuLinks now close menu on click

This was confusing for me on mobile


vesameskanen (commented):
Thanks for contribution, some menus  indeed behaved strangely.

I fixed some  formatting and prop validation errors. Otherwise this change is OK.

I also added a similar change for disruption menu - main menu no longer pops back when disruption modal closes.

PR #5026, opened 2024-07-04
DT-6358 Add button for attaching to an itinerary

SamiPuttonen:
## Proposed Changes

  - Starting point to giving customer a possibility to attach to a certain trip. 
  - New button added to itinerary detail. We may want to change that button's location and it's function? 

## Pull Request Check List

## Review

  - Merge the pull request


app/component/itinerary/ItineraryDetails.js
```diff
@@ -42,6 +42,15 @@ import Emissions from './Emissions';
 import EmissionsInfo from './EmissionsInfo';
 import FareDisclaimer from './FareDisclaimer';
 
+import SecondaryButton from '../SecondaryButton';
+
+const followItineraryRoute = '../../../FollowTheItinerary';
+
+const followItinerary = e => {
+  e.stopPropagation();
+  window.open(followItineraryRoute, '_blank');
```
optionsome: Instead of window.open, you should usually use `context.router.push(path)`. However, I think in this case, you could use the `<Link>` component from the found library.

app/component/itinerary/ItineraryDetails.js
```diff
@@ -330,6 +340,19 @@ class ItineraryDetails extends React.Component {
...
+              {config.itineraryFollowerOn && (
+                <div>
+                  <SecondaryButton 
+                    ariaLabel="follow" 
+                    buttonName="followtheitinerary"
+                    buttonClickAction={followItinerary}
+                    buttonIcon="icon.icon_mapMarker-via-map"
+                    smallSize /> 
+                </div>
+              )}
```
optionsome, line 353: As mentioned. I think it's better to either use `<Link>` component here directly or probably preferable create a new component which uses the `<Link>` component. I think you can somewhat mimic where `MobileTicketPurchaseInformation`  is located in this component and how it's done (although it's an external link and the design is slightly different than what the style for this link should be according to https://www.figma.com/design/kt6RO4afqFd8Ej8EmpuWpx .

app/component/itinerary/ItineraryPage.js
```diff
@@ -84,6 +84,7 @@ const LOADSTATE = {
   UNSET: 'unset',
   LOADING: 'loading',
   DONE: 'done',
+  FOLLOW: 'follow',
```
optionsome: I'm not sure if this is necessary.

app/component/itinerary/ItineraryDetails.js
optionsome: Also, we should probably hide this link in desktop view completely.

app/routes.js
```diff
@@ -304,6 +304,13 @@ export default config => {
         }}
       </Route>
 
+      <Route
+        path="/FollowTheItinerary"
+        getComponent={() =>
+          import('./component/itinerary/FollowTheItinerary').then(getDefault)
+        }
+      />
```
optionsome: @vesameskanen 's opinion was that we don't need a new route but it's sufficient just to add new state variable to the `ItineraryDetails` component. I have some difficulty to decide what is the most optimal. Having own route for this does have the benefit of the "state" being more permanent in cases where the mobile phone decides to throw the page out of memory and it's loaded with fresh state (however there are currently other potential issues when that happens as it's not guaranteed that the same itinerary will be loaded). Also, users can easily navigate out of the "follow" view by just going back to the previous page if there is a new route.

However, if we use a new route. I think the correct place for it is to be a child route of the `ItineraryDetails` component in this file and the full path would be something like`/${PREFIX_ITINERARY_SUMMARY}/:from/:to/:hash/:secondHash/${ITINERARY_FOLLOW}?`. The `:secondHash` causes some headache and maybe we should try to refactor the routes a bit because of that so it's more clear what is in `hash` or `secondHash`.  We might want to reuse the same map in ItineraryDetails and in the new follow component and not sure what is the optimal way to achieve that.

app/translations.js
```diff
@@ -1063,6 +1063,8 @@ const translations = {
     'fetch-new-route': 'Fetch a new route',
     'few-seats-available': 'Few seats available',
     finnish: 'Finnish',
+    follow: 'Follow route',
+    followtheitinerary: 'Follow your trip',
```
optionsome, line 1068: I think we use the kebab-case for the translation keys.

optionsome (changes requested):

app/routes.js
optionsome: We try to use Finnish path names (although I think we have sometimes failed at it).

hbruch: It would be great, if you could define constants for path names, so forks can replace them by English/local path names.

SamiPuttonen: Url path is now at path.js and it's in finnish. 

Currently this hole feature is still under construction by designers. So I guess that many things might change.

PR #5028, opened 2024-07-14
Added missing translations for German (de)

larsschwarz:
## Proposed Changes

  - Added a few missing German (de) translations


optionsome:
Do these look good @hbruch ?

hbruch:
The translations look good. 
Only concern I have: beyond the terms translated in this PR, there are currently about 300 terms more not yet translated to German. 

@larsschwarz would you be open to discuss a joint effort to get German translations complete?

larsschwarz:
@hbruch Totally. I just started with a few translations, because some of these are directly visible on the initial route selection UI, but I agree, having all existing items translated should be the way to go.

optionsome:
I've merged this since the translations were good but feel free to open new pr(s) to add more missing translations.

PR #5036, opened 2024-08-13
DT-6352 start search from the user location when in the configured waltti area

sharhio:
## Proposed Changes

  - When user has already given the permission to use location, add current location as the starting point for a route search by default. 
  - Only do this if the user is within the configured service area of the instance. 
  - Configurable on/off in the instance settings, configured for all waltti instances.

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/IndexPage.js
+++ b/app/component/IndexPage.js
@@ -148,11 +148,7 @@ class IndexPage extends React.Component {
       !this.props.origin.address &&
       this.props.locationState?.hasLocation &&
       this.props.locationState;
-    if (
-      config.startSearchFromUserLocation &&
-      currentLocation &&
-      !currentLocation.isReverseGeocodingInProgress
-    ) {
+    if (currentLocation && !currentLocation.isReverseGeocodingInProgress) {
       const originPoint = [currentLocation.lon, currentLocation.lat];
       if (inside(originPoint, config.areaPolygon)) {
         this.context.executeAction(storeOrigin, currentLocation);
--- a/app/component/IndexPage.js
+++ b/app/component/IndexPage.js
@@ -9,6 +9,7 @@ import DTAutoSuggest from '@digitransit-component/digitransit-component-autosugg
 import DTAutosuggestPanel from '@digitransit-component/digitransit-component-autosuggest-panel';
 import { getModesWithAlerts } from '@digitransit-search-util/digitransit-search-util-query-utils';
 import { createUrl } from '@digitransit-store/digitransit-store-future-route';
+import inside from 'point-in-polygon';
 import { configShape, locationShape } from '../util/shapes';
 import storeOrigin from '../action/originActions';
 import storeDestination from '../action/destinationActions';
@@ -36,6 +37,10 @@ import {
   getNearYouModes,
   useCitybikes,
 } from '../util/modeUtils';
+import {
+  checkPositioningPermission,
+  startLocationWatch,
+} from '../action/PositionActions';
 
 const StopRouteSearch = withSearchContext(DTAutoSuggest);
 const LocationSearch = withSearchContext(DTAutosuggestPanel);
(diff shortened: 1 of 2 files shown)


app/component/IndexPage.js
```diff
@@ -105,6 +110,18 @@ class IndexPage extends React.Component {
       this.context.executeAction(storeDestination, destination);
     }
 
+    if (this.context.config.startSearchFromUserLocation) {
+      checkPositioningPermission().then(permission => {
+        if (
+          permission.state === 'granted' &&
+          this.props.locationState.hasLocation === false &&
+          !this.props.locationState.isLocationingInProgress &&
+          this.props.locationState.status === 'no-location'
```
vesameskanen, line 117: Complex if chains are sort of unpleasant. Would it be enough to check state === 'granted' && status === 'no-location' ?

app/component/IndexPage.js
```diff
@@ -280,6 +302,17 @@ class IndexPage extends React.Component {
...
     ];
 
+    if (
+      config.startSearchFromUserLocation &&
+      currentLocation &&
+      config.areaPolygon
+    ) {
+      const originPoint = [currentLocation.lon, currentLocation.lat];
+      if (inside(originPoint, config.areaPolygon)) {
+        this.context.executeAction(storeOrigin, currentLocation);
```
vesameskanen: render should not have any side effects. Debugger gives warnings about cascading updates. Maybe this  code should be in componentDidUpdate.

vesameskanen (changes requested):

vesameskanen:
Other observations: I tested kuopio reittiopas. Origin was automatically placed to tampere. This is surprising.

app/component/IndexPage.js
```diff
@@ -280,6 +302,17 @@ class IndexPage extends React.Component {
       'Stops',
     ];
 
+    if (
+      config.startSearchFromUserLocation &&
+      currentLocation &&
+      config.areaPolygon
```
vesameskanen: It is  not necessary to test areaPolygon. Bad config should crash the UI.

app/component/IndexPage.js
sharhio: Absolutely! I think the strange Kuopio glitch was caused by the silly placement, I managed to get the same problem but only in Kuopio for some reason. It didn't happen anymore after I changed it.

app/component/IndexPage.js
```diff
@@ -128,6 +143,19 @@ class IndexPage extends React.Component {
...
     const { location } = match;
 
+    const currentLocation =
+      config.startSearchFromUserLocation &&
+      !this.props.origin.address &&
+      this.props.locationState?.hasLocation &&
+      this.props.locationState;
+
+    if (config.startSearchFromUserLocation && currentLocation) {
+      const originPoint = [currentLocation.lon, currentLocation.lat];
```
vesameskanen, line 152: This code is now executed many times. I suggest checking if currentLocation.reverseGeocoding is no longer in progress and origin and currentLocation coords are different or something like that.

vesameskanen (changes requested):

app/component/IndexPage.js
```diff
@@ -148,8 +148,11 @@ class IndexPage extends React.Component {
       !this.props.origin.address &&
       this.props.locationState?.hasLocation &&
       this.props.locationState;
-
-    if (config.startSearchFromUserLocation && currentLocation) {
+    if (
+      config.startSearchFromUserLocation &&
+      currentLocation &&
```
vesameskanen: One minor thing more: no ned to check config.startSearchFromUserLocation again, it was alredy checked when setting currentLocation.

vesameskanen (changes requested):

app/component/IndexPage.js
sharhio: Oh no, so it was!

PR #5037, opened 2024-08-13
DT-6182 rental scooter updates

sharhio:
## Proposed Changes

- Removed confusing route instructions for leaving scooters (for example the corner of street x and y).
- Adjusted the icons and itinerary timeline to align properly.
- Functionality to show closest alternative scooter from other operators on the journey.
- Support for multiple data sources under the configuration of a single rental vehicle operator.
- City specific configuration for hsl scooters.
- Scooter station journeys no longer show up in search results.
- Co2 information removed from scooter journeys.
- Fixed an error when scooter is the last leg. 
- Test fixes
- Lint fixes

_**The new configuration requires a data container update:**_ https://github.com/HSLdevcom/OpenTripPlanner-data-container/pull/410

## Pull Request Check List

## Review

  - Merge the pull request


app/component/itinerary/BicycleLeg.js
```diff
@@ -61,6 +65,16 @@ export default function BicycleLeg(
     networkConfig && networkConfig.type === RentalNetworkType.Scooter;
   const settings = getSettings(config);
   const scooterSettingsOn = settings.scooterNetworks?.length > 0;
+  const LOADSTATE = {
+    UNSET: 'unset',
+    LOADING: 'loading',
```
vesameskanen: It seems that LOADING is not used anywhere (no spinner rendering), so it can be removed.

app/component/itinerary/BicycleLeg.js
```diff
@@ -154,6 +175,58 @@ export default function BicycleLeg(
   const destination = bicycleWalkLeg?.to.stop
     ? bicycleWalkLeg?.to.name
     : leg.to.name;
+
+  async function makeNearestScooterQuery(from) {
+    setNearestScooterState({ loading: LOADSTATE.LOADING });
```
vesameskanen: LOADING state won't be used later.

app/component/itinerary/BicycleLeg.js
```diff
@@ -154,6 +175,58 @@ export default function BicycleLeg(
...
+      lat: from.lat,
+      lon: from.lon,
+      maxResults: config.cityBike.maxNearbyRentalVehicleAmount,
+      first: config.cityBike.maxNearbyRentalVehicleAmount,
+      maxDistance: config.cityBike.maxDistanceToRentalVehiclesInMeters,
+      filterByModes: ['SCOOTER'],
+      filterByPlaceTypes: ['VEHICLE_RENT'],
+    };
+
+    const result = await fetchQuery(
```
vesameskanen: I guess there should be try-catch around this await block, to handle IO errors properly.

app/component/itinerary/BicycleLeg.js
```diff
@@ -154,6 +175,58 @@ export default function BicycleLeg(
...
+      nearestQuery,
+      planParams,
+      {
+        force: true,
+      },
+    ).toPromise();
+    if (!result) {
+      setNearestScooterState({ loading: LOADSTATE.DONE });
+    } else {
+      const nearest = result?.viewer?.nearest?.edges;
```
vesameskanen: result  is not null, remove ?

app/component/itinerary/BicycleLeg.js
```diff
@@ -154,6 +175,58 @@ export default function BicycleLeg(
...
+      // filter out the ones that are not scooters or from the same network as the original scooter
+      const filteredNearest = nearest
+        ?.filter(
+          n =>
+            n?.node?.place?.__typename === 'RentalVehicle' && // eslint-disable-line no-underscore-dangle
```
vesameskanen: I would expect that nearest edges are not null/undefined. no need for n?

app/component/itinerary/RentalVehicleLinkContainer.js
```diff
@@ -0,0 +1,107 @@
...
+
+import withBreakpoint from '../../util/withBreakpoint';
+import Icon from '../Icon';
+import ExternalLink from '../ExternalLink';
+
+function RentalVehicleLinkContainer(
+  { isScooter, rentalVehicle, language, mobileReturn },
+  { config },
+) {
+  if (!rentalVehicle && !isScooter) {
```
vesameskanen: Is the component name correct? It seems to apply to scooters only. Should it be ScooterLinkContainer?

Also, I suggest making rentalVehicle  required prop and removing null check. Maybe isCooter is also irrelevant check here: it does not make sense to pass a prop isScooter = false and then return immediately null.

app/component/itinerary/RentalVehicleLinkContainer.js
```diff
@@ -0,0 +1,107 @@
...
+    return null;
+  }
+
+  const network = rentalVehicle?.rentalNetwork.networkId;
```
vesameskanen: no ? after rentalVehicle.

app/component/itinerary/WalkLeg.js
```diff
@@ -42,16 +42,16 @@ function WalkLeg(
...
+    previousLeg?.[toOrFrom]?.rentalVehicle?.rentalNetwork.networkId;
 
   const networkType = getRentalNetworkConfig(
     previousLeg?.rentedBike && network,
     config,
   ).type;
   const isScooter = networkType === RentalNetworkType.Scooter;
   const returnNotice =
-    previousLeg && previousLeg.rentedBike && !isScooter ? (
+    previousLeg && previousLeg.rentedBike ? (
```
vesameskanen: previousLeg?.rentedBike

app/configurations/config.hsl.js
```diff
@@ -491,7 +491,7 @@ export default {
         timeBeforeSurcharge: 120 * 60,
         showRentalStations: true,
       },
-      bolt: {
+      bolt_helsinki: {
```
vesameskanen, line 494: I guess this can be now plain bolt , as the network covers other cites too. Close thye respective otp data PR.

app/util/vehicleRentalUtils.js
```diff
@@ -204,26 +204,27 @@ export const mapVehicleRentalToStore = vehicleRentalStation => {
   return newStation;
 };
 
-export const getRentalVehicleLink = (rentalVehicle, network, networkConfig) => {
+export const getRentalVehicleLink = (rentalVehicle, networkConfig) => {
   if (!networkConfig || !rentalVehicle) {
     return null;
   }
 
   const { ios, android, web } = rentalVehicle?.rentalUris || {};
```
vesameskanen: rentalVehicle cannot be null, remove ?

app/util/vehicleRentalUtils.js
```diff
@@ -47,7 +47,7 @@ export const getRentalNetworkId = networks => {
 };
 
 export const getRentalNetworkConfig = (networkId, config) => {
-  if (!networkId || !networkId.toLowerCase) {
+  if (!networkId || !networkId.toLowerCase || !config) {
```
vesameskanen: can config really be null? If it is should the code crash?

This function is quite poor in general. It makes lots of nonsense checks and tries to work no matter what. Fail fast is the right way.

vesameskanen (changes requested):

app/component/itinerary/RentalVehicleLinkContainer.js
sharhio: This is the unfortunate result of a convoluted thought process where technically we could support vehicle types other than scooter but currently we only do support scooters. So, scooter link does make more sense right now.

app/util/vehicleRentalUtils.js
sharhio: Config shouldn't be checked. I think it was for debugging and then forgot to remove...

sharhio: Config shouldn't be checked. I think the check was left in by accident after debugging as it's related to the test fixes (faulty configs).

vesameskanen (approved):

PR #5043, opened 2024-08-22
DT-6182 scooter fixes

sharhio:
## Proposed Changes

Fixes:
  - Link behaviour when no app is installed. If no app is installed deeplink caused strange behaviour, in some cases a blank, black page. If navigating to deeplink fails, navigate to operator page instead, as is done also when no deeplink exists. 
  - Scooter map markers use a translated label.
  - Make sure missing transitplan edges do not break the scooter merge. A scooter itinerary can exist without other itineraries.
  - Relaxed scooter plan filtered to only contain scooter plans.
  - An unused scooter path removed.

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen (commented):
In scooter page, set scooter as start or end point from 'reitti pysäkille' button. Itinerary search origin or destination field will display english name 'scooter'.

PR #5045, opened 2024-08-26
DT-6358 Add navigator view

vesameskanen:
New view to which navigation features can be rendered.  Currently the placeholder button in ItineraryDetails is hidden; remove the false condition to test the view transition.

### DIFF
--- a/app/component/itinerary/ItineraryDetails.js
+++ b/app/component/itinerary/ItineraryDetails.js
@@ -11,6 +11,7 @@ import TicketInformation from './TicketInformation';
 import ItinerarySummary from './ItinerarySummary';
 import Legs from './Legs';
 import BackButton from '../BackButton';
+import SecondaryButton from '../SecondaryButton';
 import MobileTicketPurchaseInformation from './MobileTicketPurchaseInformation';
 import {
   compressLegs,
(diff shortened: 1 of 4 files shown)


app/component/itinerary/ItineraryDetails.js
```diff
@@ -290,6 +293,18 @@ class ItineraryDetails extends React.Component {
                   legs={itinerary.legs}
                 />
               )),
+            this.props.navigate && false && (
```
SamiPuttonen: I this navigate properties should be read at config, like config.navigate? In config file that properties is named like this: navigation. So I think we need use it here also (config.navigation).

app/component/itinerary/ItineraryPage.js
```diff
@@ -139,6 +140,7 @@ export default function ItineraryPage(props, context) {
   const [weatherState, setWeatherState] = useState({ loading: false });
   const [topicsState, setTopicsState] = useState(null);
   const [mapState, setMapState] = useState({});
+  const [navigation, navigate] = useState(false);
```
SamiPuttonen: Method name: setNavigate or setNavigation?

app/component/itinerary/ItineraryPage.js
```diff
@@ -732,6 +734,9 @@ export default function ItineraryPage(props, context) {
       if (altLoadingDone() && !mapHashToPlan()?.edges?.length) {
         selectStreetMode(); // back to root view
       }
+    } else if (navigation) {
+      // turn off tracking when user navigates away from tracking view
+      navigate(false);
```
SamiPuttonen: Method name: setNavigate?

SamiPuttonen (commented):
Little changes needed here, I think.

SamiPuttonen:
Here remember to use mobile view to see things.

PR #5050, opened 2024-08-28
DT-6393 fix ticket disclaimer typos and translations

vesameskanen:
Also, simplify configuration.

### DIFF
--- a/app/configurations/config.default.js
+++ b/app/configurations/config.default.js
@@ -420,7 +420,6 @@ export default {
   },
 
   hideExternalOperator: () => false,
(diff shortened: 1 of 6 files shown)


app/configurations/config.default.js
```diff
@@ -421,16 +421,6 @@ export default {
 
   hideExternalOperator: () => false,
   // Ticket information feature toggle
```
sharhio: This comment should be removed with the showTicketInformation setting.

vesameskanen: Well spotted! Removed now.,

PR #5057, opened 2024-09-02
Hide park and ride parking for waltti except Tampere

HenrikSundell:
## Proposed Changes

  -

## Pull Request Check List

## Review

  - Merge the pull request


app/configurations/config.tampere.js
```diff
@@ -289,4 +290,13 @@ export default configMerger(walttiConfig, {
...
 
   showTenWeeksOnRouteSchedule: true,
+
+  parkAndRide: {
+    showParkAndRide: true,
+    showParkAndRideForBikes: true,
+    parkAndRideMinZoom: 13,
+    pageContent: {
+      default: HSLParkAndRideUtils,
+    },
```
optionsome: I think these can probably be left out as I think these are inherited from the waltti config.

PR #5073, opened 2024-09-13
DT-6182 New designs for scooters

sharhio:
## Proposed Changes

 - disclaimer styles updated & configurable for instances
 - scooter info link styles (new accessible shade of blue)
 - scooter settings moved to the end of the list
 - scooter plans limited to 60 min in the future
 - direct scooter only journeys configurable for instances
 - rental vehicles layer fault tolerance
 - unnecessary kela config removed 

## Pull Request Check List

## Review

  - Merge the pull request


app/util/planParamUtil.js
```diff
@@ -198,13 +206,18 @@ export function planQueryNeeded(
...
         transitModes.length > 0 &&
         !wheelchair &&
         (relaxSettings
           ? settings.scooterNetworks.length === 0
-          : settings.scooterNetworks.length > 0)
+          : settings.scooterNetworks.length > 0) &&
+        moment(parseDateTime(arriveBy, time).earliestDeparture).diff(
+          moment(),
+          'minutes',
+        ) < config.vehicleRental.maxMinutesToRentalJourneyStart
```
vesameskanen: Does this work correctly when arriveBy== true?  Code seems to compare moment(undefined) with current time. Maybe arriveBy case needs different logic. If time is tomorrow, scooters surely should not be used.

Also, use of  parseDateTime here seems totally unnecessary as that only converts unix time  to a string and then you convert it  back to moment object.  Please use Date.now() to get epoch ms and compare it to time prop which is  epoch seconds using simple math computation. We are trying to get rid of moment.

vesameskanen (changes requested):

app/util/planParamUtil.js
sharhio: ArriveBy was accidentally overlooked. Also forgot that we are getting rid of moment. The arriveBy limit is a little hard to define as a journey could take a long time but a scooter journey might be relevant in the beginning, although there is no guarantee that the suggestion would be in the beginning of the journey. I added another configuration for the arriveBy time constraint.

vesameskanen:
Some suggestions:
- Please rename parseDateTime, it does not parse anything but does the opposite, or maybe move code back inline.
- Should after and before thresholds both use either minutes or hours? Somehow different unit looks strange.

PR #5079, opened 2024-09-17
DT-6412 Add different coloring for external ferries.

VillePihlava:
Add different coloring for external ferries in the HSL config. This needs to be coordinated with the similar changes for scooters. 

cc @sharhio


app/component/RouteNumberContainer.js
```diff
@@ -52,6 +56,8 @@ RouteNumberContainer.defaultProps = {
   className: '',
   withBicycle: false,
   occupancyStatus: undefined,
+  externalFeed: false,
+  mode: undefined,
```
vesameskanen, line 57: it does not make sense to pass these two props separately. Mode can already include external postifix.

app/component/itinerary/LegInfo.js
```diff
@@ -32,9 +33,11 @@ export default function LegInfo(
   const { constantOperationRoutes } = config;
   const shouldLinkToTrip =
     !constantOperationRoutes || !constantOperationRoutes[leg.route.gtfsId];
-  const mode = isCallAgency
+  let mode = isCallAgency
     ? 'call'
     : getRouteMode({ mode: leg.mode, type: leg.route.type });
+  const feedId = getFeedWithoutId(leg.route?.gtfsId);
+  mode = isExternalFeed(feedId, config) ? `${mode}-external` : mode;
   const capacity = getCapacityForLeg(config, leg);
```
vesameskanen, line 41: Should getRouteMode() handle the postfixing (using some new params)? I see these code lines repeating in many places.

app/component/itinerary/Legs.js
```diff
@@ -188,7 +189,11 @@ export default class Legs extends React.Component {
         !leg.interlineWithPreviousLeg
       ) {
         const mode = getRouteMode({ mode: leg.mode, type: leg.route?.type });
-        legs.push(<TransitLeg mode={mode} {...transitLegProps} />);
+        const feedId = getFeedWithoutId(leg.route?.gtfsId);
+        const changedMode = isExternalFeed(feedId, this.context.config)
+          ? `${mode}-external`
+          : mode;
+        legs.push(<TransitLeg mode={changedMode} {...transitLegProps} />);
       } else if (leg.mode === 'AIRPLANE') {
```
vesameskanen, line 199: See my comment above.

sass/themes/hsl/_theme.scss
```diff
@@ -14,6 +14,7 @@ $hsl-bus-express: #ca4000;
 $hsl-rail-red: #8c4799;
 $hsl-ferry-blue: #007a97;
 $hsl-blue-accessible: #0074bf;
+$hsl-external-gray: #c5cad2;
```
vesameskanen: I guess  this is not needed, jus define external-feeed-color directly.

app/configurations/config.hsl.js
```diff
@@ -90,6 +90,7 @@ export default {
   useRoutingFeedbackPrompt: true,
 
   feedIds: ['HSL', 'HSLlautta', 'Sipoo'],
+  internalFeedIds: ['HSL'],
```
vesameskanen, line 94: It will surely be easier to maintain list of external feeds. See for example matka config.

vesameskanen (commented):
Some general notes:
- I guess line color comes from the design, but gray map line is not well visible on the map
- The new ferry icon is  sort of small and seems not have much left margin on top of an itinerary summary bar.

app/component/RouteNumberContainer.js
VillePihlava: This is now changed

app/component/itinerary/LegInfo.js
VillePihlava: I added the logic to getRouteMode. I left the old way in one place, because it was unclear whether the change could be applied

app/component/itinerary/Legs.js
VillePihlava: I added the logic to getRouteMode. I left the old way in one place, because it was unclear whether the change could be applied

sass/themes/hsl/_theme.scss
VillePihlava: external-feed-color is now defined directly

app/configurations/config.hsl.js
VillePihlava: This has been changed to externalFeedIds

VillePihlava:
More general notes:
* The external ferries should not have route numbers, therefore the map marker is only the icon itself
  * However, in case numbers are needed in the future, the option to use black numbers with external ferries has been added
* I added translations to the translation file for ids that were producing errors
* The halo for the external line color is very visible and is not displayed for external routes because of this

image
image
image
image
image

app/component/itinerary/Itinerary.js
```diff
@@ -114,11 +115,23 @@ export function RouteLeg(
       />
     );
   } else {
+    // Because className uses leg.mode instead of just mode, the externalFeed boolean is needed separately
+    // instead of using it through getRouteMode. It is unclear whether mode could be used for both mode
+    // and className.
+    const externalFeed = isExternalFeed(
```
vesameskanen: Mode should be used - bus-express etc have their own style classes., see sass/base/_helper-classes.scss.
If something breaks then it is a bug and should  be fixed.

sass/themes/hsl/_theme.scss
```diff
@@ -46,6 +46,7 @@ $metro-color: $hsl-metro-orange;
 $rail-color: $hsl-rail-red;
 $ferry-color: $hsl-ferry-blue;
 $wait-color: #aaa;
+$external-feed-color: #c5cad2;
```
vesameskanen: This seems to be the same value as in default theme. I wonder if it is necessary to repeat the value here -  maybe it gets inherited from  the default theme.

vesameskanen (changes requested):

app/component/itinerary/Itinerary.js
VillePihlava: I removed this

sass/themes/hsl/_theme.scss
VillePihlava: The idea behind this was that because the changes were only to the HSL config, the value is supposed to be in the hsl theme. However, without a default value in the default css config, there is an error. I removed this from the HSL theme now

PR #5080, opened 2024-09-18
Fix mobile bottom sheet scroll effects on map

vesameskanen:
The code which tries to manage the map behavior by bottom sheet scrolling now makes sense and works.

### DIFF
--- a/app/component/MobileView.js
+++ b/app/component/MobileView.js
@@ -4,7 +4,9 @@ import { configShape } from '../util/shapes';
 import MapBottomsheetContext from './map/MapBottomsheetContext';
 import MobileFooter from './MobileFooter';
 
-function slowlyScrollTo(el, to = 0, duration = 1000) {
+const BOTTOM_SHEET_OFFSET = 20;
+
+function slowlyScrollTo(el, to = BOTTOM_SHEET_OFFSET, duration = 1000) {
   const element = el;
   const start = element.scrollTop;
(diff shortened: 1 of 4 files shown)


app/component/map/MapBottomsheetContext.js
```diff
@@ -1,5 +1,3 @@
-import React from 'react';
+import { createContext } from 'react';
 
-const context = React.createContext({ paddingBottomRight: [0, 0] });
-
-export default context;
+export default createContext(0);
```
Antiik91, line 4: Maybe this needs a comment to clarify meaning to this number, since without it this might seem little cryptic.

app/component/MobileView.js
```diff
@@ -4,7 +4,9 @@ import { configShape } from '../util/shapes';
 import MapBottomsheetContext from './map/MapBottomsheetContext';
 import MobileFooter from './MobileFooter';
 
-function slowlyScrollTo(el, to = 0, duration = 1000) {
+const BOTTOM_SHEET_OFFSET = 20;
+
+function slowlyScrollTo(el, to = BOTTOM_SHEET_OFFSET, duration = 1000) {
```
Antiik91, line 9: Is this change 0 -> 20 intentional?

vesameskanen: yes, otherwise navi buttons wil not show properly.

PR #5084, opened 2024-09-18
Ticket link testing for Lahti

sharhio:
## Proposed Changes

  - Ticket link configuration for Lahti

Requires fare files renamed https://github.com/HSLdevcom/digitransit-kubernetes-deploy/pull/167

## Pull Request Check List

## Review

  - Merge the pull request


app/configurations/config.lahti.js
```diff
@@ -172,4 +172,58 @@ export default configMerger(walttiConfig, {
...
+
+  showTicketLinkOnlyWhenTesting: true,
+  settingsForFeatureTesting: {
+    walkSpeed: 0.69,
+    bikeSpeed: 8.33,
+    includeBikeSuggestions: false,
+    transferPenalty: 1600,
+  },
+
+  ticketPurchaseLink: function purchaseTicketLink(fare) {
```
vesameskanen: I guess this code could be placed in waltti base config. Looks identical wth jyväskylä.

vesameskanen (changes requested):

app/util/fareUtils.js
```diff
@@ -119,6 +119,7 @@ export const shouldShowFarePurchaseInfo = (config, breakpoint, fares) => {
     !unknownFares &&
     fares?.length === 1 &&
     config.ticketPurchaseLink &&
+    config.ticketLinkOperatorCode &&
     breakpoint !== 'large'
```
vesameskanen, line 123: I wonder if this removes ticket button from hsl planner? If so, maybe add some dummy operator code like 'hsl' to its config as well.

There seems to be really many ticket related switches in the config. Let's avoid adding more if possible.

vesameskanen (changes requested):

app/util/fareUtils.js
sharhio: It does! I forgot about hsl. I added a value for it now.

PR #5102, opened 2024-09-30
DT-6400 analytics for scooters

sharhio:
## Proposed Changes

- Citybike and scooter setting analytics separated.
- Analytics event for selecting scooter settings after getting a relaxed scooter journey, using the settings button in the relaxed scooter itinerary and then selecting scooters. 

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/util/vehicleRentalUtils.js
+++ b/app/util/vehicleRentalUtils.js
@@ -138,7 +138,7 @@ const addAnalytics = (action, name) => {
  * @returns the updated citybike networks
  */
 
-export const updateVehicleNetworks = (currentSettings, newValue) => {
+export const updateVehicleNetworks = (currentSettings, newValue, type) => {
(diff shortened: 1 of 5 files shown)


app/util/vehicleRentalUtils.js
```diff
@@ -138,7 +138,7 @@ const addAnalytics = (action, name) => {
  * @returns the updated citybike networks
  */
 
-export const updateVehicleNetworks = (currentSettings, newValue) => {
+export const updateVehicleNetworks = (currentSettings, newValue, type) => {
   let chosenNetworks;
```
vesameskanen: Please update the paramete names, as well as the function docs. 

It is not a good idea to call every parameter either  'settings'  or 'value' :) 

An old sin, not your fault of course.

PR #5114, opened 2024-10-07
Add 'Stations' search target

vesameskanen:
Previously it was not possible to search stations only using digitransit search libraries. Stations were always searched as a part of 'Stops' or 'Locations' search. This is somewhat inconsistent behavior.

This PR adds a dedicated 'Stations' search target. This change makes it easier to use other than Digitransit Pelias search service with the UI.

### DIFF
--- a/digitransit-search-util/packages/digitransit-search-util-execute-search-immidiate/index.js
+++ b/digitransit-search-util/packages/digitransit-search-util-execute-search-immidiate/index.js
@@ -234,7 +234,7 @@ const routeLayers = [
   'route-AIRPLANE',
   'route-FUNICULAR',
 ];
-const locationLayers = ['favouritePlace', 'venue', 'address', 'street'];
+const locationLayers = ['venue', 'address', 'street'];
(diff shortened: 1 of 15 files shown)


digitransit-search-util/packages/digitransit-search-util-execute-search-immidiate/index.js
```diff
@@ -404,7 +407,9 @@ export function getSearchResults(
     }
   }
 
-  if (allTargets || targets.includes('Stops')) {
+  if (allTargets || targets.includes('Stops') || targets.includes('Stations')) {
+    const useStops = targets.includes('Stops');
+    const useStations = targets.includes('Stations');
```
partisaani:
```suggestion
  const useStops = targets.includes('Stops');
  const useStations = targets.includes('Stations');
  if (allTargets || useStops  || useStations ) {
```
I think this would improve the readability of the if-clause and reduce the cognitive complexity a bit.

partisaani (changes requested):

partisaani (approved):

PR #5115, opened 2024-10-08
Navitop fixes part 7: Proper handling of displayed messages and User position to naviContainer

Antiik91:
## Proposed Changes

  - Users does not see duplicate messages anymore, and dismissed messges are not shown again.
  - User location to NaviContainer

## Pull Request Check List

## Review

  - Merge the pull request


app/translations.js
```diff
@@ -1300,6 +1300,7 @@ const translations = {
     'navigation-wait': 'Wait at the stop',
     'navileg-bicycle': 'Cycle to',
     'navileg-car': 'Drive to',
+    'navileg-mode-citybike': 'Kaupunkipyöriä on asemalla {available} kpl',
```
partisaani, line 1303: In English

app/translations.js
```diff
@@ -5418,6 +5420,7 @@ const translations = {
     'navigation-wait': 'Vänta på hållplatsen',
     'navileg-bicycle': 'Cycla till',
     'navileg-car': 'Kör till',
+    'navileg-mode-citybike': 'Kaupunkipyöriä on asemalla {available} kpl',
```
partisaani, line 5423: På Svenska

app/component/itinerary/NaviContainer.js
```diff
@@ -34,11 +37,12 @@ function NaviContainer({
   relayEnvironment,
   setNavigation,
   mapRef,
+  // eslint-disable-next-line no-unused-vars
```
partisaani: Possible leftover

app/component/itinerary/NaviContainer.js
```diff
@@ -196,4 +215,12 @@ const withRelay = createFragmentContainer(NaviContainer, {
...
 
-export { NaviContainer as Component, withRelay as default };
+const NaviContainerWithTracking = connectToStores(
+  getContext({ config: configShape })(withRelay),
+  [PositionStore],
+  ({ getStore }) => ({
+    position: getStore(PositionStore).getLocationState(),
+  }),
+);
+
```
vesameskanen: store connect is not needed. We do not need to (ands should not) re-render on every position change.

app/component/itinerary/NaviContainer.js
```diff
@@ -161,6 +171,15 @@ const withRelay = createFragmentContainer(NaviContainer, {
...
           lon
+          vehicleRentalStation {
+            name
+            rentalNetwork {
+              networkId
+            }
+            availableVehicles {
+              total
+            }
+          }
```
vesameskanen, line 183: Add this also to realtime leg query.

app/component/itinerary/NaviTop.js
```diff
@@ -42,73 +42,113 @@ function findTransferProblem(legs) {
...
+        backgroundColor: '#FDF3F6',
+        iconColor: '#DC0451',
+        iconId: 'icon-icon_caution_white_exclamation',
+        id,
+      };
+    default:
+      return null;
+  }
+};
+
```
vesameskanen: This file is getting long, make a little component 'NaviMessage' out of this.

Probably NaviAlerts could also be a new component.

app/component/itinerary/NaviTop.js
```diff
@@ -42,73 +42,113 @@ function findTransferProblem(legs) {
...
 const getScheduleInfo = (nextLeg, intl) => {
-  const { start, realtimeState, to, mode } = nextLeg;
```
vesameskanen, line 46: I think 'leg' is a better name for parameter. Somehow being 'next' is irrelevant here.

app/component/itinerary/NaviTop.js
```diff
@@ -118,36 +158,33 @@ const getAlerts = (realTimeLegs, intl) => {
   const canceled = realTimeLegs.filter(leg => leg.realtimeState === 'CANCELED');
   const transferProblem = findTransferProblem(realTimeLegs);
   const late = realTimeLegs.filter(leg => leg.start.estimate?.delay > 0);
-
+  let content;
+  const id = 'alert-todo-proper-id';
   if (canceled.length > 0) {
```
vesameskanen: id is not common, alerts below can happen parallel. use unique ids.

app/component/itinerary/NaviTop.js
```diff
@@ -118,36 +158,33 @@ const getAlerts = (realTimeLegs, intl) => {
...
+    content = <div className="notifiler">Osa matkan lähdöistä on peruttu</div>;
     // Todo: No current design
     // todo find modes that are canceled
     alerts.push({
-      content: <div className="notifiler">Osa matkan lähdöistä on peruttu</div>,
-      backgroundColor: '#DC0451',
-      iconColor: '#888',
-      iconId: 'icon-icon_caution_white_exclamation',
+      ...generateStackMessage('ALERT', content, id),
     });
```
vesameskanen: no need to copy the generated alert jsx. Just push it.

app/component/itinerary/NaviTop.js
```diff
@@ -157,7 +194,10 @@ const getAlerts = (realTimeLegs, intl) => {
 function NaviTop({ focusToLeg, time, realTimeLegs }, { intl, config }) {
   const [currentLeg, setCurrentLeg] = useState(null);
   const [show, setShow] = useState(true);
+  // All notifications including those user has dismissed.
   const [notifications, setNotifications] = useState([]);
+  // notifications that are shown to the user.
+  const [displayNotifs, setdisplayNotifs] = useState([]);
   const focusRef = useRef(false);
```
vesameskanen, line 24: displayNotifs sounds like a switch to put notifications on/off. Maybe visibleNotifications ?

app/component/itinerary/NaviTop.js
```diff
@@ -263,7 +317,7 @@ function NaviTop({ focusToLeg, time, realTimeLegs }, { intl, config }) {
       </button>
       {nextLeg && (
         <NaviStack
-          notifications={notifications}
```
vesameskanen, line 266: WOuld it be more correct to check if visibleNotifications.length >0 instead of nextLeg ?

app/component/itinerary/NaviTop.js
```diff
@@ -176,46 +216,60 @@ function NaviTop({ focusToLeg, time, realTimeLegs }, { intl, config }) {
...
+          if (time < legTime(first.start)) {
+            focusToLeg(first);
+          } else {
+            focusToLeg(last);
+          }
         }
+        focusRef.current = true;
+      }
+      if (notifs.length > 0 || currentNots.length < displayNotifs.length) {
+        const combined = currentNots.concat(...notifs);
```
vesameskanen: Nofits  - currentnots - displaynotifs -  notifications ... this is very hard to understand.  Please try to refactor. Maybe better names will help. What is 'main' notification, for example?

The code is getting so complicated that we should make all efforts to make it readable again.

app/component/itinerary/NaviTop.js
```diff
@@ -244,15 +298,15 @@ function NaviTop({ focusToLeg, time, realTimeLegs }, { intl, config }) {
...
   const handleRemove = index => {
-    setNotifications(notifications.filter((_, i) => i !== index));
+    setdisplayNotifs(displayNotifs.filter((_, i) => i !== index));
   };
 
   return (
     <>
       <button type="button" className="navitop" onClick={handleClick}>
         <div className="info">{info}</div>
         <div type="button" className="navitop-arrow">
```
vesameskanen, line 133: Time to call 'info' as something more descriptive. Maybe naviTopContent?

app/component/itinerary/NaviTop.js
```diff
@@ -176,46 +216,60 @@ function NaviTop({ focusToLeg, time, realTimeLegs }, { intl, config }) {
...
+      let currentNots = displayNotifs;
+      if (!isSame) {
+        // remove Old main notification when new leg is started.
+        currentNots = currentNots.filter(n => n.type !== 'main');
+        if (newLeg) {
+          focusToLeg?.(newLeg);
+          setCurrentLeg(newLeg);
+        }
+        const alerts = getAlerts(realTimeLegs, intl);
+        if (alerts.length > 0) {
```
vesameskanen: This is not correct. Alerts can arise anytime during the trip, not only when current leg changes.

app/component/itinerary/NaviTop.js
Antiik91: This is meant to be a placeholder for now, untill I start to develop correct alert messages. Then we get the proper id's, which are unique.

partisaani (changes requested):

app/component/itinerary/NaviTop.js
Antiik91: Decided to go with activeNotifications. Since those are the notifications that are active as a user perspective.

app/component/itinerary/NaviTop.js
Antiik91: Reworked around this a bit. Should be more clear now

app/component/itinerary/NaviContainer.js
Antiik91: Got it. I will revert this change, and do the location in another pr, since it will contain more changes regarding it as well.

PR #5123, opened 2024-10-16
DT-6446 New analytics for Waltti

sharhio:
## Proposed Changes

  - Configurable analytics script
  - Configured for Waltti and excluded for Varely as it uses waltti config but is not under waltti
  - Hostname taken from request to distinguish between different aliases of the same instances (fe. reittiopas.hameenlinna vs hameenlinna.digitransit)
  - Dev and test environments excluded

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/configurations/config.varely.js
+++ b/app/configurations/config.varely.js
@@ -170,4 +170,5 @@ export default configMerger(walttiConfig, {
   showNearYouButtons: true,
   allowLogin: false,
   routeNotifications: [],
+  analyticsScript: '',
 });
--- a/app/util/analyticsUtils.js
+++ b/app/util/analyticsUtils.js
@@ -35,18 +35,18 @@ export function addAnalyticsEvent(event) {
  *
  * @return string
  */
-export function getAnalyticsInitCode(GTMid, analyticsScript, hostname) {
-  if (analyticsScript && hostname && !hostname.match(/dev|test/)) {
-    return analyticsScript(hostname);
+export function getAnalyticsInitCode(config, hostname) {
+  if (config.analyticsScript && hostname && !hostname.match(/dev|test/)) {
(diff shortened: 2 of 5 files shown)


app/configurations/config.varely.js
```diff
@@ -170,4 +170,5 @@ export default configMerger(walttiConfig, {
   showNearYouButtons: true,
   allowLogin: false,
   routeNotifications: [],
+  analyticsScript: '',
```
vesameskanen, line 173: maybe not needed? Empty string vs function in waltti config.

app/util/analyticsUtils.js
```diff
@@ -35,16 +35,20 @@ export function addAnalyticsEvent(event) {
  *
  * @return string
  */
-export function getAnalyticsInitCode(GTMid) {
-  if (!GTMid) {
-    return '';
+export function getAnalyticsInitCode(GTMid, analyticsScript, hostname) {
```
vesameskanen: I suggest changing this to getAnalyticsInitCode(config, hostname)

vesameskanen (changes requested):

app/configurations/config.varely.js
sharhio: The config is needed here to stop the analytics script from being added to varely instance from waltti config. Empty string is the minimalist way to stop it but I was wondering if it's a little bit confusing.

partisaani: Especially since the field is called a script, would a void function `() => {}` be less confusing?

vesameskanen: Empty string works as javascript thinks it is a 'falsy' value.  I guess we can accept this.

vesameskanen (approved):

PR #5146, opened 2024-11-01
DT-6483: Refactor top card

Antiik91:
## Proposed Changes

  - Top Card is now expandable, clicking the card show additional information about leg's destination.
  - Refactored and renamed components for clarity. 

## Pull Request Check List

## Review

  - Merge the pull request


app/component/itinerary/NaviCardContainer.js
```diff
@@ -98,6 +85,7 @@ function NaviTop(
       if (legChanged) {
         focusToLeg?.(newLeg);
```
vesameskanen, line 85: if ( newLeg) ...

app/component/itinerary/NaviCardContainer.js
```diff
@@ -98,6 +85,7 @@ function NaviTop(
...
         setCurrentLeg(newLeg);
+        setCardExpanded(false);
       }
 
       if (incomingMessages.size || legChanged) {
```
vesameskanen: This code block is in wrong level.  Currentleg and newLeg ban be null while waiting, stiull new messages may appear.

Also  if (!focusRef.current && focusToLeg) {}  block is under wrong if block

app/component/itinerary/NaviCardContainer.js
```diff
@@ -163,11 +158,21 @@ function NaviTop(
...
-          <NaviLeg leg={currentLeg} nextLeg={nextLeg} legType="wait" />
+          <NaviCard
+            leg={currentLeg}
+            nextLeg={nextLeg}
+            cardExpanded={cardExpanded}
+            legType="wait"
+          />
         );
       } else {
         naviTopContent = (
```
vesameskanen: no need to duplicate jsx item, just set legType conditionally

app/component/itinerary/NaviStack.js
```diff
@@ -4,9 +4,12 @@ import cx from 'classnames';
...
 // eslint-disable-next-line no-unused-vars
-const NaviStack = ({ messages, handleRemove, show }) => {
+const NaviStack = ({ messages, handleRemove, cardExpanded }) => {
+  // Becuase of the way the NaviCard is placed (fixed) we need to adjust the top position depending
+  // on if the  card is expanded or not
+  const top = cardExpanded ? '212px' : '150px';
   return (
-    <div className={cx('info-stack', !show ? 'slide-out' : 'slide-in')}>
+    <div className={cx('info-stack', 'slide-in')} style={{ top }}>
       {messages.map((notification, index) => (
```
vesameskanen, line 9: style={{'212px'}}  won't work

app/component/itinerary/NaviCardContainer.js
```diff
@@ -108,11 +96,18 @@ function NaviTop(
           ? activeMessages.filter(m => m.expiresOn !== 'legChange')
           : activeMessages;
 
+        // handle messages that are updated.
+        const updatedMessages = currActiveMessages.map(msg => {
+          const incoming = incomingMessages.get(msg.id);
```
vesameskanen: How about renaming  currActiveMessages  -> previousValidMessages ?

vesameskanen:
NaviSecondaryInfo -> naviCardExtension ?

app/component/itinerary/NaviCardContainer.js
Antiik91: I'm confused as why? 

 newLeg is defined like this:   ```const newLeg = realTimeLegs.find(leg => {
      return legTime(leg.start) <= time && time <= legTime(leg.end);
    });``` in useEffect.

 So basically every 10 seconds we get the same leg until we eventually get the new leg. Thats why I introduced the `legChanged` param to handle the situation where the leg is actually changed.

app/component/itinerary/NaviStack.js
Antiik91: True, this will be changed when the PR is out of draft.

app/component/itinerary/NaviCardContainer.js
Antiik91: Valid point! Will fix this!

app/component/itinerary/NaviCardContainer.js
Antiik91: Done.

app/component/itinerary/navigator.scss
partisaani, line 1: Values defined in `px` could be revisited against [_spacing.scss](https://github.com/HSLdevcom/digitransit-ui/blob/v3/sass/base/_spacing.scss)

app/component/itinerary/NaviUtils.js
```diff
@@ -168,3 +169,68 @@ export const getItineraryAlerts = (realTimeLegs, intl, messages) => {
...
+        iconOptions.iconId = 'icon-icon_bus-stop-express-lollipop';
+        iconOptions.className = 'bus-stop';
+        break;
+      case 'speedtram':
+        iconOptions.iconId = 'icon-icon_speedtram-stop-lollipop';
+        iconOptions.className = 'speedtram-stop';
+        break;
+      default:
+        iconOptions.iconId = `icon-icon_${mode.toLowerCase()}-stop-lollipop`;
+    }
```
partisaani: Complexity could be eased by using an object literal

app/component/itinerary/NaviCardContainer.js
```diff
@@ -37,17 +36,9 @@ function NaviTop(
...
 
-  useEffect(() => {
-    const timer = setTimeout(() => {
-      setShowMessages(false);
-    }, 5000);
-
-    return () => clearTimeout(timer);
-  }, []);
-
   useEffect(() => {
```
partisaani, line 42: This looks like a good candidate for a custom hook as the logic inside `useEffect` is all in all quite multifaceted.

partisaani (changes requested):

app/component/itinerary/NaviCardContainer.js
Antiik91: Yeah, this needs to be refactored, but I think it would be more suited for seperate PR

app/component/itinerary/navigator.scss
Antiik91: I'll check on this. There is some issues for example card height, that is still hard coded for desings, and there is no corresponding variable in _spacings.scss

vesameskanen (commented):
Looks good to me. Review comment about object literals is an imporovement, though.

vesameskanen (approved):

PR #5150, opened 2024-11-05
DT-6476 Store and retrieve itinerary for Navigator from localStorage

partisaani:
## Proposed Changes

  - Stores a started Navigator itinerary to localStorage where it can then be retrieved when browser session is reinstated (i.e. page refresh or on reopening a locked smart device)

## Known issues
- `ItineraryPage` test suite and the feature's unit tests are missing for now. Mocking `ItineraryPage` so that the feature could be tested has proven to be quite a timeconsuming task itself.

## Pull Request Check List

## Review

  - Merge the pull request


app/component/itinerary/ItineraryPage.js
```diff
@@ -704,39 +746,17 @@ export default function ItineraryPage(props, context) {
...
-
   useEffect(() => {
     setCurrentTimeToURL(config, match);
     updateLocalStorage(true);
     addFeedbackly(context);
+
+    const storedItinerary = getLatestNavigatorItinerary();
+
+    setNavigation(
+      storedItinerary?.end && moment(storedItinerary.end).isAfter(Date.now()),
```
vesameskanen: We are trying to get rid of moment.  So use storedItinerary.end > Date.now()

app/component/itinerary/ItineraryPage.js
```diff
@@ -1130,10 +1154,9 @@ export default function ItineraryPage(props, context) {
             />
           )}
           <NaviContainer
-            itinerary={combinedEdges[selectedIndex]?.node}
+            itinerary={naviModeItinerary}
```
vesameskanen, line 1156: I noticed that NaviContainer actually only needs the legs array. Time to change the prop to legs?

app/component/itinerary/NaviContainer.js
```diff
@@ -96,6 +97,10 @@ function NaviContainer(
           return { ...l };
         });
         setRealTimeLegs(rtLegs);
+        setLatestNavigatorItinerary({
+          ...itinerary,
+          legs: rtLegs,
+        });
       });
```
vesameskanen: I wonder if it is necessary to re-save the itinerary all the time? Page reload might briefly show initial itinerary before it gets updated, but I am not sure if this is a problem. what is your opinion? Maybe the realtime query should be made when the component mounts.

vesameskanen (changes requested):
Currently there is a problem: if user exits the navigation view by closing the browser or by browsing to a new address, and later makes a new itinerary search and opens an itinerary, navigation activates immediately. Also, opening a direct link to an itinerary might actually end up showing another itinerary.

Possible mitigations:
- Itinerary page could erase saved itinerary at page mounting if url path does not specify the detail view
- Save url with itinerary and restore navigation only when urls match

app/component/itinerary/ItineraryPage.js
partisaani: I recall making the same observation during development as well, but I cannot remember why I did not include it as a tentative change. I'll refactor to pass legs only if possible.

app/component/itinerary/NaviContainer.js
partisaani: Refactoring the component to send the first query on mount sounds like a better solution. I'll refactor.

partisaani:
> Currently there is a problem: if user exits the navigation view by closing the browser or by browsing to a new address, and later makes a new itinerary search and opens an itinerary, navigation activates immediately. Also, opening a direct link to an itinerary might actually end up showing another itinerary.

I will revisit the feature to tackle this and the rest of the issues that you pointed out.

partisaani:
I tried to debug the described issue by navigating away from an ongoing Navigator journey and then back, but I am unable to reproduce the issue.

https://github.com/user-attachments/assets/fb65beb5-99ec-430c-af41-947b0708dfa2

PR #5153, opened 2024-11-08
Use graphQL alias to query leg id

vesameskanen:
When leg's id is renamed as legId, relay no longer changes fare product id information.


partisaani (approved):

partisaani:
LGTM 👍

PR #5155, opened 2024-11-11
DT-6531 switch deprecated plan query to a new one

vesameskanen:
Also refactor walk routing in Stop and Stops near you maps


partisaani (approved):
LGTM

PR #5160, opened 2024-11-12
DT-6532:  In vehicle cards

Antiik91:
## Proposed Changes

  - Shows in transit information in navigator
  - 

## Pull Request Check List

## Review

  - Merge the pull request


app/component/itinerary/NaviCardExtension.js
partisaani, line 1: DRY and KISS; I would refactor the two embedded return values as individual components for reduced complexity of this parent component
```jsx
return (
  <div className="extension">
    <div className="extension-divider" />
    {legType === 'in-transit' ? <TransitLegExtender {...props} /> : <WalkLegExtender {...props} />}
  </div>
)
```
also simplifies testing for all of the three components as for parent you can just sinon.stub() the `*Extender` components with an easily locatable attribute and for the `*Extender`s you can just make sure the component looks as it should with given props

partisaani (changes requested):

app/component/itinerary/NaviCardExtension.js
Antiik91: Good point, although I reckon that we get a lot of navigation components before we are finished with this, so we don't want to spread too thin here.

I'll take another opinion here as well before I start to refactor this.

app/component/itinerary/NaviCardContainer.js
```diff
@@ -151,27 +151,26 @@ function NaviCardContainer(
...
-          leg={currentLeg}
-          nextLeg={nextLeg}
-          cardExpanded={cardExpanded}
-          legType={legType}
-        />
-      );
     } else {
-      naviTopContent = `Tracking ${currentLeg?.mode} leg`;
+      legType = 'in-transit';
     }
```
vesameskanen: Maybe handle leg types like this:

```
export const LEGTYPE = {
  WAIT: 'WAIT',
  MOVE: 'MOVE',
  TRANSIT: 'TRANSIT',
}
```

Add this to  naviUtils and import it from there to other classes which refer to leg types.

app/component/itinerary/NaviCardContainer.js
```diff
@@ -151,27 +151,26 @@ function NaviCardContainer(
...
+      <NaviCard
+        leg={currentLeg}
+        nextLeg={nextLeg}
+        cardExpanded={cardExpanded}
+        legType={legType}
+      />
+    );
   } else if (time > legTime(last.end)) {
     naviTopContent = <FormattedMessage id="navigation-journey-end" />;
   } else {
```
vesameskanen, line 177: There are now 2 separate ways to render wait state. Maybe these formatted messages could be moved to NaviCard? In other words,  leg would not be required prop and legType could be also `ITINERARY_PENDING` and `ITINERARY_ENDED` or something like that.

app/component/itinerary/NaviCard.js
```diff
@@ -12,13 +12,19 @@ const iconMap = {
...
   WAIT: 'icon-icon_navigation_wait',
+  BUS: 'icon-icon_bus',
+  RAIL: 'icon-icon_rail',
+  SUBWAY: 'icon-icon_subway',
+  TRAM: 'icon-icon_tram',
+  FERRY: 'icon-icon_ferry',
 };
 
 export default function NaviCard({ leg, nextLeg, legType, cardExpanded }) {
   const iconName = legType === 'wait' ? iconMap.WAIT : iconMap[leg.mode];
```
vesameskanen: nextLeg is not a required prop, can be undefined

app/component/itinerary/NaviCardExtension.js
vesameskanen: I don't have a stong opinion about this. I guess we could put all this navi stuff to a sub folder first, and after that start splitting into smaller components.

app/component/itinerary/NaviCardExtension.js
```diff
@@ -21,10 +25,51 @@ const NaviCardExtension = ({ leg }, { config }) => {
...
+    const now = new Date();
+    const idx =
+      intermediatePlaces.findIndex(
+        p =>
+          new Date(
+            p.arrival.estimated?.time || p.arrival.scheduledTime,
+          ).getTime() > now.getTime(),
+      ) ?? -1;
+
+    const count = idx > -1 ? intermediatePlaces.length - idx : 0;
```
vesameskanen: Use Date.now() and legTime.

app/component/itinerary/NaviCardExtension.js
```diff
@@ -21,10 +25,51 @@ const NaviCardExtension = ({ leg }, { config }) => {
...
+    const stopCountElement = <span className="realtime"> {count}</span>;
```
vesameskanen: I guess plain `stopCount` isa bit  simpler. We don't have a convention to name jsx snippets as elements.

app/component/itinerary/NaviStack.js
```diff
@@ -29,10 +39,12 @@ NaviStack.propTypes = {
   ).isRequired,
   handleRemove: PropTypes.func.isRequired,
   cardExpanded: PropTypes.bool,
+  legType: PropTypes.string,
 };
 
 NaviStack.defaultProps = {
   cardExpanded: false,
+  legType: '',
```
vesameskanen: legType does not have logical connection to NaviStack which may contain messages for many legs. I suggest passing the class postfix itself, or maybe handling the vertical offset completely in parent component.

vesameskanen (changes requested):

app/component/itinerary/NaviCardExtension.js
Antiik91: I'll create the subfolder in this pr. And after I start refactoring the wait leg, I'll get back to this.

app/component/itinerary/NaviCardContainer.js
Antiik91: This will be considered in the next PR, where I'll refactor the wait leg

vesameskanen (approved):

PR #5162, opened 2024-11-12
DT-6533 Improved navigator and local storage cooperation

partisaani:
## Proposed Changes
Navigator is now more context aware:
  - If `from`, `to`, `time`, `arriveBy` and `hash` parameters are identical to those of the stored journey
    - Traversing to `ItineraryPage` detail view will automatically trigger the Navigator journey.
  - If `from`, `to`, `time` and `arriveBy` parameters are identical to those of the stored journey
    - Traversing to `ItineraryPage` list view does not necessarily purge the stored itinerary from memory.
      - Local storage entry is purged if user navigates to list view from active navigator journey either by using the arrow icon or by browsing back using browser history
    - Traversing to `ItineraryPage` detail view does not purge the stored itinerary from memory, but a Navigator journey is not triggered
  - If any of the parameters is different
    - Traversing to `ItineraryPage` list view or detail view will purge the stored Navigator journey from local storage

## Pull Request Check List

  - [x] A reasonable set of unit tests is included

## Review

  - Merge the pull request

### DIFF
--- a/app/component/itinerary/ItineraryDetails.js
+++ b/app/component/itinerary/ItineraryDetails.js
@@ -1,17 +1,15 @@
+import cx from 'classnames';
+import connectToStores from 'fluxible-addons-react/connectToStores';
+import { matchShape, routerShape } from 'found';
 import PropTypes from 'prop-types';
 import React from 'react';
-import { createFragmentContainer, graphql } from 'react-relay';
-import cx from 'classnames';
-import { matchShape, routerShape } from 'found';
 import { FormattedMessage, intlShape } from 'react-intl';
-import connectToStores from 'fluxible-addons-react/connectToStores';
-import { configShape, itineraryShape, relayShape } from '../../util/shapes';
-import TicketInformation from './TicketInformation';
-import ItinerarySummary from './ItinerarySummary';
-import Legs from './Legs';
-import BackButton from '../BackButton';
-import StartNavi from './StartNavi';
-import MobileTicketPurchaseInformation from './MobileTicketPurchaseInformation';
+import { createFragmentContainer, graphql } from 'react-relay';
+import {
+  getFaresFromLegs,
+  shouldShowFareInfo,
+  shouldShowFarePurchaseInfo,
+} from '../../util/fareUtils';
 import {
   compressLegs,
   getTotalBikingDistance,
@@ -25,22 +23,24 @@ import {
   legContainsBikePark,
   legContainsRentalBike,
 } from '../../util/legUtils';
-import { BreakpointConsumer } from '../../util/withBreakpoint';
 import { streetHash } from '../../util/path';
-import {
-  getFaresFromLegs,
-  shouldShowFareInfo,
-  shouldShowFarePurchaseInfo,
-} from '../../util/fareUtils';
+import { configShape, itineraryShape, relayShape } from '../../util/shapes';
 import {
   getFormattedTimeDate,
   isToday,
   isTomorrow,
 } from '../../util/timeUtils';
-import VehicleRentalDurationInfo from './VehicleRentalDurationInfo';
+import { BreakpointConsumer } from '../../util/withBreakpoint';
+import BackButton from '../BackButton';
 import Emissions from './Emissions';
 import EmissionsInfo from './EmissionsInfo';
 import FareDisclaimer from './FareDisclaimer';
+import ItinerarySummary from './ItinerarySummary';
+import Legs from './Legs';
+import MobileTicketPurchaseInformation from './MobileTicketPurchaseInformation';
+import StartNavi from './StartNavi';
+import TicketInformation from './TicketInformation';
+import VehicleRentalDurationInfo from './VehicleRentalDurationInfo';
 
 class ItineraryDetails extends React.Component {
   static propTypes = {
--- a/app/component/itinerary/ItineraryDetails.js
+++ b/app/component/itinerary/ItineraryDetails.js
@@ -42,7 +42,6 @@ import Emissions from './Emissions';
 import EmissionsInfo from './EmissionsInfo';
 import FareDisclaimer from './FareDisclaimer';
 
-/* eslint-disable prettier/prettier */
 class ItineraryDetails extends React.Component {
   static propTypes = {
     itinerary: itineraryShape.isRequired,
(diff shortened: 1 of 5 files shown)


app/component/itinerary/ItineraryDetails.js
partisaani, line 1: Whitespace and format changes only

app/component/itinerary/ItineraryPage.js
```diff
@@ -745,16 +739,63 @@ export default function ItineraryPage(props, context) {
...
   }
 
+  /**
+   * Checks if itinerary in local storage was set with identical parameters.
+   *
+   * @param {{from: string, to: string, time: number, arriveBy: boolean, hash: number}} storedParams
+   * @param {boolean} shouldCompareHash false bypasses hash check when user navigates to ItineraryPage root with the list of suggested itineraries
+   * @returns true if stored itinerary was set with identical parameters
+   */
+  const isMatchingParams = (storedParams, shouldCompareHash) => {
```
Antiik91: I think these should be moved in some util file if possible, in order not to bloat the ItineraryPage more than necessary.

Antiik91 (approved):

app/component/itinerary/ItineraryPageUtils.js
```diff
@@ -535,3 +536,50 @@ export function quitIteration(plan, newPlan, planParams, startTime) {
...
+export const storedParamsMatchURL = (storedParams, match) => {
+  if (!storedParams || !match) {
+    return false;
+  }
+
+  return (
+    storedParams.from === match?.params?.from &&
+    storedParams.to === match?.params?.to &&
+    storedParams.time === match?.location?.query?.time &&
+    storedParams.arriveBy === match?.location?.query?.arriveBy
```
vesameskanen: no need for match?, it is tested in the beginning of the function. Actually, can match be undefined?

app/component/itinerary/StartNavi.js
```diff
@@ -1,18 +1,34 @@
...
+  const handleClick = () => {
+    setNavigation(true);
+    setLatestNavigatorItinerary({
+      itinerary,
+      params: {
+        from: match.params.from,
+        to: match.params.to,
+        arriveBy: match.location.query.arriveBy,
+        time: match.location.query.time,
+        index: match.params.hash,
```
vesameskanen: Unfortunately, we have to deal with secondaryHash as well.

app/component/itinerary/ItineraryPage.js
```diff
@@ -750,11 +746,13 @@ export default function ItineraryPage(props, context) {
...
-    setNavigation(
-      storedItinerary?.end && Date.parse(storedItinerary.end) > Date.now(),
-    );
+    const { params: storedParams } = getLatestNavigatorItinerary();
+    if (
+      (hash === undefined || hash === null) &&
+      !storedParamsMatchURL(storedParams, match)
+    ) {
+      clearLatestNavigatorItinerary();
+    }
```
vesameskanen: I think you need something like this here:
 ```
 if (isStoredItineraryRelevant(match)) { 
    setnavigation(true);
} else {
  clearLatestNavigatorItinerary();
}
```
relevance testing can handle all  hash checking logic

File: app/component/itinerary/ItineraryPage.js [hunk 5]
```diff
@@ -776,6 +774,9 @@ export default function ItineraryPage(props, context) {
       makeRelaxedQuery();
       makeRelaxedScooterQuery();
     }
+    if (isStoredItineraryRelevant(match)) {
+      setNavigation(true);
+    }
   }, [
```
vesameskanen: I don't think this is necessary. Itinerary search parameter change never triggers navi mode.

app/component/itinerary/ItineraryPage.js
```diff
@@ -787,6 +788,9 @@ export default function ItineraryPage(props, context) {
   useEffect(() => {
     navigateMap();
     setMapState({ center: undefined, zoom: undefined, bounds: undefined });
+    if (isStoredItineraryRelevant(match)) {
+      setNavigation(true);
+    }
```
vesameskanen: Not necessary. Hash change should not trigger navi mode.

vesameskanen (changes requested):
Note that when you open itinerary detaul view of a  bike and transit itinerary, hash is not itinerary index. It can be found from secondaryHash.

As far as I can see, we need to 'auto relaunch' navigation only in one occasion: page reloading. This is handled in useEffect [].

app/component/itinerary/StartNavi.js
partisaani: Will add

app/component/itinerary/ItineraryPageUtils.js
partisaani: It is likely match cannot be undefined, but I don't think being prepared for such occasion is detrimental

app/component/itinerary/ItineraryPage.js
partisaani: Will remove

app/component/itinerary/ItineraryPage.js
partisaani: Will remove

partisaani:
> As far as I can see, we need to 'auto relaunch' navigation only in one occasion: page reloading. This is handled in useEffect [].

Will refactor with this in mind. For possible future reference the current implementation enables the user to browse between itineraries and their detail views. Browsing to the exact same search result auto-launches Navigator.

Only explicit backtrack from an initiated journey or a change in `from`, `to`, `time` or `arriveBy` params signal to purge the stored itinerary.

app/component/itinerary/ItineraryPageUtils.js
```diff
@@ -535,3 +536,52 @@ export function quitIteration(plan, newPlan, planParams, startTime) {
...
+  } = match;
+
+  return (
+    (itinerary &&
+      Date.parse(itinerary.end) > Date.now() &&
+      storedParamsMatchURL(params, match) &&
+      params?.index == secondHash) ||
+    params?.index == hash
+  );
+};
```
vesameskanen, line 563: - last condition is apparently wrong, shoudl be ( params.index == secondHash || params.index == hash )
- would it be easier and safer to save hash and secondHash and compare both ?
- let code crash if match params is not defined. unnecessary validation to keep code running  although everyhing is all wrong is not wise

vesameskanen (changes requested):

app/component/itinerary/ItineraryPageUtils.js
partisaani: Refactored and fixed.

vesameskanen (approved):
very good

PR #5165, opened 2024-11-14
Development feature flag for navigation

vesameskanen:
Navigation shows up only in local server and in development environment


partisaani:
I'd like to suggest we nest feature flagged properties under a set feature flag object so that their nature is crystal clear.

vesameskanen:
please push a suggested change to this branch

vesameskanen:
Actually, large part of configuration consists of 'feature flags'. A value will trigger a feature which appears in the UI.

partisaani:
That is very much true, and my wording was not as well thought initially. What I meant was to have i.e.
```javascript
  experimental: {
    navigation: process.env.RUN_ENV === 'development' || process.env.NODE_ENV !== 'production',
  },
```
which would help spotting these in development features among the other properties that are used to flag features on and off between service providers.

If you feel like it's just an unnecessary overhead, the suggestion may be disregarded - the change itself looks good, and I am for merging it to `v3`

partisaani (approved):
LGTM

PR #5166, opened 2024-11-15
DT-6476 Refactor navigation initializer into callback function

partisaani:
## Proposed Changes
Refactors navigation initialization into a callback function passed down from `ItineraryPage`. The current implementation uses an unexpected itinerary version as it lacks some fragment properties (i.e. `headSign`) due to react-relay logic if the stored itinerary is the one passed down to `StartNavi` from `ItineraryDetails`

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/itinerary/ItineraryDetails.js
+++ b/app/component/itinerary/ItineraryDetails.js
@@ -53,7 +53,7 @@ class ItineraryDetails extends React.Component {
     currentLanguage: PropTypes.string,
     changeHash: PropTypes.func,
     openSettings: PropTypes.func.isRequired,
-    setNavigation: PropTypes.func,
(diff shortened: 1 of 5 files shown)


vesameskanen (commented):
Names! 

I would call the function prop as startNavigation. 
- what does 'on' prefix indicate ? I find it just confusing.
- We could always add 'callback' postfix to function props, and 'xxxComponent' to all component names, etc., but should we ? :)

partisaani:
> I would call the function prop as startNavigation.

Good suggestion. Will refactor.

> what does 'on' prefix indicate? I find it just confusing.

I've seen it as a convention to indicate that something is triggered by user action and in this case it's the onClick of `StartNavi` element

PR #5172, opened 2024-11-21
Refactor location search target configuration code 

vesameskanen:
In a recent search change stations were added as a separate search target. Unfortunately itinerary page's search missed addition of the new target.

Now the fairly complex configuration code, which was earlier repeated in many places , is  available in WithSearchContext.js and is used from there in every search instance.


partisaani (approved):
LGTM

PR #5173, opened 2024-11-21
DT-6552 Ability to pass explicit itinerary for ItineraryPageMap

partisaani:
## Proposed Changes
ItineraryPageMap now correctly renders the itinerary retrieved from localStorage

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/itinerary/ItineraryPage.js
+++ b/app/component/itinerary/ItineraryPage.js
@@ -74,9 +74,9 @@ import {
   updateClient,
 } from './ItineraryPageUtils';
 import ItineraryTabs from './ItineraryTabs';
+import planConnection from './PlanConnection';
 import NaviContainer from './navigator/NaviContainer';
 import NavigatorIntroModal from './navigator/navigatorintro/NavigatorIntroModal';
-import planConnection from './PlanConnection';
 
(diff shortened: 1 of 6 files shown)


app/component/itinerary/ItineraryPage.js
```diff
@@ -151,6 +151,9 @@ export default function ItineraryPage(props, context) {
   const [topicsState, setTopicsState] = useState(null);
   const [mapState, setMapState] = useState({});
   const [naviMode, setNaviMode] = useState(false);
+  const [storedItinerary, setStoredItinerary] = useState(
+    getLatestNavigatorItinerary(),
+  );
```
vesameskanen, line 157: I wonder if this should be put on a ref. State changes trigger rendering and I do not think stored itinerary is data which is supposed to do that.

vesameskanen (commented):
It is a bit hard to understand if the itinerary put on local state should also be cleared together with the stored item. Seems to work fine though.

app/component/itinerary/ItineraryPage.js
partisaani: I was thinking about `useRef` as well but the fact that we have `storeItineraryAndStartNavigation` function in the component as well steered me towards `useState`.

PR #5174, opened 2024-11-22
Reconstruct itinerary after realtime modifications

vesameskanen:
When realtime transit legs change, non-transit legs before, after and in-between them must be shifted and scaled so that the realtime modified itinerary providwes a consistent chain of legs.

### DIFF
--- a/app/util/timeUtils.js
+++ b/app/util/timeUtils.js
@@ -166,9 +166,5 @@ export function timeStr(dateTime) {
  * Epoch ms to ISO-8601/RFC3339 datetime str
  */
 export function epochToIso(ms) {
-  const date = new Date(ms);
-  let iso = date.toISOLocaleString();
-  const msSeparator = iso.indexOf('.');
-  iso = iso.subString(0, msSeparator);
-  return iso + date.getISOTimezoneOffset();
+  return moment(ms).format();
 }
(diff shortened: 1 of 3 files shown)


app/util/timeUtils.js
```diff
@@ -166,9 +166,5 @@ export function timeStr(dateTime) {
  * Epoch ms to ISO-8601/RFC3339 datetime str
  */
 export function epochToIso(ms) {
-  const date = new Date(ms);
-  let iso = date.toISOLocaleString();
-  const msSeparator = iso.indexOf('.');
-  iso = iso.subString(0, msSeparator);
-  return iso + date.getISOTimezoneOffset();
+  return moment(ms).format();
```
partisaani, line 169: In case we don't want to use moment:
```suggestion
const offsetMinutes = new Date().getTimezoneOffset();
const date = new Date(ms - offsetMinutes * 60000);
return date.toISOString();
```

partisaani (approved):
LGTM

app/util/timeUtils.js
vesameskanen: Hmm this gives right time but wrong timezone postfix. I guess there is a need to do string manipulation to replace 000z with +02:00 and also consider daylight saving. Time conversions are so laborius :(

partisaani: Note: toISOString() always appends the time string with Z which signals that the timestamp is UTC. Thus, reconversion from the string would cause a two-hour shift in wrong direction.

PR #5177, opened 2024-11-25
DT-6513 fix geojson zoom errors

vesameskanen:
When leaflet map changes by geometric bounds prop, digitransit map component will not know the new zoom until leaflet map has rendered the changed bounds. We must invoke a new render cycle where new zoom level is used for geoJson items. This is now done from itinerary page and near you page.

Another included bug fix: mobile prop from itinerary page was not passed to search config, and 'select from map' was therefore missing. Fixed now.

This PR also contains some minor code cleanup changes:
- removed dead scss variabes
- removed unused context, parameters and query sections
- removed unused eslint disable comments


app/component/map/NearYouMap.js
```diff
@@ -141,6 +141,7 @@ function NearYouMap(
     position,
     showWalkRoute,
     prioritizedStopsNearYou,
+    setMWTRef,
```
partisaani, line 144: Passing state mutating function is sometimes considered bad practice but since the ancestor function in `NearYouPage` looks like
```javascript
setMWTRef = ref => {
  this.MWTRef = ref;
};
```
without even having MWTRef as a field, I think this one _may_ be OK.

partisaani (approved):
LGTM with comment

partisaani:
Ticket identifier in title seems to refer to some totally unrelated issue

app/component/map/NearYouMap.js
vesameskanen: Right - i don't think stored ref is part of NearYouMap state, calling the callback will have no effect. This is quite typical use of React refs.

PR #5182, opened 2024-12-03
DT-6558 update non-transit leg progress

vesameskanen:
- New utilities for fast and accurate geometric computations in linear space
- If geolocation is off, estimate leg progress from time
- if geolocation is on, project  location to leg geometry and use traversed distance for estimates


partisaani (approved):
LGTM

PR #5188, opened 2024-12-05
Catch and handle errors in geoJson data fetching gracefully

vesameskanen:


app/store/GeoJsonStore.js
```diff
@@ -104,23 +107,28 @@ class GeoJsonStore extends Store {
...
+        const responses = await Promise.all(urlArr.map(u => getJson(u)));
+        const mapped = responses.map(r => {
+          if (metadata) {
+            MapJSON(r, metadata);
+          }
+          return styleFeatures(r);
+        });
+        for (let i = 1; i < mapped.length; i++) {
+          mapped[0].features = mapped[0].features.concat(mapped[i].features);
         }
```
partisaani: Could these loops be combined with something like
```javascript
  const features = [];
  const mapped = responses.map(r => {
    if (metadata) {
      MapJSON(r, metadata);
    }
    const styledFeature = styleFeatures(r);
    features.push(...styledFeature.features);
    return styledFeature;
  });

  mapped[0].features = features;
```

partisaani (approved):
LGTM with suggestion

app/store/GeoJsonStore.js
vesameskanen: ok, I will refactor

vesameskanen: logic is now much easier to understand

partisaani: Even better! 👏

partisaani (approved):

PR #5193, opened 2024-12-10
Call agency fixes

vesameskanen:
- Call agency status is now consistently decided everywhere using one single function
- Call button is not rendered if data lacks agency phone number


partisaani (approved):
LGTM

PR #5197, opened 2024-12-11
Better validation for alerts

vesameskanen:
- Alerts from past legs ignored
- Transfer alerts from past transfers ignored
- No transfer alert if user has reached the departure stop


partisaani (approved):
LGTM

PR #5202, opened 2024-12-16
Various navigation bug fixes and improvements

vesameskanen:
- Estimate transfer progress by user location only after transfer has started. Before that, use elapsed time.
- Transit leg state card now updates. Updating happens also when user is waiting at the stop.
- Transit leg state card shows route name and realtime in green
- Transit leg state card expires when it should
- Other cards do not expire because of comparison of undefined and time
- Warning logic about absence of realtime works as designed (considers trip start, not leg start time)


app/component/itinerary/navigator/NaviCardContainer.js
```diff
@@ -94,30 +96,29 @@ function NaviCardContainer(
...
       // Current active messages. Filter away expired messages.
       const previousValidMessages = legChanged
-        ? activeMessages.filter(m => m.expiresOn < time)
+        ? activeMessages.filter(m => !m.expiresOn || m.expiresOn > time)
         : activeMessages;
 
       // handle messages that are updated.
       const keptMessages = previousValidMessages.filter(
-        msg => !!incomingMessages.get(msg.id),
+        msg => !incomingMessages.get(msg.id),
```
partisaani, line 121: Is this 180° turn in the logic completely intentional?
image

partisaani (approved):
LGTM with question/comment

app/component/itinerary/navigator/NaviCardContainer.js
vesameskanen: Yes, I made a bug earlier. It is hard to remember if true keeps or removes :)

PR #5204, opened 2024-12-17
add via points to expanded view of the itinerary

tkalvas:
## Proposed Changes

  The via points are now displayed in the expanded itinerary view.

  The user can only add one via point.

## Pull Request Check List

## Review

  - Merge the pull request


app/component/itinerary/BicycleLeg.js
```diff
@@ -275,6 +275,13 @@ export default function BicycleLeg(
                     to={`/${PREFIX_STOPS}/${fromStop.gtfsId}`}
                   >
                     {origin}
+                    {leg.isViaPoint && (
+                      <Icon
+                        img="icon-icon_mapMarker-via"
+                        className="itinerary-mapmarker-icon"
+                        color={config.colors.primary}
```
vesameskanen: Primary color is the theme color (blue in HSL). Via marker is always black (icon has hardcoded black fill).  

Remove color prop from all Leg viapoints render calls.

vesameskanen (changes requested):
Some observations:
- Maybe in case of a via station, render only one via marker when transfer involves a stop change at the station
- If user accidentally sets the via stop to be the start or end stop of the journey, UI crashes
- When any of the shared libraries defined in yarn work spaces changes, library version must be updated:

https://github.com/HSLdevcom/digitransit-ui/blob/v3/digitransit-component/packages/digitransit-component-autosuggest-panel/package.json#L3

https://github.com/HSLdevcom/digitransit-ui/blob/v3/digitransit-component/packages/digitransit-component/package.json#L18

After updating the versions, remember to add updated yarn.lock into to PR.

Use normal version conventions ( breaking change -> major version grows).

PR #5210, opened 2024-12-20
DT-6620:Fix top cards width and another small adjustments

Antiik91:
## Proposed Changes

  - Top cards (main card and notifications) are now 100% width, with 8px margin on left and right sides
  - Fix transferProblem warning's styles
  - Top card change animation.
    - Plays when leg is changing, 2 second long animation where top card goes behind the top banner and comes back with updated info
    - Display the previous leg data for little longer than we need, so the animation goes smoothly.
  - Removed navistack top position calculation, since it's no longer needed.
  

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/itinerary/navigator/NaviCardContainer.js
+++ b/app/component/itinerary/navigator/NaviCardContainer.js
@@ -56,6 +56,7 @@ function NaviCardContainer(
     nextLeg,
     firstLeg,
     lastLeg,
+    previousLeg,
     isJourneyCompleted,
   },
   context,
@@ -179,7 +180,7 @@ function NaviCardContainer(
   }, [time]);
 
   // LegChange fires animation, we need to keep the old data until card goes ot of the view.
-  const l = legChanging ? lastLeg : currentLeg;
+  const l = legChanging ? previousLeg : currentLeg;
   const legType = handleLegChange(l, firstLeg, time);
 
   const containerTopPosition =
@@ -242,6 +243,7 @@ NaviCardContainer.propTypes = {
   nextLeg: legShape,
   firstLeg: legShape,
   lastLeg: legShape,
+  previousLeg: legShape,
   isJourneyCompleted: PropTypes.bool,
 
   /*
@@ -256,6 +258,7 @@ NaviCardContainer.defaultProps = {
   nextLeg: undefined,
   firstLeg: undefined,
   lastLeg: undefined,
+  previousLeg: undefined,
   isJourneyCompleted: false,
 };
 
--- a/app/component/itinerary/navigator/NaviCardContainer.js
+++ b/app/component/itinerary/navigator/NaviCardContainer.js
@@ -18,6 +18,7 @@ import { updateClient, getTopics } from '../ItineraryPageUtils';
 
 const TIME_AT_DESTINATION = 3; // * 10 seconds
 const TOPBAR_PADDING = 8; // pixels
+const HIDE_TOPCARD_DURATION = 2000; // milliseconds
 
 function addMessages(incominMessages, newMessages) {
   newMessages.forEach(m => {
@@ -44,6 +45,7 @@ const handleLegChange = (leg, firstLeg, time) => {
   }
   return legType;
 };
+
 function NaviCardContainer(
   {
     focusToLeg,
@@ -125,13 +127,14 @@ function NaviCardContainer(
         ...getAdditionalMessages(nextLeg, time, intl, config, messages),
       ]);
     }
+    let timeoutId;
     if (legChanged) {
       updateClient(topics, context);
       setCardExpanded(false);
       setLegChanging(true);
-      setTimeout(() => {
+      timeoutId = setTimeout(() => {
         setLegChanging(false);
-      }, 2000);
+      }, HIDE_TOPCARD_DURATION);
       if (currentLeg) {
         focusToLeg?.(currentLeg);
       }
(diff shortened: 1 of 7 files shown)


app/component/itinerary/navigator/NaviCardContainer.js
```diff
@@ -24,6 +24,26 @@ function addMessages(incominMessages, newMessages) {
...
+        legType = LEGTYPE.MOVE;
+      }
+    } else {
+      legType = LEGTYPE.TRANSIT;
+    }
+  } else {
+    legType = LEGTYPE.WAIT;
+  }
+  return legType;
+};
```
partisaani, line 47: Well refactored! A pure function such as this can live it's merry, optimized life outside the component.

app/component/itinerary/navigator/NaviCardContainer.js
```diff
@@ -73,14 +93,6 @@ function NaviCardContainer(
     updateClient(topics, context);
   }, []);
 
-  useEffect(() => {
-    if (cardRef.current) {
-      const contentHeight = cardRef.current.getBoundingClientRect();
-      // Navistack top position depending on main card height.
-      setTopPosition(contentHeight.bottom + TOPBAR_PADDING);
-    }
-  }, [currentLeg, cardExpanded]);
```
partisaani, line 82: Got rid of an extra useEffect 🙌

app/component/itinerary/navigator/NaviCardContainer.js
```diff
@@ -116,6 +128,10 @@ function NaviCardContainer(
     if (legChanged) {
       updateClient(topics, context);
       setCardExpanded(false);
+      setLegChanging(true);
+      setTimeout(() => {
+        setLegChanging(false);
+      }, 2000);
```
partisaani: Should we grab the timeout into a const and clear it in the useEffect's return function?

app/component/itinerary/navigator/navigator.scss
```diff
@@ -48,12 +48,19 @@
...
+  &.show-card {
+    animation: slideDownFromTop 2s ease-out forwards;
+  }
 }
 
 .navitop {
-  margin-bottom: 5px;
-  width: 92%;
-  margin-left: 8px;
+  margin: 0 8px 5px 8px;
```
partisaani: `8px` = `var(--space-s)`

app/component/itinerary/navigator/navigator.scss
```diff
@@ -62,6 +69,7 @@
   align-items: center;
   letter-spacing: -0.3px;
   box-shadow: 0 2px 4px 0 rgba(51, 51, 51, 0.2);
+  width: calc(100vw - 16px);
```
partisaani: What's the magic `16px`?

app/component/itinerary/navigator/navigator.scss
```diff
@@ -601,6 +612,16 @@
...
+@keyframes hideTopCard {
+  from {
+    transform: translateY(0);
+  }
+
+  to {
+    transform: translateY(-100%);
+  }
+}
+
```
partisaani, line 626: Probably same as `@keyframes slideUpToTop`

app/component/itinerary/navigator/navigator.scss
Antiik91: The magic is that we for some unkown reason for the moment  need to adjust the width with it's margins. Otherwise margin-right wont work with position: fixed. AFAIK

app/component/itinerary/navigator/navigator.scss
Antiik91: SlideUpToTop has this property
`
  to {
    transform: translateY(calc(-100vh - 100%));
  }
}`

And it did not respond well with this animation. I should double check that though!

partisaani: The animation which moves navi-card-container completely outside viewport uses `slideUpToTop`. If I interpret the desired outcome of `hideTopCard` animation correctly, the two _should_ use identical keyframes. However, the mileage may vary as `slideUpToTop` aims to move both the top card and any extensions or extra cards below it outside viewport.

Antiik91: Double checked, and some reason that proprety messes up the animation. So I suggest we keep my change as is.

app/component/itinerary/navigator/navigator.scss
partisaani: The some unknown reason should be figured out sooner than later

app/component/itinerary/navigator/navigator.scss
partisaani: Let's keep yours as well and then refactor later if possible

app/component/itinerary/navigator/navigator.scss
Antiik91: With new changes the magic is little less magical

partisaani (approved):

app/component/itinerary/navigator/NaviUtils.js
```diff
@@ -368,6 +368,7 @@ export const getItineraryAlerts = (
           content,
           id: `canceled-${legId}`,
           hideClose: true,
+          expiresOn: alert.effectiveEndDate,
         });
```
vesameskanen: expiresOn should be in epcoh not unix time.

app/component/itinerary/navigator/NaviCardContainer.js
```diff
@@ -161,32 +180,27 @@ function NaviCardContainer(
...
-      } else {
-        legType = LEGTYPE.MOVE;
-      }
-    } else {
-      legType = LEGTYPE.TRANSIT;
-    }
-  } else {
-    legType = LEGTYPE.WAIT;
-  }
+  // LegChange fires animation, we need to keep the old data until card goes ot of the view.
```
vesameskanen: typo ot -> out

vesameskanen (changes requested):

vesameskanen (approved):

PR #5213, opened 2024-12-27
Fix navigator translations and scheduled departure info

vesameskanen:
- Transit mode expressions now obey finnish grammar
 - Ferry pier added as an alternative to stop / station  expressions
 - Some translation ids were obscure or indicated incorrect context, fixed
 - Earlier design was not to show non-realtime departure warning if transit connection had not been moving 2 minutes. This lead to situation where no schedule information was shown for non-realtime departures from first stop. Fixed by changing the card from warning to info type.


partisaani:
LGTM

PR #5216, opened 2024-12-30
Proper rerouting logic and misc improvements

vesameskanen:
New itinerary search works now as follows:
- During transit leg, search from next stop after a decent threshold 
- During walking, search from user location if available, otherwise from the leg end
- During waiting/before journey start, search from next leg start
- Use actual geolocation, not 'POS' path, which seems to be slow/unreliable when geolocation is already enabled  

Other fixes:
- Alert generation and rendering code refactored
- Alert content now updates by newer versions
- Alerts can be closed and they won't reappear
- Static schedule warning is not shown before 2 minutes to first departure
- Wait leg focusing works correctly
- Semantic fix: current leg is not set before itinerary starts


partisaani (approved):
LGTM

PR #5219, opened 2025-01-02
Miscellaneous changes

vesameskanen:
- Remove unused sentry reporting code
- Let transfer alerts expire automatically when transfer time is over
- Refactor some styles
- Fix distance based check of reaching leg end 
- Rename and refactor


app/component/itinerary/navigator/NaviUtils.js
```diff
@@ -292,7 +293,9 @@ export const getTransitLegState = (leg, intl, messages, time) => {
     }
     content = (
       <div className="navi-info-content">
-        <FormattedMessage id="navileg-mode-schedule" />
+        <span className="navi-header">
```
Antiik91: `navi-header` is a bit ambiguous and suggests that these are part of some main header.

How about `notification-header`? It's a bit longer but in my opinnion it's more clear.

Antiik91 (changes requested):
otherwise LGTM

Antiik91 (approved):

PR #5225, opened 2025-01-13
DT-6652 More robust leg state management

partisaani:
## Proposed Changes
Freezes leg start and end when `Date.now() >= leg.start|end`. This should prevent legs in the past from reappearing as current

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/itinerary/navigator/hooks/useRealtimeLegs.js
+++ b/app/component/itinerary/navigator/hooks/useRealtimeLegs.js
@@ -3,7 +3,7 @@ import polyUtil from 'polyline-encoded';
 import { useCallback, useEffect, useState } from 'react';
 import { fetchQuery } from 'react-relay';
 import { GeodeticToEcef, GeodeticToEnu } from '../../../../util/geo-utils';
-import { LegMode, legTime } from '../../../../util/legUtils';
+import { legTime } from '../../../../util/legUtils';
 import { epochToIso } from '../../../../util/timeUtils';
 import { legQuery } from '../../queries/LegQuery';
 
@@ -96,8 +96,8 @@ function matchLegEnds(legs) {
   }
 }
 
-function getLegsOfInterest(realTimeLegs, now) {
-  if (!realTimeLegs?.length) {
+function getLegsOfInterest(legs, now) {
+  if (!legs?.length) {
     return {
       firstLeg: undefined,
       lastLeg: undefined,
@@ -105,23 +105,6 @@ function getLegsOfInterest(realTimeLegs, now) {
       nextLeg: undefined,
     };
   }
-  const legs = realTimeLegs.reduce((acc, curr, i, arr) => {
-    acc.push(curr);
-    const next = arr[i + 1];
-
-    // A wait leg is added, if next leg exists but it does not start when current ends
-    if (next && legTime(curr.end) !== legTime(next.start)) {
-      acc.push({
-        id: null,
-        legGeometry: { points: null },
-        mode: LegMode.Wait,
-        start: curr.end,
-        end: next.start,
-      });
-    }
-
-    return acc;
-  }, []);
 
   const currentLeg = legs.find(
     ({ start, end }) => legTime(start) <= now && legTime(end) >= now,
@@ -134,7 +117,7 @@ function getLegsOfInterest(realTimeLegs, now) {
     lastLeg: legs[legs.length - 1],
     previousLeg,
     currentLeg,
-    nextLeg: realTimeLegs.find(({ start }) => legTime(start) >= nextStart),
+    nextLeg: legs.find(({ start }) => legTime(start) >= nextStart),
   };
 }
 
@@ -212,18 +195,9 @@ const useRealtimeLegs = (relayEnvironment, initialLegs) => {
 
     setTimeAndRealTimeLegs(prev => {
       const rtLegs = prev.realTimeLegs.map(l => {
-        const freezeStart = legTime(l.start) <= now;
-        const freezeEnd = legTime(l.end) <= now;
         const rtLeg =
           l.legId && rtLegMap[l.legId] ? { ...rtLegMap[l.legId] } : null;
         if (rtLeg) {
-          if (l.freezeStart) {
-            delete rtLeg.start;
-          }
-          if (l.freezeEnd) {
-            delete rtLeg.end;
-          }
-
           return {
             ...l,
             ...rtLeg,
@@ -231,16 +205,19 @@ const useRealtimeLegs = (relayEnvironment, initialLegs) => {
               ...l.to,
               vehicleRentalStation: rtLeg.to.vehicleRentalStation,
             },
-            freezeStart,
-            freezeEnd,
           };
         }
 
-        return { ...l, freezeStart, freezeEnd };
+        return l;
       });
       // shift non-transit-legs to match possibly changed transit legs
       matchLegEnds(rtLegs, now);
-      return { ...prev, time: now, realTimeLegs: rtLegs };
+      const rtLegsWithFreezes = rtLegs.map(l => ({
+        ...l,
+        freezeStart: legTime(l.start) <= now,
+        freezeEnd: legTime(l.end) <= now,
+      }));
+      return { ...prev, time: now, realTimeLegs: rtLegsWithFreezes };
     });
   }, [queryAndMapRealtimeLegs]);
 
@@ -257,16 +234,15 @@ const useRealtimeLegs = (relayEnvironment, initialLegs) => {
   const { firstLeg, lastLeg, currentLeg, nextLeg, previousLeg } =
     getLegsOfInterest(realTimeLegs, time);
 
-  // return wait legs as undefined as they are not a global concept
   return {
     realTimeLegs,
     time,
     origin,
     firstLeg,
     lastLeg,
-    previousLeg: previousLeg?.mode === LegMode.Wait ? undefined : previousLeg,
-    currentLeg: currentLeg?.mode === LegMode.Wait ? undefined : currentLeg,
-    nextLeg: nextLeg?.mode === LegMode.Wait ? undefined : nextLeg,
+    previousLeg,
+    currentLeg,
+    nextLeg,
   };
 };
 
--- a/app/component/itinerary/navigator/hooks/useRealtimeLegs.js
+++ b/app/component/itinerary/navigator/hooks/useRealtimeLegs.js
@@ -194,29 +194,43 @@ const useRealtimeLegs = (relayEnvironment, initialLegs) => {
     );
 
     setTimeAndRealTimeLegs(prev => {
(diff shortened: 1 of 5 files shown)


app/component/itinerary/navigator/hooks/useRealtimeLegs.js
```diff
@@ -150,35 +138,43 @@ function getLegsOfInterest(
...
+        clonedLeg.freezeEnd = legTime(clonedLeg.end) <= time;
+        return clonedLeg;
+      }),
+    };
+  }
+  return {
+    time,
+    realTimeLegs: [],
+  };
+}
```
partisaani, line 146: Planar leg calculation was refactored here.

app/component/itinerary/navigator/hooks/useRealtimeLegs.js
```diff
@@ -197,61 +193,70 @@ const useRealtimeLegs = (relayEnvironment, initialLegs = []) => {
...
+      });
+      // shift non-transit-legs to match possibly changed transit legs
+      matchLegEnds(rtLegs, now);
+      return { ...prev, time: now, realTimeLegs: rtLegs };
     });
-    // shift non-transit-legs to match possibly changed transit legs
-    matchLegEnds(rtLegs);
-    setRealTimeLegs(rtLegs);
-  }, [planarLegs, queryAndMapRealtimeLegs]);
+  }, [time, realTimeLegs, origin, queryAndMapRealtimeLegs]);
```
partisaani: `time, realTimeLegs, origin` don't seem to be 100 % necessary. I can push a commit without them to keep things simpler.

app/component/itinerary/navigator/hooks/useRealtimeLegs.js
```diff
@@ -197,61 +193,70 @@ const useRealtimeLegs = (relayEnvironment, initialLegs = []) => {
...
-            vehicleRentalStation: rtLeg.to.vehicleRentalStation,
-          },
-        };
-      }
-      // copy leg times so that modification will not change original times
-      return { ...l, start: { ...l.start }, end: { ...l.end } };
+    setTimeAndRealTimeLegs(prev => {
+      const rtLegs = prev.realTimeLegs.map(l => {
+        const freezeStart = legTime(l.start) <= now;
+        const freezeEnd = legTime(l.end) <= now;
```
vesameskanen: I believe this should be done after merging new realtime to previous legs. We should make the freeze based on the very latest informatioin.

app/component/itinerary/navigator/hooks/useRealtimeLegs.js
```diff
@@ -123,24 +123,12 @@ function getLegsOfInterest(
     return acc;
   }, []);
```
vesameskanen, line 108: This is now totally unnecessary. Fake wait legs just add complexity. Use realTimeLegs array.

vesameskanen (changes requested):

app/component/itinerary/navigator/hooks/useRealtimeLegs.js
```diff
@@ -123,24 +123,12 @@ function getLegsOfInterest(
...
 
-  let currentLeg = legs.find(
-    ({ start, end }) => legTime(start) <= time && legTime(end) >= time,
+  const currentLeg = legs.find(
+    ({ start, end }) => legTime(start) <= now && legTime(end) >= now,
```
Antiik91, line 110: There is a lot refactoring going on from `time` variable to `now` variable. It was previously discouraged, but I guess it's necessary here? And do not cause sync problems with other places that uses the time step?

partisaani: > It was previously discouraged

I did not remember this! I refactored `time` as `now` in order to make it more obvious that at all times we are interested in making sure that timestamps in objects are compared against this very moment in time instead of just some random time

app/component/itinerary/navigator/hooks/useRealtimeLegs.js
```diff
@@ -212,35 +195,29 @@ const useRealtimeLegs = (relayEnvironment, initialLegs) => {
...
       });
       // shift non-transit-legs to match possibly changed transit legs
       matchLegEnds(rtLegs, now);
-      return { ...prev, time: now, realTimeLegs: rtLegs };
+      const rtLegsWithFreezes = rtLegs.map(l => ({
+        ...l,
+        freezeStart: legTime(l.start) <= now,
+        freezeEnd: legTime(l.end) <= now,
+      }));
+      return { ...prev, time: now, realTimeLegs: rtLegsWithFreezes };
```
vesameskanen: It seems that things still happen in wrong order. Transfer legs ares matched - before - freezing with current realtime has been applied.

I  suggest removing the latter array map and the first map loop can map 'l' to: 

```
let newLeg;
if (l.legId && rtLegMap[l.legId] )  {
  newLeg = {
            ...l,
            ...rtLeg,
            to: {
              ...l.to,
              vehicleRentalStation: rtLeg.to.vehicleRentalStation,
            },
            freezeStart,
            freezeEnd,
          };
} else {
  newLeg =  leg;
}
newleg.freezeStart = ...
newLeg.freezeEnd = ...

return newLeg
 ```

**Review by vesameskanen — CHANGES_REQUESTED** [review]

File: app/component/itinerary/navigator/hooks/useRealtimeLegs.js (hunk 6 above)
Comment (partisaani) [reply app/component/itinerary/navigator/hooks/useRealtimeLegs.js]: So basically, revert to the previous implementation without deletes?

```javascript
setTimeAndRealTimeLegs(prev => {
  const rtLegs = prev.realTimeLegs.map(l => {
    const freezeStart = legTime(l.start) <= now;
    const freezeEnd = legTime(l.end) <= now;
    const rtLeg = l.legId && rtLegMap[l.legId] ? { ...rtLegMap[l.legId] } : null;
    if (rtLeg) {
      return {
        ...l,
        ...rtLeg,
        to: {
          ...l.to,
          vehicleRentalStation: rtLeg.to.vehicleRentalStation,
        },
        freezeStart,
        freezeEnd,
      };
    }
    return { ...l, freezeStart, freezeEnd };
  });
  // shift non-transit-legs to match possibly changed transit legs
  matchLegEnds(rtLegs, now);
  return { ...prev, time: now, realTimeLegs: rtLegs };
});
```

vesameskanen: Once again a new version.  Now you freeze transit legs before applying latest realtime.

app/component/itinerary/navigator/hooks/useRealtimeLegs.js
```diff
@@ -194,29 +194,43 @@ const useRealtimeLegs = (relayEnvironment, initialLegs) => {
     );
 
     setTimeAndRealTimeLegs(prev => {
+      // Maps previous legs with fresh real time transit legs. If transit leg start is in the past according to previous state,
+      // the transit leg is marked as frozen to prevent the start from shifting in UI.
+      // rtLegMap does not contain legs that have ended in the past as they've been filtered before updates are queried
       const rtLegs = prev.realTimeLegs.map(l => {
+        const freezeStart = l.freezeStart || legTime(l.start) <= now;
         const rtLeg =
```
vesameskanen, line 201: I would remove legTime check here, and do new freezing only at the end.

app/component/itinerary/navigator/hooks/useRealtimeLegs.js
```diff
@@ -194,29 +194,43 @@ const useRealtimeLegs = (relayEnvironment, initialLegs) => {
...
+
           return {
             ...l,
-            ...rtLeg,
+            ...rtLeg, // delete above prevent this from overwriting a previous, frozen state
             to: {
               ...l.to,
               vehicleRentalStation: rtLeg.to.vehicleRentalStation,
             },
+            freezeStart,
```
vesameskanen: no need to assign freezeStart, it will be done again in last array mapping.

app/component/itinerary/navigator/hooks/useRealtimeLegs.js
```diff
@@ -194,29 +194,43 @@ const useRealtimeLegs = (relayEnvironment, initialLegs) => {
...
+
+      // Freezes any leg.start|end in the past
       const rtLegsWithFreezes = rtLegs.map(l => ({
         ...l,
-        freezeStart: legTime(l.start) <= now,
-        freezeEnd: legTime(l.end) <= now,
+        freezeStart: l.freezeStart || legTime(l.start) <= now,
+        freezeEnd: l.freezeEnd || legTime(l.end) <= now,
       }));
+
```
vesameskanen: Two issues:
- Why constantly copy complete legs? Just assign the two freezing values to existing legs.
-  I wonder if it would be better to remove check for previous freeze state and  just  compare legTime with now.  Frozen times won't move and now does not go back, so simpler expression is enough. Anyway,  I understand that checking existing state is faster but somehow this  looks a bit confusing.

app/component/itinerary/navigator/hooks/useRealtimeLegs.js
```diff
@@ -234,6 +248,7 @@ const useRealtimeLegs = (relayEnvironment, initialLegs) => {
   const { firstLeg, lastLeg, currentLeg, nextLeg, previousLeg } =
     getLegsOfInterest(realTimeLegs, time);
 
+  // return wait legs as undefined as they are not a global concept
```
vesameskanen: I find this comment confusing as there are no wait legs.

vesameskanen (changes requested):
I believe this implementation works. I still added a couple of comments, as some values seem to be evaluated multiple times.

app/component/itinerary/navigator/hooks/useRealtimeLegs.js
partisaani: Leftover, removing.

app/component/itinerary/navigator/hooks/useRealtimeLegs.js
partisaani: Leg copying is quite prominent as our eslint config complains if I try to directly assign properties inside `Array.prototype.map`
image

`l.freezeEnd|freezeStart` shortcut was added, as you said, because it's a bit faster but I do admit it adds some unnecessary clutter

app/component/itinerary/navigator/hooks/useRealtimeLegs.js
partisaani: Fixed in [defdc2b](https://github.com/HSLdevcom/digitransit-ui/pull/5225/commits/defdc2b5cc0db02504c0acde0365139efee028de)

app/component/itinerary/navigator/hooks/useRealtimeLegs.js
partisaani: Fixed in [defdc2b](https://github.com/HSLdevcom/digitransit-ui/pull/5225/commits/defdc2b5cc0db02504c0acde0365139efee028de)

PR #5227, opened 2025-01-14
DT-6447 try restarting frozen location watch

vesameskanen:
Normally geolocation changes slightly all the time. Try restarting location watch in navigator once after 30 seconds if position remains the same.

This PR also contains some minor refactoring and a regression bug fix for walk path projections.


partisaani (approved):
LGTM

PR #5229, opened 2025-01-15
DT-6675 stop stop page navigation fight

vesameskanen:
- Refactor  stop page walk routing
-  Allow user to navigate the map after focusing on the  walk route


partisaani (approved):
LGTM

PR #5231, opened 2025-01-16
DT-6654: Combine walk and wait cards

Antiik91:
## Proposed Changes

  - Walk to a stop now has walk instructions, as well as information about the vehicle you are boarding.
  - Expand icon has now two seperate sizes depending on where it is used. (16px * 16px & 24px * 24px)

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/itinerary/navigator/NaviCardExtension.js
+++ b/app/component/itinerary/navigator/NaviCardExtension.js
@@ -15,7 +15,6 @@ import ZoneIcon from '../../ZoneIcon';
 import { legShape, configShape } from '../../../util/shapes';
 import { getDestinationProperties, LEGTYPE, withRealTime } from './NaviUtils';
 import { getRouteMode } from '../../../util/modeUtils';
-import RouteNumber from '../../RouteNumber';
 import RouteNumberContainer from '../../RouteNumberContainer';
 import NaviBoardingInfo from './NaviBoardingInfo';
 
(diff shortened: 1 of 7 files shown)


app/component/itinerary/navigator/NaviCardExtension.js
```diff
@@ -129,18 +131,66 @@ const NaviCardExtension = ({ legType, leg, nextLeg }, { config }) => {
       </div>
     );
   }
+  if (legType === LEGTYPE.MOVE && nextLeg) {
+    const { headsign, route, start } = nextLeg;
```
vesameskanen: This crashes or does something funny if nextLeg is not transit. Please change to  nextLeg?.transitLeg

app/component/itinerary/navigator/NaviCardExtension.js
```diff
@@ -129,18 +131,66 @@ const NaviCardExtension = ({ legType, leg, nextLeg }, { config }) => {
...
+      <div className="extension">
+        {stopInformation()}
+        <div className="extension-divider" />
+        <div className="wait-leg">
+          <div className="route-info">
+            <div className="expand">
+              <Icon img="navi-expand" className="icon-expand-small" />
+              <div className="wait-duration">
+                <FormattedMessage
+                  id="navileg-arrive-at"
```
vesameskanen: Just a minor issue, but obviously translation id `navileg-arrive-at` is misleading in thsi context.

vesameskanen (changes requested):
Maybe we could use a bit more modular approach here, instead of copypasting jsx snippets and styles around. Route information appears identical both in MOVE extension and WAIT card. Transit extension also shows similar route icon with headsign.

Styling of such shared elements should also happen in a shared manner, instead of repeating the applied style class as a new nested sub style.

https://files.slack.com/files-pri/T03HA371Q-F088Q5ZKCQ7/image.png?is_viewed=1

PR #5234, opened 2025-01-22
DT-6689 fix fake journey end

vesameskanen:
- length of the remaining part of current leg is passed as a prop, useful for many purposes 
- journey end check uses the actual distance along the  geometry of the last leg
- closing message bar updates navigator view immediately


partisaani (approved):
LGTM

PR #5242, opened 2025-01-28
DT-6586 Manual start for Navigator journey

partisaani:
## Proposed Changes
- Refactored `<button>` element from `NaviCardContainer` into `NaviCard` as the element is tightly coupled with the card.
- Adds a component which houses a functionality for user to initialize their Navigator journey ahead of its schedule with the following logic: if the first leg is...
 1. a walking leg, the walk leg is set to begin immediately, and its start and end times are shifted. A gap appears between the first and second legs and thus a wait is added between them.
 2. a transit leg, a `forceStart` flag is added to the transit leg object which is then used as a signal to instruct the user to wait for their transit

## Pull Request Check List

## Review

  - Merge the pull request


app/component/itinerary/navigator/NaviCard.js
```diff
@@ -27,104 +27,103 @@ const iconMap = {
...
 
-    mainCardContent = (
-      <>
-        <Icon img={iconName} className="mode" color={iconColor} />
-        <div className={`instructions ${cardExpanded ? 'expanded' : ''}`}>
-          <NaviInstructions
+  return (
+    <button
+      type="button"
+      className={`navi-top-card ${cardExpanded ? 'expanded' : ''}`}
```
sharhio, line 86: Might be nicer to use the classnames library for this (and other similar strings).
Like this:
className={cx('navi-top-card', {expanded: cardExpanded})}

partisaani: So far I've used cx for cases where there's two or classes which are assigned based on condition. In single cases such as this I find it doesn't really clarify things, and the function call overhead cannot be easily justified.

vesameskanen (commented):
I noticed that when itinerary has a walk leg at start and you start the trip early, walk instructions appear immediately, then slide up and return again. This is undesirable. Is that something top be fixed in a new PR or should it be fixed within this PR?

partisaani:
> I noticed that when itinerary has a walk leg at start and you start the trip early, walk instructions appear immediately, then slide up and return again. This is undesirable. Is that something to be fixed in a new PR or should it be fixed within this PR?

This issue also occurs whenever cards are swapped during a trip, so I feel like it may deserve a ticket/fix of its own.

PR #5247, opened 2025-01-31
DT-5933 Show settings changed only when there are changes

sharhio:
## Proposed Changes

  - Show settings changed only when the number is above 0.
  - Styles fixed
  - Aria-label added
  - check all setting arrays before counting

## Pull Request Check List

## Review

  - Merge the pull request


app/component/itinerary/CustomizeSearch.js
```diff
@@ -275,7 +278,11 @@ class CustomizeSearch extends React.Component {
               </div>
             </div>
           )}
-          <RestoreDefaultSettingSection config={config} />
+          <div className="settings-section background">
+            <div className="settings-option-container">
+              <RestoreDefaultSettingSection config={config} />
+            </div>
+          </div>
```
vesameskanen, line 285: Could you please refactor these double div sections away from CustomizeSearch.js? One div can have many style classes.

vesameskanen (changes requested):

app/component/itinerary/CustomizeSearch.js
sharhio: Of course! There are a lot of these double divs in this class, should they all be removed while I'm at it?

sharhio: There are different padding settings for these divs. When they are separated as they are now, both padding settings are applied, but if they are combined one will override the other, which causes there to be less padding than before. Should the css settings be changed to make it work in the same way as before with less divs or should the divs stay separate?

partisaani (approved):
LGTM

PR #5260, opened 2025-02-10
DT-5933 Reset to default settings, count citybike settings correctly

sharhio:
## Proposed Changes

  - Citybike settings were incorrectly counted as changed when they are not shown on the ui. Only count active networks.

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/util/planParamUtil.js
+++ b/app/util/planParamUtil.js
@@ -1,6 +1,10 @@
 import moment from 'moment';
 import isEqual from 'lodash/isEqual';
-import { getTransitModes, isTransportModeAvailable } from './modeUtils';
+import {
+  getTransitModes,
+  isTransportModeAvailable,
+  networkIsActive,
+} from './modeUtils';
 import { otpToLocation, getIntermediatePlaces } from './otpStrings';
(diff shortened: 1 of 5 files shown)


app/util/planParamUtil.js
```diff
@@ -62,13 +67,45 @@ export function getDefaultSettings(config) {
...
+ */
+export function validateCitybikeSettings(
+  config,
+  executeAction,
+  defaultSettings,
+  customizedSettings,
+) {
+  if (
+    customizedSettings.allowedBikeRentalNetworks?.length > 0 &&
+    !useCitybikes(config.vehicleRental?.networks, config)
```
vesameskanen: This seems wrong. One selected network might be hidden but some networks are still active, and the logic does nothing.

I suggest filtering the  customizedSettings.allowedBikeRentalNetworks with active networks and using that result to count the changes. It is also better not to change the settings before the user actually asks for a reset.

vesameskanen (changes requested):

app/util/planParamUtil.js
sharhio: My bad, I fixed the logic to check each network. It was decided by the designers that the inactive networks should be automatically reset.

PR #5264, opened 2025-02-12
fix(settings): sort funicular before carpool

hbruch:
## Proposed Changes

  - in settings dialog sort funicular before carpool


hbruch:
Sorry, wrong base repo

PR #5276, opened 2025-02-19
DT-6663 update transfer time bubbles on map

vesameskanen:
Navigation view uses a callback function to update the itinerary page's stored itinerary.  When itinerary changes, itinerary map will re-render, showing possibly changed transfer times.

### DIFF
--- a/app/component/itinerary/ItineraryPage.js
+++ b/app/component/itinerary/ItineraryPage.js
@@ -696,6 +696,16 @@ export default function ItineraryPage(props, context) {
     setStoredItinerary(itineraryWithParams);
   };
 
+  const updateStoredItinerary = legs => {
+    setStoredItinerary({
+      ...storedItinerary,
+      itinerary: {
+        ...storedItinerary.itinerary,
+        legs,
+      },
+    });
+  };
+
   // save url-defined location to old searches
   function saveUrlSearch(endpoint) {
     const parts = endpoint.split('::'); // label::lat,lon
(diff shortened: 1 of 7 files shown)


app/component/itinerary/ItineraryPage.js
```diff
@@ -696,6 +696,16 @@ export default function ItineraryPage(props, context) {
...
+  const updateStoredItinerary = legs => {
+    setStoredItinerary({
+      ...storedItinerary,
+      itinerary: {
+        ...storedItinerary.itinerary,
+        legs,
+      },
+    });
+  };
+
```
partisaani, line 708: Should this also update the itinerary we've stored in local storage?

app/component/itinerary/navigator/hooks/useRealtimeLegs.js
```diff
@@ -220,7 +225,7 @@ const useRealtimeLegs = (relayEnvironment, initialLegs, position) => {
         l.freezeStart = l.freezeStart || legTime(l.start) <= now;
         l.freezeEnd = l.freezeEnd || legTime(l.end) <= now;
       });
-
+      updateLegs(rtLegs);
```
partisaani: While not probable, it is possible for `updateLegs` to be undefined. Should be surrounded with if

partisaani (changes requested):

app/component/itinerary/ItineraryPage.js
vesameskanen: I don't think so. Continuous writing to localStorage feels bad and should be avoided unless there is a good reason.

app/component/itinerary/navigator/hooks/useRealtimeLegs.js
vesameskanen: UpdateLegs is defined as a required prop. How can it be undefined? If it is, code should crash, right?

partisaani: That's true, but such safeguard is missing from the hook's point of view.

partisaani (approved):

PR #5283, opened 2025-02-28
DT-6586 Improve force start UX

partisaani:
## Proposed Changes
- Refactored `NaviStarter` higher in hierarchy
  - It isn't a leg card so it shouldn't be under `NaviCardContainer`
  - Less prop drilling
  - Significant reduction in its render condition complexity
- Refactored `NaviStarter` to be self-responsible on its visibility
  - This was easier to achieve inside the component and clutters `NaviContainer` less
 - Store information when an itinerary was force started along the itinerary's local storage object
 - Added simple `loading` flag in `useRealTimeLegs` hook to let its consumer(s) know whether all relevant data is available after initialization
   - Without the flag a force started itinerary is initially rendered with no current leg causing incorrect animation sequence to be fired.

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/itinerary/ItineraryPage.js
+++ b/app/component/itinerary/ItineraryPage.js
@@ -1238,6 +1238,7 @@ export default function ItineraryPage(props, context) {
             mapLayerRef={mapLayerRef}
             isNavigatorIntroDismissed={isNavigatorIntroDismissed}
             updateLegs={updateStoredItinerary}
+            forceStartAt={storedItinerary?.params?.forceStartAt || undefined}
           />
         </>
       );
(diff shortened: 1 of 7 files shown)


app/component/itinerary/ItineraryPage.js
```diff
@@ -1238,6 +1238,7 @@ export default function ItineraryPage(props, context) {
             mapLayerRef={mapLayerRef}
             isNavigatorIntroDismissed={isNavigatorIntroDismissed}
             updateLegs={updateStoredItinerary}
+            forceStartAt={storedItinerary?.params?.forceStartAt || undefined}
           />
```
vesameskanen: `|| undefined` seems unnecessary as that's what is evaluated when forcestart is not set

vesameskanen (changes requested):

PR #5285, opened 2025-03-04
Fix circle line misalignment

VillePihlava:
image

### DIFF
--- a/app/component/itinerary/itinerary.scss
+++ b/app/component/itinerary/itinerary.scss
@@ -2112,7 +2112,7 @@ $itinerary-tab-switch-height: 48px;
     .itinerary-leg-action {
       border-top: 1px solid #ddd;
       padding: 1.2em 0;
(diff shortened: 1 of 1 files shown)


app/component/itinerary/itinerary.scss
```diff
@@ -2112,7 +2112,7 @@ $itinerary-tab-switch-height: 48px;
     .itinerary-leg-action {
       border-top: 1px solid #ddd;
       padding: 1.2em 0;
-      margin-bottom: 1em;
+      margin-bottom: 0.1em;
```
partisaani, line 2115: Removing the margin altogether might also work since no class in this file is instructing an element to have `margin-bottom: 0.1em;`

VillePihlava: Yeah probably, this change was to set the margin to look similar to the old one, which was 1px

PR #5293, opened 2025-03-10
DT-6624 Notification card styles

sharhio:
## Proposed Changes

  - Margins, transfer-info and route-number style adjustments.

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/itinerary/navigator/navigator.scss
+++ b/app/component/itinerary/navigator/navigator.scss
@@ -86,6 +86,8 @@
 
 .navi-transfer-container {
   display: inline-flex;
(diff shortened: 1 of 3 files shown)


vesameskanen (commented):
Why do you use `isMounted` variable? Looks really strange to me.

sharhio:
The variable stops the "Can't perform a React state update on an unmounted component" warning. Because the function is async the component may already have been unmounted by the time the fetch finishes, which causes the warning. There is no danger of memory leak in this instance, I just don't like the clutter on console.

PR #5301, opened 2025-03-13
DT-6711 multilanguage ticketlink config

sharhio:
## Proposed Changes

  - Ticketlink config now has different language versions.

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen (commented):
I wonder if we should make a little util func 'localizedUrl(object, lang)' which tries to get a language specific link but defaults to the first entry if such translation is not found? I suspect that often all links will point to the same address. Also, some planners might have even more languages than we do.

PR #5305, opened 2025-03-18
DT-6659 Navigator: ask permission for geolocation

sharhio:
## Proposed Changes

  - New information modal about geolocation.
  - Navigation asks for geolocation at intro and itinerary before moving to the navigation view.

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen (changes requested):
Works usually fine, but I found one case where entry to navigation breaks:
- Open a journey planner which excludes your geolocation (e.g. Tampere planner from Helsinki). Make sure geolocation is allowed. Index page will check your geolocation but as you are not in the service area, it will not be automatically set to origin.
- Make an itinerary search
- Try to enter navigation view via the intro modal. Start button will not work. You can return back to itinerary detail view, though.

PR #5308, opened 2025-03-21
DT-6793 navigation fixes and improvements

vesameskanen:
- Force started navigation works when there afre more than one legs before transit
- More accurate estimate for reaching a stop
- Initial logic for validating a leg change by vehicle position
- Allow bike in subway for matka and hsl


partisaani (approved):
LGTM

PR #5316, opened 2025-03-26
DT-6800 future route suggestion bug fixes

vesameskanen:
- Ugly GTFS identification is removed from the label
- All search utilities use consistently localadmin, not locality. Locality is not even defined for all geocoding items.
Screenshot 2025-03-26 at 8 20 58
Screenshot 2025-03-26 at 8 29 06


partisaani (approved):
LGTM, but I am not 100 % familiar with the issue, so another review is probably for the best

PR #5318, opened 2025-03-26
DT-6586 Force start adjustments, bump arrival time buffer

partisaani:
## Proposed Changes
- Fixes issue where a wait card was not shown if a force started navigator journey starts with transit
- Increased the arrival time buffer from 1 minute to 5 minutes
- ~Every navigator journey is now force started 2 minutes before the actual itinerary is scheduled to start~

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/itinerary/navigator/NaviContainer.js
+++ b/app/component/itinerary/navigator/NaviContainer.js
@@ -19,6 +19,7 @@ import { addAnalyticsEvent } from '../../../util/analyticsUtils';
 const ADDITIONAL_ARRIVAL_TIME = 300000; // 5 min in ms
 const LEGLOG = true;
 const TOPBAR_PADDING = 8; // pixels
+const START_BUFFER = 120000; // 2 min in ms
 
 function NaviContainer(
   {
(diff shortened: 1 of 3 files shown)


app/component/itinerary/navigator/NaviContainer.js
```diff
@@ -101,13 +106,15 @@ function NaviContainer(
 
   if (LEGLOG) {
     // eslint-disable-next-line
-    console.log(...summaryString(realTimeLegs, time, previousLeg, currentLeg, nextLeg));
+    console.log(
+      ...summaryString(realTimeLegs, time, previousLeg, currentLeg, nextLeg),
+    );
```
partisaani: No actual change here, prettier just doing its thing

vesameskanen (commented):
Forced start might need some further tuning. If itinerary starts after 2min10s, user sees a quick flash of start now modal which then closes automatically. Not good.

We need to consider if navigation should always start a bit in advance. This is perhaps good because first walk does not include transfer margin time. 

If extra margin is wanted,  we need two step transition in the start.  If HSL does not want it,  forced start activation shoudl be considered only once when navi opens.

partisaani:
We can leave that out for now and just fix the issue for missing wait card if transit leg initiates the journey.

vesameskanen (approved):

PR #5321, opened 2025-03-28
Option to simulate transfer problems

vesameskanen:
Maximal biking speed activates simulated transfer problem. First transit leg gets forced delays in steps: 1.5 min- 3 min - 5 min - 3 min - 1.5 min - back to normal. 
 
This PR also starts using  the actual transfer slack from the itinerary search settings instead of a hard coded value.

Transit leg state card had invalid translation id, fixed now.


partisaani (approved):
LGTM

PR #5323, opened 2025-03-29
DT-6586 Responsibility split for useRealtimeLegs

partisaani:
## Proposed Changes
Make codebase healthier by
- moving util functions into separate file
- extracted state management into a separate sub-hook

The changes were made in order to prepare to pay the substantial amount of accumulated technical debt and lack of unit tests

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen (approved):
I would avoid use of folders which contain only a single file (hooks/utils) unless there is a reason to expect that it will contain many files in the future. Anyway, this is not an important issue.

PR #5332, opened 2025-04-07
DT-6557 Fix alert icon on itinerary

sharhio:
## Proposed Changes

  - Alert icon was missing on itinery list.
  - Leg alert validity check fixed.
  - Also simplified the logic and renamed for clarity (function deals with lists, not single alert).

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/util/alertUtils.js
+++ b/app/util/alertUtils.js
@@ -260,6 +260,16 @@ export const getActiveAlertSeverityLevel = (alerts, referenceUnixTime) => {
   );
 };
 
+export const isAlertActive = (referenceUnixTime, alert) => {
+  if (!alert) {
+    return false;
+  }
+  return (
+    isAlertValid(alert, referenceUnixTime) &&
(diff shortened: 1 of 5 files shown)


app/util/alertUtils.js
```diff
@@ -280,17 +290,9 @@ export const isAlertActive = (
...
 
-  if (alerts.length === 0) {
-    return false;
-  }
-
-  const filteredAlerts = alerts.filter(alert =>
-    isAlertValid(alert, referenceUnixTime),
+  return (
+    alerts.filter(alert => isAlertActive(referenceUnixTime, alert)).length > 0
   );
```
vesameskanen: This is really a minor issue but I would prefer alerts.find instead of filter which unnecessarily processes the whole array.

vesameskanen (changes requested):

vesameskanen:
image

Maybe this needs attention as well.

PR #5337, opened 2025-04-10
DT-6624 Less mechanical animations

sharhio:
## Proposed Changes

  - Cards sliding up and down use a cubic-bezier easing function.
  - Added padding to avoid main card edge showing when it should be hidden.
  - Simplified the logic slightly (no need for a different slide up animation when a journey ends).
  - Fixed a glitch where the cards start sliding down, then jump back up and then slide down again.
  - Main card and info stack now move individually. (Info stack goes up with all info cards to remove the disappearing cards nicely.)

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/itinerary/navigator/NaviStack.js
+++ b/app/component/itinerary/navigator/NaviStack.js
@@ -3,9 +3,9 @@ import PropTypes from 'prop-types';
 import cx from 'classnames';
 import NaviMessage from './NaviMessage';
 
-const NaviStack = ({ messages, handleRemove }) => {
+const NaviStack = ({ messages, handleRemove, cardAnimation }) => {
   return (
-    <div className={cx('info-stack', 'slide-in')}>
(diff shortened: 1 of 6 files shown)


app/component/itinerary/navigator/NaviStack.js
```diff
@@ -29,6 +30,7 @@ NaviStack.propTypes = {
     }),
   ).isRequired,
   handleRemove: PropTypes.func.isRequired,
+  cardAnimation: PropTypes.string.isRequired,
```
partisaani, line 33: A bit nitpicky, but I feel like this should be renamed to something less ambiguous i.e. `cardAnimationClass` to clarify how it's used in the component.

partisaani (changes requested):

vesameskanen (approved):

PR #5345, opened 2025-04-17
Dt 6844 migrate from relay container to hooks

hjvuor:
## Proposed Changes

  - Refactored ItineraryList to use the relay useFragment hook instead of createFragmentContainer
  - ~~Modifyed ItineraryList.test.js to wrap the rendered contents inside a mock relay environment~~
  - ~~added relay-test-utils to dev dependencies~~
  - UPDATE
  - Refactored Itinerary to also use the hook
  - To pass tests, mocked useFragment in init.js
       - This way tests pass without any modification
       - The mocked hook assumes that the component is passed complete data as a prop
       - No longer need for relay-test-utils
 While mocking the useFragment hook, I also noticed that relay was incorrectly imported in init.js and the fixed import broke the test for FuzzyTripLink. After removing the incorrect mock the old tests passed.

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/itinerary/ItineraryList.js
+++ b/app/component/itinerary/ItineraryList.js
@@ -1,6 +1,6 @@
 import PropTypes from 'prop-types';
 import React from 'react';
-import { createFragmentContainer, graphql } from 'react-relay';
+import { useFragment, graphql } from 'react-relay';
 import { FormattedMessage } from 'react-intl';
 import cx from 'classnames';
 import { matchShape } from 'found';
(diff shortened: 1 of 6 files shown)


app/component/itinerary/ItineraryList.js
```diff
@@ -47,12 +47,34 @@ function ItineraryList(
   const { location } = context.match;
   const { hash } = context.match.params;
 
-  const co2s = planEdges
+  const planEdgesData = useFragment(
```
vesameskanen: Please use the name planEdges. Most code variables contain data, so the postfix Data is really of no value.

yarn.lock
```diff
@@ -1867,7 +1867,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@babel/runtime@npm:7.27.0":
+"@babel/runtime@npm:7.27.0, @babel/runtime@npm:^7.25.0":
   version: 7.27.0
```
vesameskanen: I wonder why we need two versions of babel runtime? The changes create dependency to an older version. Please check if this can be avoided.

vesameskanen (commented):
Looks otherwise good, but  please check the error catching issue Joel mentioned in the daily. Maybe error handling can be also tested by closing the network connection while itinerary detail view is updating.

partisaani (approved):
Looks good all in all. In future iterations I would like to see the GraphQL query strings extracted into their own files but it's out of the scope of this PR and ticket

edit: as a general note, it is advisable to have a prefix describing the commit i.e. fix, chore or feat

hjvuor:
> Looks otherwise good, but please check the error catching issue Joel mentioned in the daily. Maybe error handling can be also tested by closing the network connection while itinerary detail view is updating.

Also discussed this today with Joel and seems like he confused error boundaries with suspense which is now supported with the hooks. In the future that could be possibly looked into for improved responsibility.

vesameskanen (approved):

PR #5346, opened 2025-04-22
DT-6773 option to adjust leg times by geolocation

vesameskanen:
- Some refactoring in useRealtimeLegs hook
- Remove unreliable custom code which ends walk legs by geolocation
- Refactor transit prob simulation away from component to utils
- Don't freeze all initial legs in force starting, because early transit leg will overlap them
- Don't move started initial legs if first transit leg is delayed, instead let traveller to wait at the first stop


app/component/itinerary/navigator/hooks/utils/realtimeLegUtils.js
```diff
@@ -154,46 +157,199 @@ function getLegsOfInterest(legs, now) {
...
+        estimated: {
+          ...leg.end.estimated,
+          time: epochToIso(legTime(leg.end) + gap),
+          delay: gap,
+        },
+      };
+    } else {
+      leg.end.scheduledTime = epochToIso(legTime(leg.end) + gap);
+    }
+    // leg.realtimeState = 'UPDATED';
```
partisaani, line 186: Should this be uncommented?

app/component/itinerary/navigator/hooks/useRealtimeLegs.js
```diff
@@ -94,32 +99,18 @@ const useRealtimeLegs = (
...
+
       // Shift unfrozen, non-transit-legs to match possibly changed transit legs
       matchLegEnds(newRtLegs, now);
 
-      shiftLegsByGeolocation(newRtLegs, time, vehicles, position, origin);
+      if (GEOLOCATED_LEGS) {
+        shiftLegsByGeolocation(newRtLegs, now, vehicles, position, origin);
+        // transit legs may have changed, shift again
+        matchLegEnds(newRtLegs, now);
+      }
```
partisaani, line 113: Would this be cleaner - and function as intended - as: 

```
if (GEOLOCATED_LEGS) {
  shiftLegsByGeolocation(newRtLegs, now, vehicles, position, origin);
}

// Shift unfrozen, non-transit-legs to match possibly changed transit legs
matchLegEnds(newRtLegs, now);
```

to prevent calling `matchLegEnds` twice in a rapid succession?

app/component/itinerary/navigator/hooks/useRealtimeLegs.js
```diff
@@ -24,8 +27,9 @@ const useRealtimeLegs = (
   forceStartAt,
   simulateTransferProblem,
 ) => {
-  const [{ origin, time, realTimeLegs }, setTimeAndRealTimeLegs] =
-    useInitialLegState(initialLegs);
+  const [{ origin, time, realTimeLegs }, setRealTimeLegs] =
+    useInitialLegState(initialLegs); // time = time published to hook users
```
partisaani: At this point I would probably rename the `useInitialLegState` as `useLegState` as it has actually never been just about setting the initial state, but it has also been responsible for tracking its state

partisaani (changes requested):

app/component/itinerary/navigator/hooks/useRealtimeLegs.js
vesameskanen: I believe matching must be done twice.  Geolocation logic needs a consistent leg chain with no overlapping parts. After it has done its job, non-transit legs might need shifting again. shiftLegsByGeolocation leaves that job deliberaterly to useRealtimeLegs.

app/component/itinerary/navigator/hooks/utils/realtimeLegUtils.js
vesameskanen: Hard question. realtimeState is used for showing leg time in green color. Is time change derived from vehicle position realtime ? It is not fully static, but neither an estimate, just a minimal change to avoid card change too early/late.

vesameskanen: maybe we will leave the decision to Katja and Teijo.

partisaani (approved):

PR #5348, opened 2025-04-23
DT-6784 Replacement bus services

sharhio:
## Proposed Changes

  - Support for replacement bus service types
  - Updated the HSL link icon (and related styles)
  
Testing requires data to have a route with route_type 714 (and preferably some route_desc and route_url).

## Pull Request Check List

## Review

  - Merge the pull request


app/component/itinerary/ItineraryDetails.js
```diff
@@ -262,6 +264,24 @@ class ItineraryDetails extends React.Component {
       }
     }
 
+    itinerary.legs.forEach(leg => {
+      const { route } = leg;
+      if (
+        route &&
+        route.desc?.length > 0 &&
```
vesameskanen: This might work and is simpler:
if (route?.desc?.length &&

app/component/itinerary/RouteDisclaimer.js
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import { FormattedMessage } from 'react-intl';
+import PropTypes from 'prop-types';
+import Icon from '../Icon';
+
+export default function RouteDisclaimer({
+  textId,
```
vesameskanen, line 7: Do you expect that textId and values will be needed at some point?

vesameskanen (changes requested):
Some notes:
- Vehicle marker uses bus color (should be replacement color)
- Search does not understand replacement mode but shows standard bus icon
- HSL svg sheet seems to contain invalid xml

app/component/itinerary/ItineraryDetails.js
sharhio: You're correct. The first route check was originally because of the getRouteMode function and the desc check came later, so forgot to remove.

app/component/itinerary/RouteDisclaimer.js
sharhio: I figured it might be useful to support it from the get go as we usually use textIds. It's not entirely decided yet what the info in this box will be anyway, and it might end up being a static notification, which would have a textId. This class is also very similar to the FareDisclaimer which supports both. I'm actually not entirely sure if we really need two classes that do almost exactly the same thing. The only difference is what the link looks like. These infoboxes really need some unification.

partisaani (approved):
LGTM

sharhio:
- Fixed vehicle marker color
- Added logo to search and updated dependency versions
- Fixed the svg (duplicated field)

app/component/itinerary/ItineraryDetails.js
```diff
@@ -262,6 +264,24 @@ class ItineraryDetails extends React.Component {
       }
     }
 
+    if (config.showRouteDisclaimer) {
+      itinerary.legs.forEach(leg => {
+        const { route } = leg;
+        if (
+          route?.desc?.length > 0 &&
```
vesameskanen: This sometimes evaluates as `undefined > 0`, which feels uncomfortable. How about dropping the comparison?
`route?.desc?.length` would skip undefined and value 0.

vesameskanen (changes requested):

vesameskanen (approved):

PR #5352, opened 2025-04-25
Add `maxQueryIterations` to `getPlanParams`

VillePihlava:
The old way of setting this for `PLANTYPE.CARTRANSIT` felt wrong, so I fixed it.


vesameskanen (approved):
Good change.

PR #5359, opened 2025-04-29
DT-6877 - Fixes for external ferries and call agencies

VillePihlava:
## Proposed Changes

  - Fixes for coloring in route and stop departure views for external ferries and call agencies
  - Fix padding of route number bar (at least for call agencies)
  - Lots of call agency refactors and improvements
  - Translation fixes for call agencies


vesameskanen (changes requested):
Looks good. One note though: it seems that pickupType (and possibly some other leg fields) are no longer needed in itinerary page. Please clean up PlanConnection and other queries which contain it.

PR #5396, opened 2025-05-27
DT-6904 Remove support for server-side rendering

optionsome:
## Proposed Changes

  - Support for SSR is removed
    - Some code etc. related to SSR is removed but there is probably more
  - `supercluster` dependency version had to be downgraded for unit tests to run
     - This is likely due to map code being now always included instead of just in client. When we have better support for ESM, the version can be upgraded again
  - feedbackly code is removed as it was unused and complicated context
  - Removed a lot of unrelated dead code

## Pull Request Check List

## Review

  - Merge the pull request


package.json
```diff
@@ -27,7 +27,7 @@
     "dev-nowatch": "bash -c 'NODE_ENV=development node server/server.js & NODE_ENV=development webpack-dev-server'",
     "test-accessibility": "./test/accessibility.sh",
     "test-unit": "yarn test-unit:components && yarn test-unit:store && yarn test-unit:app",
-    "test-unit:app": "mocha --require mock-local-storage --file test/unit/helpers/babel-register.js --file test/unit/helpers/init.js \"test/unit/**/*.test.js\"",
+    "test-unit:app": "mocha  --exit --require mock-local-storage --file test/unit/helpers/babel-register.js --file test/unit/helpers/init.js \"test/unit/**/*.test.js\"",
```
optionsome, line 30: It's unclear why the tests didn't exit anymore without `--exit` flag. It's probably because there are some open handles somewhere but I couldn't debug the problem and not sure if it is serious or not.

optionsome:
Note, I didn't test with login. It could be that some of the lazy loading, that I've removed, was because of login instead of SSR.

vesameskanen (approved):
great PR

PR #5416, opened 2025-06-10
DT-6183 add taxi results to alternative search results

sharhio:
## Proposed Changes

  - Make flex query if no routes were found and taxi option is not selected.
  - Combine flex results with the existing alternative scooter results.
  - Fixed translations to include taxi only versions and fixed naming. 
  - Fixed notification logic to show correct version of notification (no suggestions for modes which are not available).

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/itinerary/ItineraryPageUtils.js
+++ b/app/component/itinerary/ItineraryPageUtils.js
@@ -594,6 +594,32 @@ export function mergeScooterTransitPlan(
   return sortAndMergePlans(scooterTransitEdges, transitPlan, arriveBy);
 }
 
+/**
+ * Combine two sets of external plans.
+ */
+export function sortAndMergeExternalPlans(
+  plan,
+  planToMerge,
+  arriveBy,
+  maxTotalEdges = 5,
+) {
+  const edges = plan.edges || [];
+  const edgesToMerge = planToMerge.edges || [];
+  return {
+    edges: getSortedEdges([...edges, ...edgesToMerge], arriveBy)
+      .slice(0, maxTotalEdges)
+      .map(edge => {
+        return {
+          ...edge,
+          node: {
+            ...edge.node,
+            legs: compressLegs(edge.node.legs),
+          },
+        };
+      }),
+  };
+}
+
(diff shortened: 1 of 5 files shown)


app/util/planParamUtil.js
```diff
@@ -258,13 +258,13 @@ export function planQueryNeeded(
...
+    /* special logic: relaxed flex query is made only if taxis are not allowed */
     case PLANTYPE.FLEXTRANSIT:
       return (
         config.experimental?.allowFlexJourneys &&
-        settings.includeTaxiSuggestions &&
         (transitModes.length > 0 ||
-          config.experimental?.allowDirectFlexJourneys)
+          config.experimental?.allowDirectFlexJourneys) &&
+        settings.includeTaxiSuggestions === !relaxSettings
       );
```
vesameskanen, line 268: Would !== work here?

app/component/itinerary/ItineraryPageUtils.js
```diff
@@ -594,6 +594,32 @@ export function mergeScooterTransitPlan(
...
+          node: {
+            ...edge.node,
+            legs: compressLegs(edge.node.legs),
+          },
+        };
+      }),
+  };
+}
+
 const ITERATION_CANCEL_TIME = 20000; // ms, stop looking for more if something was found
```
vesameskanen, line 623: other flex merge functions seem to filter by allowedFlexRouteTypes. That is not required here?

vesameskanen (changes requested):

app/util/planParamUtil.js
sharhio: It would indeed.

app/component/itinerary/ItineraryPageUtils.js
sharhio: Both the relaxed scooter and the relaxed flex searches do the filtering right after search (before merge) to ensure the correct notification logic is triggered only when there are relevant results.

vesameskanen (approved):

PR #5422, opened 2025-06-12
DT-6839 Navigator state management refactor

partisaani:
## Proposed Changes
- Crafted ItineraryContext to store single Itinerary related data and reduce prop drilling
- Refactored Navigator relevant state management higher up in hierarchy into the newly introduced ItineraryContext
- State is managed using useItineraryReducer hook in order to make sure relevant state is updated always in one go and to give us greater control over state updates
- Removal of obsolete code
- Import reorganization here and there

## Pull Request Check List

## Review

  - Merge the pull request


app/component/itinerary/ItineraryPageContainer.js
```diff
@@ -4,19 +4,28 @@ import { connectToStores } from 'fluxible-addons-react';
...
   <ReactRelayContext.Consumer>
-    {({ environment }) => (
-      <ItineraryPage {...props} relayEnvironment={environment} />
-    )}
+    {({ environment }) => {
+      return (
+        <ItineraryContextProvider
+          getStore={getStore}
+          relayEnvironment={environment}
+        >
```
vesameskanen: I do not see any reference to getStore and relayEnvironment in ItineraryContext.

app/component/itinerary/navigator/NaviContainer.js
```diff
@@ -92,38 +87,51 @@ function NaviContainer(
...
     }
     prevPos.current = position;
-  }, [time]);
+  }, [params.updatedAt]);
 
   useEffect(() => {
-    if (firstLeg && time > legTime(firstLeg.start) - START_BUFFER) {
+    if (
+      firstLeg &&
+      !firstLeg.freezeStart &&
```
vesameskanen: What is the reason for the new condition   `!firstLeg.freezeStart  ` ?

vesameskanen (changes requested):
It is nice that the big useRealtimeLegs has been split to smaller parts. Code also changes look good.  However, testing indicates some strange regressions:

- If you set biking speed to max (that triggers simulated transfer problem), navi crashes because updatedAt is true, not a number representing a time

- Leg time adjustement by geolocation (useProcessLegs.js / GEOLOCATED_LEGS switch) no longer works. Adjustement gets stuck to first transit leg. I suspect that code somehow sees a vehicle position which represents frozen state. There were similar problems in the past.

app/component/itinerary/ItineraryPageContainer.js
partisaani: Might be a leftover from some development iteration, I'll look into this.

app/component/itinerary/navigator/NaviContainer.js
partisaani: To prevent startItinerary from being called every time legs are updated. I'll investigate if there's a better/more correct way to prevent this from happening.

vesameskanen: ok. Maybe comparison with forceStartAt  is more reliable. If first leg is a transit leg, freezeStart will not be set.

partisaani:
> * If you set biking speed to max (that triggers simulated transfer problem), navi crashes because updatedAt is true, not a number representing a time

Solved by correcting params (https://github.com/HSLdevcom/digitransit-ui/pull/5422/commits/dff13dd17714f9defba7db7cbc1a54aa1645a569)

> * Leg time adjustement by geolocation (useProcessLegs.js / GEOLOCATED_LEGS switch) no longer works. Adjustement gets stuck to first transit leg. I suspect that code somehow sees a vehicle position which represents frozen state. There were similar problems in the past.

Solved by correct usage of useCallback and its dependency array (https://github.com/HSLdevcom/digitransit-ui/pull/5422/commits/4fdcea8fba868bdda7e478ec8c5c5e231f142ba9)

vesameskanen (approved):

PR #5426, opened 2025-06-17
DT-6943 fix: navi starter rendering fixed

vesameskanen:
StartItinerary call to start navigation immediately did not trigger rendering. This had bad side effects:

- Start modal was rendered for 10 seconds although code already triggered navigation
- When user clicked 'Start now' , instructions appeared after a long delay (10 s in worst case)

Rendering is now triggered using local state variable. Also, itinerary start is set one seconds in the past to ensure that NaviCardContainer finds a leg to render.


partisaani (approved):

partisaani (approved):
LGTM

PR #5432, opened 2025-06-24
DT-4133 Replace all moment dependencies in shared components

hjvuor:
## Proposed Changes

  - Replaces Moment.js in digitransit-util, digitransit-component and digitransit-store
  - Cleanup and remove last moment dependencies in app and server
  - Removes moment and moment-timezone from package.json
  - Removes config.timezoneData as its no longer needed
  - In some instances where we just need a simple unix timestamp, use native date instead of importing luxon or moment
  - **I believe unit-tests for the packages are not included in the CI build**
      - fixing one package would require to fix all of the packages as lerna runs the test commands recursively
## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/digitransit-component/packages/digitransit-component-autosuggest-panel/package.json
+++ b/digitransit-component/packages/digitransit-component-autosuggest-panel/package.json
@@ -28,7 +28,7 @@
   "author": "Digitransit Authors",
   "license": "(AGPL-3.0 OR EUPL-1.2)",
   "peerDependencies": {
-    "@digitransit-component/digitransit-component-autosuggest": "^3.0.6",
+    "@digitransit-component/digitransit-component-autosuggest": "^4.0.0",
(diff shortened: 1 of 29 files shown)


digitransit-component/packages/digitransit-component-autosuggest-panel/package.json
```diff
@@ -28,7 +28,7 @@
   "author": "Digitransit Authors",
   "license": "(AGPL-3.0 OR EUPL-1.2)",
   "peerDependencies": {
-    "@digitransit-component/digitransit-component-autosuggest": "^3.0.6",
+    "@digitransit-component/digitransit-component-autosuggest": "^4.0.0",
```
vesameskanen, line 31: This package changed. I think we have to  increment the major version for it as well.

vesameskanen (changes requested):
Good work! Finally we get rid of moment.

PR #5439, opened 2025-07-16
DT-6539 Logos for social media share links

sharhio:
## Proposed Changes

  - Uses assets already loaded and added in the manifest when available, otherwise uses static images
  - Added png versions of svg logos (more likely to show up in the preview)
  - Renamed some ambiguously named logos
  - Fixed broken hsl logo file
  - Configurations updated

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen (commented):
Images seem to live in 2 folders: configurations/images and static/img.  UI suggest using a single consistent folder.

sharhio:
> Images seem to live in 2 folders: configurations/images and static/img. UI suggest using a single consistent folder.

Sure. In that case all of the images will have to be in the static/img as the build refuses to include some of the images in the manifest for unknown reasons.

vesameskanen (commented):
Sorry for overlooking this before, but I  just checked how UI loads various assets:
- geojson  files are loaded via CDN
- favicons seem to come from UI server directly
- static/img/nearby-stop_animation.gif (shown in empty nearest fav tab)  is loaded directly from UI server
- svg asssets seem to load from url 'data:image/svg+xml', I don't know how that works out

I guess we have to carefully avoid moving more asset loading away from CDN. Maybe it only means that img folder needs to go under assets folder? Could you verify this? 

I am not sure if we have to start using versioned names when using CDN, just like with geojson files. Browsers might cache files such as 'citylogo.png' anyway, so CDN might not play any role in naming.

PR #5443, opened 2025-07-22
AB#135 Fix TerminalPage bugs related to multiple departures

hjvuor:
## Proposed Changes

  - Leverages stoptime.stop.gtfsId in addition to other fields to make keys unique in case the same pattern departs at the same time on different platforms
  - This was broken in both DepartureList and TimetablePage
  - In addition fixes a bug where the same pattern was displayed multiple times in the filter modal, which was related to the same issue

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/stop/FilterTimeTableModal.js
+++ b/app/component/stop/FilterTimeTableModal.js
@@ -85,12 +85,10 @@ class FilterTimeTableModal extends React.Component {
       )
       .filter(o => o)
       .sort(routeCompare)
-      // deduplicate identical headsign + shortName combinations
+      // deduplicate patterns with same code
       .filter(
-        (li, index, self) =>
-          self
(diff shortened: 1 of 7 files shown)


app/component/stop/FilterTimeTableModal.js
hjvuor, line 1: Noticed that in some cases, we actually have patterns with identical shortName and headsign that still have unique codes, in this case if we deduplicate like this, some data is lost as the timetable is filtered based on the patternCode. I think there could be a design that identifies different patterns for the user in the UI even though they have identical shortNames and text.

If deduplication is done based on patternCode, it solves the case in Kontula but for example 114N in Leppävaara is still duplicated as it has two codes with the same shortName and headsign

tkalvas (approved):

PR #5447, opened 2025-08-07
DT-6861 - Change digitransit-component-autosuggest

VillePihlava:
## Proposed Changes
This does not fully fix all issues from the ticket

Changes:
 - close mobile search dialog on blur (e.g. when tab is pressed), the focus goes to the input search element that the mobile search dialog was opened on
 - fix shift+tab combination
 - make escape work similarly to tab
 - do not automatically open mobile search dialog when search is focused

### DIFF
--- a/digitransit-component/packages/digitransit-component-autosuggest/package.json
+++ b/digitransit-component/packages/digitransit-component-autosuggest/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@digitransit-component/digitransit-component-autosuggest",
-  "version": "4.0.0",
+  "version": "4.1.0",
   "description": "digitransit-component autosuggest module",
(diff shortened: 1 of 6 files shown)


hjvuor (changes requested):
You should bump the patch version number, otherwise looks good

VillePihlava:
> You should bump the patch version number, otherwise looks good

Fixed

digitransit-component/packages/digitransit-component-autosuggest/package.json
hjvuor, line 1: should also update the dependency in digitransit-component/packages/digitransit-component

hjvuor (changes requested):

hjvuor (approved):

PR #5450, opened 2025-08-11
AB#107 Fix Autosuggest icons to use route.color if it is available

hjvuor:
## Proposed Changes

  - Changes suggestion icon to always use route_color from gtfs if it is specified
  - Route color overrides possible mode specific colors
  - Also replaces recompose/pure with React.memo since it is the modern equivalent
    - Decided to also remove usage of pure from app folder in this pr since there were only a few instances

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/digitransit-component/packages/digitransit-component-suggestion-item/src/index.js
+++ b/digitransit-component/packages/digitransit-component-suggestion-item/src/index.js
@@ -299,7 +299,7 @@ const SuggestionItem = pure(
       >
         <Icon
           color={
-            item.properties.color
+            item?.properties?.color
               ? `#${item.properties.color}`
               : modeIconColor || iconColor
           }
--- a/digitransit-component/packages/digitransit-component-suggestion-item/src/index.js
+++ b/digitransit-component/packages/digitransit-component-suggestion-item/src/index.js
@@ -1,8 +1,7 @@
 /* eslint-disable import/no-extraneous-dependencies */
 import PropTypes from 'prop-types';
-import React, { useState } from 'react';
+import React, { useState, memo } from 'react';
 import cx from 'classnames';
-import pure from 'recompose/pure';
 import Icon from '@digitransit-component/digitransit-component-icon';
 import styles from './helpers/styles.scss';
 
@@ -263,7 +262,7 @@ function hasVehicleStationCode(stationId) {
  *    loading={false}
  * />
  */
-const SuggestionItem = pure(
+const SuggestionItem = memo(
   ({
     item,
     content,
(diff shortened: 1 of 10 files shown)


digitransit-component/packages/digitransit-component-suggestion-item/src/index.js
```diff
@@ -297,7 +296,14 @@ const SuggestionItem = pure(
       <span
         className={`${styles[iconId]} ${item?.properties?.mode?.toLowerCase()}`}
       >
-        <Icon color={modeIconColor || iconColor} img={iconId} />
+        <Icon
+          color={
```
vesameskanen, line 300: It seems that item must be defined, so ? is not necessary here. Please fix itemn references in this function.

digitransit-component/packages/digitransit-component-suggestion-item/src/index.js
```diff
@@ -517,11 +523,32 @@ SuggestionItem.propTypes = {
...
+        name: PropTypes.string,
+        localadmin: PropTypes.string,
+      }),
+    }),
   }),
   // eslint-disable-next-line
   content: PropTypes.array,
   className: PropTypes.string,
   isMobile: PropTypes.bool,
+  ariaFavouriteString: PropTypes.string,
```
vesameskanen, line 549: Not required prop, but no default value.

digitransit-component/packages/digitransit-component-suggestion-item/src/index.js
```diff
@@ -517,11 +523,32 @@ SuggestionItem.propTypes = {
...
+      arrowClicked: PropTypes.bool,
+      destination: PropTypes.shape({
+        name: PropTypes.string,
+        localadmin: PropTypes.string,
+      }),
+      origin: PropTypes.shape({
+        name: PropTypes.string,
+        localadmin: PropTypes.string,
+      }),
+    }),
```
vesameskanen, line 543: I wonder if these props are always populated? SuggestionItem might be a route like 'M1'. browser's debugger does not complain but I am wondering if props should be a minimal common set of fields, instead of all possible values which might be present.

vesameskanen (changes requested):

digitransit-component/packages/digitransit-component-suggestion-item/src/index.js
hjvuor, line 1: PropTypes had to be defined in this manner to keep linter silent. In addition to Autosuggest the SuggestionItem is also used in Favourites bar where the item is very much of different shape

vesameskanen (approved):

PR #5452, opened 2025-08-13
Do not call onClick handler at every key event in stops near you buttons

vesameskanen:
The buttons now filter selection events themselves. This will fix accessibility error in hsl.fi front page.


hjvuor (approved):
Changes look good and its good that the key press filtering is done in the component itself.

PR #5456, opened 2025-08-15
Refactor control panel component

vesameskanen:
Control panel provided several components but only one added language bundles to i18next. Translations in hsl.fi front page were broken because they do not use the mentioned root component. 

- Remove dead sub component
- Initialize i18next using useEffect and do not render before it becomes ready
- Add missing translations


hjvuor (approved):
Looks good to me!

PR #5459, opened 2025-08-19
DT-6962 - Order search results with a letter at the end of `shortName` first

VillePihlava:


vesameskanen (changes requested):
Could the longname be a secondary sort criteria? Now in Matka search, search 8 gives very many exact hits and they appear in random order by long name.

vesameskanen (approved):

PR #5460, opened 2025-08-19
Fix stop view header alignment

sharhio:
## Proposed Changes

  - Consistent styles with no hardcoded widths (supports all languages)

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/stop/stop.scss
+++ b/app/component/stop/stop.scss
@@ -285,19 +285,22 @@
   }
 
   .route-number-header {
-    flex: 0 0 3.5em;
+    flex: 0 0 fit-content;
+    margin-right: 0.5em;
   }
 
   .route-destination-header {
-    flex: 1 0 6em;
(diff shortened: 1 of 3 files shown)


vesameskanen (commented):
Would it be possible to move 'Määränpää' header to start where the respective column starts? Now in HSL and Waltti theme, the terminal page renders it a bit too much over the shortname column.

sharhio:
> Would it be possible to move 'Määränpää' header to start where the respective column starts? Now in HSL and Waltti theme, the terminal page renders it a bit too much over the shortname column.

I made it theme dependent but I'm not very happy with it. The flex layout is not very suitable for aligning these columns as the sizes of contents, the sizes of fonts (different between titles and contents as well) and the lengths of different languages cause too much variance. The table should probably be a grid layout instead so headers and columns would always be aligned.

PR #5467, opened 2025-08-26
DT-6863 Landing page accessibility improvements 

partisaani:
## Proposed Changes
Assorted accessibility improvements on the page users first see:
- Clear search history -button now reacts when user browses onto it using keyboard
- Tweaked Favourites list toggling button's aria-attributes as suggested by Q-Factory review
- Fixed incorrectly initialized i18n in dialog modal component

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/digitransit-component/packages/digitransit-component-autosuggest-panel/package.json
+++ b/digitransit-component/packages/digitransit-component-autosuggest-panel/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@digitransit-component/digitransit-component-autosuggest-panel",
-  "version": "5.5.1",
+  "version": "5.5.2",
   "description": "digitransit-component autosuggest-panel module",
   "main": "index.js",
   "files": [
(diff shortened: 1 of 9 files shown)


vesameskanen (changes requested):
One request: could you keep library versions exactly in sync? I am not sure if that matters much, but it is the safest way. Dialog modal-was bumped to 0.3.8 but other libraries seem to ask for 0.3.7.

Please note that all packages using dialog-modal need version bump, and that change may spread further. A bit boring  but we do not have a tool for cascading version bumps.

I also have a concern about i18next initialization in other libraries. Many seem to use it and the code usually incorrectly adds language bundles although init promise has not resolved yet. We already know that this broke control panel translations in hsl.fi. The same can happen again.

vesameskanen:
Maybe i18next is out of the scope of this PR. I took a quick look and there is a need for overall refactoring in many components. It seems wiser to handle one component in each PR - convert classes to function components and fix i18next use at the same time.

PR #5470, opened 2025-08-27
DT-6962 - Fix search ordering

VillePihlava:


vesameskanen (approved):
Seems to work now fine

PR #5481, opened 2025-09-16
AB#178 refactor i18next usage

vesameskanen:
It appears that i18next conventions in digitransit ui code have been mostly wrong.  In this PR, following more correct conventions are adopted:

- i18next is initialized only once in main app initialization. This is how it is designed to work.
- addResourceBundle sets 'deep' option when adding per module bundles. Without the option, new bundle is supposed to overwrite earlier bundles (docs https://www.i18next.com/overview/api#addresourcebundle are a bit unclear about this, though). The reported bug in embedded search demonstrates this: rendering mobile search erased most existing translations
-  UI code no longer repeatedly uses changeLanguage to set target language of translations. This is doomed to fail: changeLanguage is  async and rendering can happen before it has any effect. Instead, target language is passed along every i18next.t() call.
- There were lots of  missing lang props in digitransit components, and the prop defaulted to 'fi'. Code usuallly worked because other components were frequently calling changeLanguage, which has global (not component private) effect.

This PR also contains miscellaneous minor fixes detected along the way.

Component major versions are bumped in tris PR, because the change is not backward compatible. Main app must now do the init call when setting up the application, otherwise the rest of i18next calls  will cause unhandled errors.

There is one major concern left: shared digitransit-component libs add their translation bundles at module loading, right after imports. This means that i18next init by the main app must be executed first, and there is no guarantee that application loading happens that way. For some unknown reason, this seems to be the case.  If translations are added later, for example using a react useEffect, the components seem to render too fast without translations. I have not found a way to fix this.

Enabling 'debug' option in app/client.js i18next.init will help to see translation problems. For example, embedded search in path /haku also renders first without translations, but a state change later fixes this.


hjvuor (approved):
Looks good!

PR #5487, opened 2025-09-22
AB#106 svg cleanup

vesameskanen:
- All svg icons processed with svgo to reduce unnecessary transforms and redundant hierarchy
- coded citybike icon variations instead of duplicating them just to get different badge color and size (50 icons less) 
- remove redundant metadata and attributes
- use consistent fill color from currentColor instead of hard coded colors or class defined colors
- name icons in more sensible and consistent way
- info icon of matka disruption modal was invisible, fixed now

Sprite sheet sizes are now less than one half from what they used to be (HLS 610k -> 260k).


digitransit-component/packages/digitransit-component-icon/src/assets/check.svg
hjvuor, line 1: Look like this icon is broken

vesameskanen: yes it was, should be fixed now

PR #5497, opened 2025-10-06
DT-6961 Fix broken purchase link tampere

hjvuor:
## Proposed Changes

  - This PR fixes the mobile ticket purchase link in tampere
  - Adds the ticket link function to the tampere config as the data needs a different mapping than other waltti cities

### DIFF
--- a/app/component/itinerary/MobileTicketPurchaseInformation.js
+++ b/app/component/itinerary/MobileTicketPurchaseInformation.js
@@ -59,12 +59,7 @@ export default function MobileTicketPurchaseInformation(
           onClick={() =>
             addAnalyticsEvent({ event: 'journey_planner_open_app' })
           }
-          href={config.ticketPurchaseLink(
-            fare,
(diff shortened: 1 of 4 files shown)


vesameskanen (changes requested):
Just a couple of minor things: 
- I would move zoneMapping out from the function, so it won't be redefined at every render
- Could you refactor the function parameters? I find it confusing that appName  and operatorCode are passed as parameters, although they are defined in the config itself.

hjvuor:
> Just a couple of minor things:
> 
> * I would move zoneMapping out from the function, so it won't be redefined at every render
> * Could you refactor the function parameters? I find it confusing that appName  and operatorCode are passed as parameters, although they are defined in the config itself.

The function is copied from the waltti config with the addition of zoneMapping, to sort of override that function I shouldn't touch the parameters. I'm also not sure if I should move the zoneMapping outside the function scope since it is only used inside and it doesn't make sense to expose it in the config.

PR #5504, opened 2025-10-13
AB#229 Indicate to the user that the settings have been changed

sharhio:
## Proposed Changes

  - Indicator icon for changed settings
  - Popover notification of changed settings to be shown once when settings have been changed
  - New hovering button for resetting to default settings
  - A notification that the settings have been restored to defaults

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen (changes requested):
New elements look nice! Some remarks:
- In sv lang, the settinsg changed element at the bottom of search settings gets so high that it obscures the lowest option. Text on the left side has really large line interval.
- Mac + voiceOver: when any of the new elements pops up, screen reader does not read it. Would it be useful to add aria-live messages to them?
- Voiceover reads a mysterious 'Sulje' button between last settings option and reset settings content. Without voiceover, it will not get focus. Trying to select that button does nothing.
- When reset settings is selected, focus seems to jump out from reittiopas.

app/component/itinerary/customize-search.scss
```diff
@@ -571,3 +699,23 @@
...
+
+@keyframes slideDownWithMargin {
+  from {
+    transform: translateY(0);
+  }
+
+  to {
+    transform: translateY(calc(100% + 16px));
+  }
+}
```
partisaani: Navigator has similar animations in `navigator.scss`. It could be looked into if these animations could be extracted into a common `animations.scss`

partisaani (changes requested):
LGTM with comment

vesameskanen (commented):
Popups about changed settings are now read fine and focus handling is OK. However, I can still see two problems with voiceOver in Mac:

- When you reset settings back to default, the green snackbar confirming the change is not read
- After applying reset settings, the strange 'Sulje' button appears into the DOM. Is that some outdated rendering code? See the screenshot below.

image

vesameskanen (approved):
Works now as expected.

I noticed that mobile settings view shifts focus out from reittiopas after the last settings option. This is an old bug and can be fixed in another PR.

PR #5528, opened 2025-11-03
Ab#174 patch 1

partisaani:
## Proposed Changes
Contains features that were missing from initial implementation as well as changes to current implementation based on discussions with designer.

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen (approved):
Good proceedings! Some further steps to consider:
- Should the view consider extended modes (use dedicated color and icon for example for speedtram)? If so, map `route.mode`using modeUtils.js/getRouteMode(). 
- Some traffic mode specific styles are missing (e.g. ferry), so the route number square renders white
- Route list uses a regular link to route page. This has kind of undesired effect that it does not update application's internal router history.  If you visit route page and click the back arrow (not the one on browser but the one we render), user ends up to front page. To avoid this, one has to add a click handler and use router.push().

partisaani:
> * Should the view consider extended modes (use dedicated color and icon for example for speedtram)? If so, map `route.mode`using modeUtils.js/getRouteMode().
> * Some traffic mode specific styles are missing (e.g. ferry), so the route number square renders white
> * Route list uses a regular link to route page. This has kind of undesired effect that it does not update application's internal router history.  If you visit route page and click the back arrow (not the one on browser but the one we render), user ends up to front page. To avoid this, one has to add a click handler and use router.push().

I shall tackle these in patch 2!

PR #5545, opened 2025-11-12
AB#307refactor path generation of the route page 

vesameskanen:
- route /  trip / pattern GTFS ids are encoded whenever they are used as part of page URL
- messy string interpolations to generate URLs replaced with a single utility function
- misc refactoring

### DIFF
--- a/app/component/routepage/PatternStopsContainer.js
+++ b/app/component/routepage/PatternStopsContainer.js
@@ -29,16 +29,16 @@ class PatternStopsContainer extends React.PureComponent {
   };
 
   render() {
+    const routeId = this.props.route.gtfsId;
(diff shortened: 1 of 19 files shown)


partisaani (approved):

app/component/routepage/PatternStopsContainer.js
```diff
@@ -29,18 +29,16 @@ class PatternStopsContainer extends React.PureComponent {
   };
 
   render() {
+    const routeId = this.props.route.gtfsId;
```
hjvuor, line 32: this.props.route might be null if an invalid routeId is typed directly in the URL. in that case this crashes

hjvuor (changes requested):

PR #5553, opened 2025-11-18
AB#174 patch 3

partisaani:
## Proposed Changes
- Only scroll and overflow the Alerts element instead of whole page
- Use SASS functions to create mode classes dynamically from a map object
- Present Stops in Disruption cards instead of their associated routes
- Added `backgroundShape==="stopsign"` which shrinks the main icon to fit a lollipop styled background representing a traffic sign
- Assorted review and feedback fixes

## Pull Request Check List

## Review

  - Merge the pull request


app/component/Icon.js
```diff
@@ -98,10 +103,38 @@ function Icon({
             r="20"
           />
         )}
-        {!dataURI && <use xlinkHref={`#${img}`} />}
-        {dataURI && (
-          <image href={dataURI} x={0} y={0} width="100%" height="100%" />
+        {backgroundShape === 'stopsign' && (
+          <>
+            <circle
+              className="icon-stopsign-circle"
```
vesameskanen: I can't find this style class anywhere.

app/component/Icon.js
```diff
@@ -98,10 +103,38 @@ function Icon({
...
+              cx="20"
+              cy="20"
+              fill={backgroundColor}
+              r="13.33"
+            />
+            <rect
+              className="icon-stopsign-pole"
```
vesameskanen: same as above.

app/component/trafficnow/utils.js
```diff
@@ -0,0 +1,80 @@
...
+    id,
+    shortName,
+    name,
+    gtfsId,
+    locationType,
+    isStop = false,
+    isStation = false,
+  },
+) => {
+  const url = `/${getUrlPrefix(locationType)}/${gtfsId}`;
```
vesameskanen: There is a need to encode GTFS identifiers before adding them to URL. I did that for route page paths in:

https://github.com/HSLdevcom/digitransit-ui/pull/5545

Stop page URLs encoding is not yet handled in consistent way - it happens in may places across the code.

app/component/trafficnow/utils.js
```diff
@@ -0,0 +1,80 @@
...
+    name: shortName || name,
+    url,
+    isStop,
+    isStation,
+  });
+};
+
+const groupEntitiesByMode = (entities, config) => {
+  const group = entities
+    .filter(e => e.__typename !== 'Unknown')
```
vesameskanen: AlertEntityType.Unknown  ?

vesameskanen (changes requested):
Good proceedings!

I am a bit unhappy how Icon component works but its troubles have accumulated along a long time period. I would expect `Icon` to be very a generic component. Now it has strange special features and dependencies. An example: If a class `household appliance` has an attribute `be vacuum cleaner` OO design is all wrong :)

Maybe we will refactor the component at a later date.

app/component/trafficnow/utils.js
partisaani: Oversight, will fix

app/component/trafficnow/utils.js
partisaani: So it's probably for the best to merge #5545 first to v3, bring those changes to next and then put the encoder to use in this use case as well.

app/component/Icon.js
partisaani: I added classes for the shapes to follow the path set by the plain circle background above

app/component/Icon.js
partisaani: I added classes for the shapes to follow the path set by the plain circle background above

partisaani:
> I am a bit unhappy how Icon component works but its troubles have accumulated along a long time period. I would expect `Icon` to be very a generic component. Now it has strange special features and dependencies. An example: If a class `household appliance` has an attribute `be vacuum cleaner` OO design is all wrong :)
> 
> Maybe we will refactor the component at a later date.

I would rather do it sooner than later. A good, simple start would be to extract the background shapes to be implemented by something like IconWithBackground

app/component/Icon.js
vesameskanen: I still don't understand - should'nt there be a css class` icon-stopsign-circle` somewhere adding some styles ?

partisaani: There were no classes defined to add styles, it was merely for consistency's sake when comparing different backgroundShapes.

I have removed the className prop from the component which can now be found in `IconBackground.js`.

vesameskanen (approved):

PR #5559, opened 2025-11-20
AB#312 Vaasa service area

tuomas-lindell:
## Proposed Changes

  - Updated Vaasa service area

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/configurations/config.vaasa.js
+++ b/app/configurations/config.vaasa.js
@@ -1,4 +1,3 @@
-/* eslint-disable */
 import configMerger from '../util/configMerger';
 import walttiConfig from './config.waltti';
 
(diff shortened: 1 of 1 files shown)


vesameskanen (changes requested):
- Vaasa service area is still a rectangle. Just update minLat ...maxLon values. Rectangle is much faster geometry to limit geocoding search and requires less data.
- I can see some formatting changes which violate our es-lint rules. For some strange reason, many configs have eslint disabled in the first line of config. Please remove that comment line and fix all es-lint errors while you update the service area  in each config. Note that you can apply yarn es-lint --fix to address most lint errors.

tuomas-lindell:
- changed service area back to rectangle
- Restored old file formatting, removed es-lint comment and did run yarn es-lint --fix, now result after running es-lint formatting is in committed file (formatting looks to me same as other config-files)

vesameskanen (changes requested):
I think minLat and maxLat have reversed values.

vesameskanen (approved):

PR #5572, opened 2025-11-26
AB#115 Migrate RoutePatternSelect from react-autosuggest to downshift

hjvuor:
## Proposed Changes

  - This PR replaces react-autosuggest in RoutePatternSelect with downshift.js
  - Fixes a bug where the dropdown was not properly closing
  - Fixes an accessibility problem where the dropdown was opened instantly when focused
  - The old logic in RoutePatternSelect was moved into RoutePatternSelectContainer and new RoutePatternSelect only implements the downshift logic and JSX
    - diff is hard to read
  - Screen reader functionality was tested using mac VoiceOver using chrome, firefox and safari, there are some inconsistences between these but all seem to read the correct information. Should also be tested using a Windows machine
  - digitransit-component-autosuggest will also be refactored to use downshift so this work is also a part of getting rid of the react-autosuggest dependency

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen (approved):
Seems to work well enough with voiceOver. I noticed that when the whole content of combobox is read, Safari repeats all list item 3 times. Chrome also does a lot of repetition and reads arrow symbols as well. Such content is a bit annoying but not a fatal error.

partisaani:
image

Tested with Win11, NVDA and Chrome|Firefox

- There is some excessive repetition which is probably due to overlapping element arias or the way NVDA handles lists
- When browsing to lowest div with 7H route, NVDA also declares aloud the two divs above it with upcoming routes.
- NVDA does not tell the user which of the fields is selected when browsing 
- When the list is opened and NVDA reads through all of the options, it fails to read aloud the H in 7H. The problem does not occur when the user focuses on the said element.

hjvuor:
voiceOver on Mac reads the aria-selected attribute if it's set to true, but I now added a custom translated text to the label if the item is selected. We decided that as the component is usable on both voiceOver and NVDA, we can merge as is and improve the usability later. The UX improvements to NVDA should be done by someone with a Windows machine to make the development easier.

On a side note I think I've accurately followed the ARIA 1.2 guidelines for a single select combobox but screen readers still act inconsistently accross different browsers which is quite annoying.

vesameskanen (approved):

PR #5574, opened 2025-11-26
AB#94 Navigate to Traffic now page using drawer menu

partisaani:
## Proposed Changes
- Disruptions CTA now navigates user to Traffic now page
- Removed the now obsolete DisruptionListContainer and DisruptionInfo component

## Pull Request Check List

## Review

  - Merge the pull request


app/component/DisruptionInfoButtonContainer.js
```diff
@@ -5,17 +5,21 @@ import { matchShape, routerShape } from 'found';
...
+) {
   const { environment } = useContext(ReactRelayContext);
   const openDisruptionInfo = () => {
-    setDisruptionInfoOpen(true);
     addAnalyticsEvent({
       category: 'Navigation',
       action: 'OpenDisruptions',
       name: null,
     });
+    router.push(`/${lang === 'fi' || !lang ? '' : `${lang}/`}${TRAFFICNOW}`);
```
vesameskanen: You should leave lang away. UI reads language from a cookie when it starts, lang is set to preferencesStore and remains as selected until user changes it. No need to touch it when routing to a new page.

vesameskanen (changes requested):

app/component/DisruptionInfoButtonContainer.js
vesameskanen: I found similar code from IndexPage. Maybe it is a remainder from the time when TrafficNow linked out from our UI to HSL.fi and lang hand to be defined.

partisaani: I'll refactor them both first thing tomorrow

vesameskanen (approved):

PR #5585, opened 2025-12-02
Refactor some digitransit library code

vesameskanen:
- Icon component always rendered all  (almost 100) icons when only one was needed, fixed
- Better IO error tolerance in search


digitransit-search-util/packages/digitransit-search-util-helpers/index.js
```diff
@@ -181,7 +181,7 @@ export const getLayerRank = (layer, source) => {
  * @param {String} term The search term that was used
  */
 export const sortSearchResults = (lineRegexp, results, term = '') => {
-  if (!Array.isArray(results)) {
+  if (!Array.isArray(results) || results.length) {
```
hjvuor: Is this a logical error? if `results` is an array with length more than 0 this evaluates to `true`, whereas before it would check if `results` is NOT an array.  Maybe the intended logic was:
 `if (!(Array.isArray(results) || results.length))` ?

hjvuor (changes requested):

digitransit-search-util/packages/digitransit-search-util-helpers/index.js
hjvuor: or maybe I am missing the point of checking for both as i just looked into this out of context

digitransit-search-util/packages/digitransit-search-util-helpers/index.js
```diff
@@ -191,7 +191,7 @@ export const sortSearchResults = (lineRegexp, results, term = '') => {
   const isLineSearch = isLineIdentifier(normalizedTerm);
 
   const orderedResults = orderBy(
-    results,
+    results.filter(r => r),
```
partisaani, line 194: This filter does seemingly nothing

partisaani (commented):
Left a note on weird Array.prototype.filter call

digitransit-search-util/packages/digitransit-search-util-helpers/index.js
hjvuor: i think this actually filters out "falsy" values from the list

vesameskanen: Tried some italian data and array was [null, null, object, object, null, ....]

vesameskanen:
Let's wait until tomorrow before merging this, to allow some dev testing time. We had to do many releases because of route page bugs, the same may happen again tomorrow.

partisaani (approved):

hjvuor (approved):

PR #5592, opened 2025-12-08
AB#172 Filter Traffic now content based on validity period

partisaani:
## Proposed Changes
- Adds FiltersContext which, for now, allows filtering alerts based on their validity period i.e. are they in effect or not

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen (approved):
Clean code and works well. One note, though: maybe this is a data issue, but hameenlinna planner seems to render same alert tens of times, see the picture below. Debugger also complains that many elements were rendered with same key. If data can be bad, maybe we have to filter duplicates ourselves.
image

PR #5594, opened 2025-12-11
Ab#137 improve how transfers are displayed

tuomas-lindell:
## Proposed Changes

  - changed waiting picto to standing, after this change sitting picto is used only for interlining legs

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen (changes requested):
This works but let's improve the current setup a  bit:
- rename icon_wait as icon_wait_sitting, and icon_navigation_wait as icon_wait_standing
- rename references to these icons respectively
- icon_wait_sitting renders smaller than other respective icons. Extra padding should be removed. I can do this if you are not familiar with the required tools.
- icon_wait_sitting seems to have a hard coded fill color. Color should be set using css styles of the content which renders the icon. Most likely it is enough to remove color definition  from svg.

vesameskanen (changes requested):
Now there seems to be some confusion.  icon_wait_sitting seems to render a standing person. In navigation components, both icons are referred,  although only the standing one should be used.

InterlineInfo.js refers to original icon_wait. icon_wait should be renamed as icon_wait_sitting, because it indeed shows a sitting person.

In sprite sheets, icon_navigation which has nothing to do with waiting was renamed to something else.

vesameskanen (approved):

PR #5595, opened 2025-12-11
AB#168 - Improve indoor routing level info

VillePihlava:
## Proposed Changes
  - Add indoor routing to itinerarydetails, navigator, and map
  - Indoor routing features include:
    - New line color
    - Displaying metro entrance icons
      - Metro icon
      - Metro entrance letter icon
      - Wheelchair icon for wheelchair accessible entrances
    - Indoor route step markers on map
      - Clicking the step marker opens a popup that shows the indoor route
    - ItineraryDetails panel that can be opened to show indoor route
      - Icons showing vertical direction
      - Text that can show vertical transportation use type, vertical direction, or destination level for elevators
    - Indoor routing capability in stop and near you views (small range walk query is made in these based on current location)
    - Navigator changes to show indoor route
    - Improved navigator CSS layout
image


app/component/itinerary/IndoorRouteInfo.js
```diff
@@ -0,0 +1,78 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import { FormattedMessage } from 'react-intl';
+import { configShape } from '../../util/shapes';
+import { isKeyboardSelectionEvent } from '../../util/browser';
+import Icon from '../Icon';
+
+export default function IndoorRouteInfo(
```
vesameskanen: In UI context, Route refers to transit lines such as tram 7 or bus 500.  Users search itineraries, not routes. In finnish we have to share the word Reitti for many purposes but in code we can avoid that.

I have a feeling that dropping Route from the names would  be good:
- IndoorInfo
- IndoorStep
- IndoorSteps or IndoorLeg
- navigator/IndoorButton

etc.

app/component/itinerary/IndoorRouteStep.js
```diff
@@ -0,0 +1,135 @@
...
+        /* webpackChunkName: "dotted-line" */ `../../configurations/images/default/dotted-line.svg`
+      ),
+      import(
+        /* webpackChunkName: "indoor-dotted-line" */ `../../configurations/images/default/indoor-dotted-line.svg`
+      ),
+    ]).then(([defaultImageUrl, insideImageUrl]) => {
+      setDefaultBackgroundImageUrl(`url(${defaultImageUrl.default})`);
+      setIndoorBackgroundImageUrl(`url(${insideImageUrl.default})`);
+    });
+  }, []);
```
vesameskanen: This convention you copied from old code looks terrible. It really makes no sense to lazy load tiny svg snippets one by one, and the code is far too verbose and complicated. I believe it is possible to convert svg entry dynamically to css compatible form, or at least use one of the methods here:

https://www.svgbackgrounds.com/how-to-add-svgs-with-css-background-image/

Url approach still fetches images one by one but at least looks simple. Embedding the data as part of css is probably the best way.

app/component/itinerary/WalkLeg.js
```diff
@@ -17,17 +17,27 @@ import {
...
-import { subwayTransferUsesSameStation } from '../../util/indoorUtils';
 import { displayDistance } from '../../util/geo-utils';
 import { durationToString } from '../../util/timeUtils';
 import { splitStringToAddressAndPlace } from '../../util/otpStrings';
 import VehicleRentalLeg from './VehicleRentalLeg';
+import IndoorRouteInfo from './IndoorRouteInfo';
+import {
+  subwayTransferUsesSameStation,
+  getIndoorRouteLegType,
+  getIndoorStepsWithVerticalTransportationUse,
```
vesameskanen: Somehow postfix 'Use' at the end does not help but makes things harder to understand.

app/component/itinerary/WalkLeg.js
```diff
@@ -75,8 +87,22 @@ function WalkLeg(
...
+    leg,
+    nextLeg,
+  );
+
+  useEffect(() => {
+    // If there is only one indoor routing step, always show it.
+    if (indoorRouteSteps.length === 1) {
+      setShowIntermediateSteps(true);
+    }
+  }, [indoorRouteSteps]);
```
vesameskanen: Do not use this kind of effect. getIndoorStepsWithVerticalTransportationUse regenerates new array in each call and effect triggers at every render. I suggest initializing showIntermediateSteps state directly with one single call to getIndoorStepsWithVerticalTransportationUse.

app/util/indoorUtils.js
```diff
@@ -6,3 +13,158 @@ export function subwayTransferUsesSameStation(prevLeg, nextLeg) {
...
+}
+
+export function getIndoorStepsWithVerticalTransportationUse(
+  previousLeg,
+  leg,
+  nextLeg,
+) {
+  return getIndoorSteps(previousLeg, leg, nextLeg).filter(step =>
+    // eslint-disable-next-line no-underscore-dangle
+    isVerticalTransportationUse(step?.feature?.__typename),
```
vesameskanen: Is it really possible that step list contains null entries?  The same applies to step.feature - can it be null and  what happens when rendering code gets such steps?

If so I suggest filtering invalid steps here so that the rest of the  code can trust them.

app/util/indoorUtils.js
```diff
@@ -6,3 +13,158 @@ export function subwayTransferUsesSameStation(prevLeg, nextLeg) {
...
+    verticalDirection &&
+    verticalDirection !== VerticalDirection.Unknown &&
+    type !== IndoorRouteStepType.ElevatorUse
+      ? `-${verticalDirection.toLowerCase()}`
+      : ''
+  }`;
+}
+
+export function getEntranceWheelchairAccessibility(leg) {
+  return leg?.steps?.find(
```
vesameskanen: I'd expect leg is not null. Let the code crash if somebody does not know how to use this func.

app/util/indoorUtils.js
```diff
@@ -6,3 +13,158 @@ export function subwayTransferUsesSameStation(prevLeg, nextLeg) {
...
+    `${type
+      ?.toLowerCase()
+      .replace('use', '')}-${verticalDirection.toLowerCase()}${
+      filled ? '-filled' : ''
+    }`
+  ];
+}
+
+export function getEntranceObject(previousLeg, leg) {
+  const entranceObjects = leg?.steps
```
vesameskanen: leg is not null, no need for ?

app/util/clusterUtils.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * Create Features for a given list of objects
+ */
+export function createFeatureObjects(objects) {
+  return objects.map(object => ({
+    type: 'Feature',
+    properties: { ...object.properties },
```
vesameskanen: Why do you copy the object?

vesameskanen (changes requested):
Indoor step rendering is fragmented especially in NaviCardExtension and quite hard to follow. Would it be possible to move most required rendering code and logic to IndoorContainer class? So, move leg rendering would be something like:

if (showIndoorRoute) {
    <IndoorContainer ...>
    { Steps.length === 1 && stopInformation() }    
}

app/component/itinerary/IndoorRouteInfo.js
VillePihlava: I refactored the names to not include route

app/component/itinerary/IndoorRouteStep.js
VillePihlava: I moved the dotted lines into the theme directories and to be configured with CSS, it seems to work

VillePihlava: but yes the old way was very ugly

app/component/itinerary/WalkLeg.js
VillePihlava: I removed the postfix for this method only, for the others I think it is important context

app/component/itinerary/WalkLeg.js
VillePihlava: I changed it to the suggestion

app/util/indoorUtils.js
VillePihlava: I refactored indoorUtils and removed many unnecessary question marks, a lot of these have been gathered from old code so I didn't bother checking the use of ? before

app/util/indoorUtils.js
VillePihlava: agreed

app/util/indoorUtils.js
VillePihlava: agreed

app/util/clusterUtils.js
VillePihlava: this was copied based on the old diff for the `RentalVehicles` file changed in this PR, I'm not sure why the old code did it that way, but I now changed it to simply: `properties: object.properties`

app/util/indoorUtils.js
VillePihlava: The feature can be null for a step if the step does not have a feature. A feature can be e.g. an entrance, stairs, an escalator, or an elevator

app/component/itinerary/navigator/NaviInstructions.js
```diff
@@ -33,7 +33,16 @@ function getBoardingParams(leg, time, config) {
...
-  { leg, nextLeg, instructions, legType, time, position, tailLength },
+  {
+    leg,
+    nextLeg,
+    instructions,
+    legType,
+    time,
+    position,
+    tailLength,
+    showIndoor,
```
vesameskanen: This prop should be renamed. These instructions are not indoor related. Call the prop as 'showDestinationInfo or something which describes what happens.

app/component/itinerary/navigator/NaviCardExtension.js
```diff
@@ -136,6 +160,18 @@ const NaviCardExtension = (
     );
   }
   if (legType === LEGTYPE.MOVE && nextLeg?.transitLeg) {
+    if (showIndoor) {
+      return (
+        <NaviIndoorCard
+          showIndoor={showIndoor}
```
vesameskanen: showIndoor prop for NaviCard is fake. Card is never rendered if showIndoor is false.

vesameskanen (changes requested):
Thanks for imporvements!

I made a couple of minor remarks. Also, please check the new svg sprite sheet entries - some icon content is wrapped in `<g>` block without any reason. Just remove those.

app/component/itinerary/navigator/NaviInstructions.js
VillePihlava: I changed the prop here to the suggestion. `showIndoor` itself is a bad name and I changed it back to `showIndoorRoute` in the other places where it is used, because that is literally what it does. I think the route part in this case is not misleading

app/component/itinerary/navigator/NaviCardExtension.js
VillePihlava: I removed it, the toggle function still has to be passed

vesameskanen (approved):

PR #5598, opened 2025-12-11
Fix: incorrect platform change card when changing vehicle

sharhio:
## Proposed Changes

  - reset previous platform value between leg changes

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/itinerary/navigator/NaviCardContainer.js
+++ b/app/component/itinerary/navigator/NaviCardContainer.js
@@ -69,12 +69,11 @@ function NaviCardContainer(
   const focusRef = useRef(false);
 
   const { intl, config, match, router } = context;
-
-  const nextLegPlatformCode = nextLeg?.from?.stop?.platformCode;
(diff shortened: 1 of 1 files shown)


vesameskanen (changes requested):
Incorrect warning still shows up occasionally - got it 4 times during 20 tests. Maybe the  logic below is more obvious:
- Do not evaluate change status at all unless next leg is a transit leg
- RESTORED status can't be set unless CHANGED status is detected first

```
const platformRef = useRef();

  if (legChanged) {
    platformRef.current = undefined;
  }
  let platformStatus = PLATFORM_STATUS.NORMAL;
  if (nextLeg?.transitLeg) {
    platformStatus = getPlatformChangeStatus(nextLeg, platformRef.current);
    if (platformStatus === PLATFORM_STATUS.CHANGED) {
      platformRef.current = nextLeg.from.stop.platformCode;
    }
  }
```

vesameskanen (approved):

PR #5601, opened 2025-12-15
AB#173 Filter Traffic now results by vehicle mode

partisaani:
## Proposed Changes
Adds a feature to filter Traffic now -results by vehicle mode
image

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/trafficnow/filters/FiltersContext.js
+++ b/app/component/trafficnow/filters/FiltersContext.js
@@ -5,7 +5,7 @@ const FilterContext = createContext();
 
 const DEFAULT_FILTERS = {
   validityPeriod: 'ALL',
-  vehicleMode: [],
+  vehicleModes: [],
 };
 
 const FilterContextProvider = ({ children }) => {
--- a/app/component/trafficnow/filters/FiltersContext.js
+++ b/app/component/trafficnow/filters/FiltersContext.js
@@ -5,6 +5,7 @@ const FilterContext = createContext();
 
 const DEFAULT_FILTERS = {
   validityPeriod: 'ALL',
+  vehicleMode: [],
 };
 
 const FilterContextProvider = ({ children }) => {
--- a/app/component/trafficnow/filters/VehicleModeFilter.js
+++ b/app/component/trafficnow/filters/VehicleModeFilter.js
@@ -1,0 +1,62 @@
+import React from 'react';
+import { FormattedMessage } from 'react-intl';
+import PropTypes from 'prop-types';
+import { useFilterContext } from './FiltersContext';
+import { useConfigContext } from '../../../configurations/ConfigContext';
+import { getTransportModes } from '../../../util/modeUtils';
+
+const VehicleModeFilter = ({ filterId }) => {
+  const config = useConfigContext();
+  const { selectedFilters, setFilter } = useFilterContext();
+
+  const handleCheck = option => {
+    const checked = selectedFilters[filterId] || [];
+
+    if (checked.includes(option)) {
+      setFilter(
+        filterId,
+        checked.filter(c => c !== option),
+      );
+    } else {
+      setFilter(filterId, [...checked, option]);
+    }
+  };
+
+  const availableModes = Object.entries(getTransportModes(config)).reduce(
+    (acc, [k, v]) => {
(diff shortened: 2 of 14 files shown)


app/component/trafficnow/filters/FiltersContext.js
```diff
@@ -5,6 +5,7 @@ const FilterContext = createContext();
 
 const DEFAULT_FILTERS = {
   validityPeriod: 'ALL',
+  vehicleMode: [],
```
vesameskanen: I guess this should be in plural ( vehicleModes ) as the value is an array.

app/component/trafficnow/filters/VehicleModeFilter.js
```diff
@@ -0,0 +1,62 @@
...
+        filterId,
+        checked.filter(c => c !== option),
+      );
+    } else {
+      setFilter(filterId, [...checked, option]);
+    }
+  };
+
+  const availableModes = Object.entries(getTransportModes(config)).reduce(
+    (acc, [k, v]) => {
```
vesameskanen: I guess we should limit the modes to the set of GTFS transit modes, and expand later if there will be an alert  data source for city bikes, taxi, scooters etc. There seems to be a util function getTransitModes, but it does not remove taxi from the mode set, though.

app/component/trafficnow/filters/filterUtils.js
```diff
@@ -11,9 +11,23 @@ const validityPeriodFilter = (alert, selectedFilters) => {
...
-    filterFns.every(fn => fn(alert, selectedFilters)),
+const vehicleModeFilter = ({ entities }, { vehicleMode }) => {
+  const modes = vehicleMode || [];
+  return (
+    modes.length === 0 ||
+    entities.some(e => {
+      const mode = e.vehicleMode?.toLowerCase() || e.mode?.toLowerCase();
+      if (e.locationType === 'STOP' && e.platformCode) {
+        return modes.includes('rail');
+      }
```
vesameskanen: I don't think platformCode indicates rail mode. Better to consider the mode as unknown. I found tram alerts when train filter is set in HSL planner.

app/component/trafficnow/utils.js
```diff
@@ -1,11 +1,42 @@
...
+    return stopOrRoute.vehicleMode.toLowerCase();
+  }
+
+  // If it's a stop with platformCode, assume rail
+  if (
+    stopOrRoute.locationType === LocationTypes.STOP &&
+    stopOrRoute.platformCode
+  ) {
+    return TransportMode.Rail.toLowerCase();
+  }
```
vesameskanen: Same as above - platformCode does not indicate rail mode.

vesameskanen (changes requested):
Nice proceedings again. In addition to the minor code remarks I made, there is one thing to improve: when an alert concerns many entities, they should be presented in sorted order. It is difficult to detect a single route from a list of 100.

app/component/trafficnow/filters/VehicleModeFilter.js
vesameskanen: Actually getTransitModes is buggy, it should filter taxi as well.  Now it filters when a mode is not set by default which is really fuzzy logic. 

Maybe we need a list of supported transit modes in constants.js.

app/component/trafficnow/filters/filterUtils.js
partisaani: Good to know. This was an attempt to alleviate an issue with Tikkurila station as it was clearly a train related stop but the only thing hinting towards that was a `platformCode`.

vesameskanen (changes requested):
Most remarks got fixed, but some more left to do:

- TrafficNowTransportModes contains citybikes, which is wrong

- Filter list should be an intersection of config.transportModes and TrafficNowTransportModes. It makes no sense to show subway filter in Rovaniemi.

- In hämeenlinna, there was an alert with lots of unknown entity types and a few stopOnRoute types. Alert got filtered when bus filter was set although stopOnRoute referred bus  routes. This feels a bit wrong.

image

vesameskanen (approved):

PR #5608, opened 2025-12-29
AB#174 General Traffic now improvements

partisaani:
## Proposed Changes
- Added missing translations
- Added useTranslationContext hook to be used anywhere while preserving backwards compatibility
- Filter related elements stick to the top of viewport when scrolling
- General improvements and additions based on stakeholder feedback

## Pull Request Check List

## Review

  - Merge the pull request


app/component/trafficnow/Header.js
```diff
@@ -4,15 +4,69 @@ import Link from 'found/Link';
...
+const AdditionalDescription = () => {
+  const intl = useTranslationsContext();
+  const {
+    URL: { HOLIDAYS_AND_EXCEPTIONS, MAJOR_CHANGES },
+  } = useConfigContext();
+
+  const lang = document.cookie
+    .split(';')
+    .find(item => item.indexOf('lang=') > -1)
+    .split('=')[1];
```
partisaani: Is there a better way to get a hold of the current language?

app/configurations/config.hsl.js
```diff
@@ -80,6 +80,16 @@ export default {
...
+    HOLIDAYS_AND_EXCEPTIONS: {
+      fi: 'https://www.hsl.fi/matkustaminen/juhlapyhat-ja-poikkeusaikataulut',
+      sv: 'https://www.hsl.fi/sv/att-resa/helger-och-avvikande-tidtabeller',
+      en: 'https://www.hsl.fi/en/travelling/bank-holidays-and-changes-to-public-transport-services',
+    },
+    /* MAJOR_CHANGES: {
+      fi: 'TODO',
+      sv: 'TODO',
+      en: 'TODO',
+    }, */
```
partisaani, line 92: It is currently unclear where this "suuret muutokset" link that was requested to be implemented should take the user

app/component/trafficnow/Header.js
vesameskanen: You should use app/store/PreferencesStore, where the current language is stored: 

```

function component (props, context) {
      const lang = context.getStore('PreferencesStore').getLanguage();
}


```

Actually the current way presented above is some sort of mistake.  Stores are used for emitting changes which trigger component rendering.  Lang does not change dynamically.  UI server sends a cookie at first page load. If user  changes the language, new page loading takes place.  We could simply store the language to configuration or use a react context.

app/util/useTranslationsContext.js
```diff
@@ -0,0 +1,25 @@
...
+);
+
+IntlContextProvider.propTypes = {
+  intl: intlShape.isRequired,
+  children: PropTypes.node.isRequired,
+};
+
+export const useTranslationsContext = () => {
+  const intl = useContext(IntlContext);
+  if (!intl) {
```
vesameskanen, line 18: I am not sure why you add new context providers? The old way of accesing config or intl from standard context looks so simple:

```

function mycomponent(props, context) {
   const intl = context.intl;
}


```

vesameskanen (changes requested):

app/component/trafficnow/Header.js
vesameskanen: You should use app/store/PreferencesStore, where the current language is stored:


function component (props, context) {
      const lang = context.getStore('PreferencesStore').getLanguage();
}


Actually the current way presented above is some sort of mistake. Stores are used for emitting changes which trigger component rendering. Lang does not change dynamically. UI server sends a cookie at first page load. If user changes the language, new page loading takes place. We could simply store the language to configuration.

app/util/useTranslationsContext.js
hjvuor: The old way of accessing context is deprecated in React versions newer than 16, so I think this is a welcoming change

partisaani: Futureproofing as @hjvuor said.

Accessing the context via a provider is also simpler in deeply nested components as there's no need to drill down the context prop.

vesameskanen (approved):
Nice - acessing language is now so simple!

PR #5610, opened 2025-12-31
feat(map): add 360-degree map rotation support

kaziNymul:
Add Google Maps-style map rotation using leaflet-rotate plugin:
- Mobile: Two-finger touch gesture to rotate
- Desktop: Shift + mouse drag to rotate

The feature is configurable via config.map.allowRotation (default: true) and can be disabled per-theme if needed.

Changes:
- Add leaflet-rotate dependency
- Enable rotation props in Map.js component
- Add allowRotation config option in config.default.js

## Proposed Changes

  -

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/map/Map.js
+++ b/app/component/map/Map.js
@@ -142,6 +142,11 @@ export default class Map extends React.Component {
 
   componentDidMount() {
     this.mounted = true;
+    // Load leaflet-rotate plugin dynamically to enable map rotation
+    // This avoids ESM/CommonJS compatibility issues in test environment
+    import('leaflet-rotate').catch(() => {
+      // Silently ignore in test environment where ESM modules may not load
+    });
     if (this.props.mapLayers.vehicles) {
       startClient(this.context);
     }
@@ -347,6 +352,11 @@ export default class Map extends React.Component {
             {...leafletEvents}
             onPopupopen={onPopupopen}
             closePopupOnClick={false}
+            rotate={config.map.allowRotation}
+            rotateControl={false}
+            touchRotate={config.map.allowRotation}
+            shiftKeyRotate={config.map.allowRotation}
+            bearing={0}
           >
             <TileLayer
               url={mapUrl}
--- a/app/configurations/config.default.js
+++ b/app/configurations/config.default.js
@@ -304,6 +304,9 @@ export default {
     showLayerSelector: true,
     showStopMarkerPopupOnMobile: true,
     showScaleBar: true,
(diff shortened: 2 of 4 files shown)


kaziNymul:
@Antiik91 @Jontzii @optionsome @vesameskanen - Could you please review this feature addition? It adds 360-degree map rotation similar to Google Maps. Thanks!

vesameskanen:
Thanks for contribution.  Some comments:
- What is the use case for map rotation by a gesture? It might cause bad user experience when the map gets rotated accidentally, especially because leaflet raster map does not keep rotated labels readable (they may appear upside down). Therefore, the option cannot be enabled by default.
- We do have new feature development which will benefit from automatic map rotation. The map should be oriented by mobile device sensors. This will help to read the map  while walking on a street.  In order to allow this, we will replace leaflet with a new vector map engine (e.g. MapLibre) at some point. New map will be able to keep labels readable on a rotating map. Considering the upcoming map engine change, do you think that merging this PR makes sense?

kaziNymul:
Thank you for the detailed feedback and sharing your roadmap!

My Use Case

My primary motivation was navigation during transport transfers (e.g., switching between bus/metro/tram). When I arrive at an unfamiliar station and need to find my next platform or exit, I often need to orient the map to match my physical direction. Currently, I find myself switching to Google Maps for this.

I Understand the Concerns

You raise valid points:
Accidental rotation - I agree this can frustrate users
Upside-down labels - Raster maps don't handle rotation gracefully
Default setting - You're right, it shouldn't be enabled by default

Given Your Roadmap

Since you're planning to:
1. Replace Leaflet with MapLibre (vector maps)
2. Implement sensor-based automatic rotation
3. Have proper label rendering on rotated maps

It makes sense to wait for the proper implementation rather than merge this as a temporary solution.

My Suggestion

I'm happy to:
- Close this PR if you prefer to wait for MapLibre integration
- Or modify it to keep `allowRotation: false` by default, available only for testing/development

Thank you for considering this feature request. Looking forward to the MapLibre update! 🙂

vesameskanen:
Hi! It seems that we do not have an obvious place for 'Allow map rotation' switch. HSL planner does not expose any map related settings to the user at the moment. Therefore, I suggest closing this PR at this point.

Many thanks for constructive discussion.

kaziNymul:
Thank you for taking the time to review and discuss this feature! I appreciate the 
constructive feedback and understanding of your platform's architecture.

I'll close this PR now. Looking forward to the future MapLibre implementation with 
sensor-based rotation! 🙂

Thanks again!

PR #5623, opened 2026-01-16
AB#74-next3

vesameskanen:
- More refactoring
- Modal for selecting modes whivch do no fit into control panel


digitransit-component/packages/digitransit-component-control-panel/src/helpers/AllModesModal.js
```diff
@@ -25,11 +25,19 @@ export default function AllModesModal({
...
           {t('title', { lng: language })}
         </h2>
+        {isMobile ? (
+          ''
+        ) : (
+          <div className={styles['modal-desc']}>
+            {t('description', { lng: language })}
+          </div>
+        )}
+        <div style={{ height: '20px' }} />
```
partisaani: If `{!isMobile && <element>}` doesn't render false, I'd opt using it especially since it's already being employed a couple of rows below

digitransit-component/packages/digitransit-component-control-panel/src/helpers/NearYouButton.js
```diff
@@ -0,0 +1,107 @@
...
+          )}
+        </span>
+        {title}
+        {withArrow && (
+          <Icon img="arrow" color={colors.primary} width={0.9} height={0.9} />
+        )}
+      </span>
+    </>
+  );
+}
```
partisaani, line 81: There may be accessibility related issues in having a component that resembles a button with an actual `<button>` element nowhere in sight

partisaani (changes requested):

digitransit-component/packages/digitransit-component-control-panel/src/helpers/NearYouButton.js
vesameskanen: It was not too hard to change to a button, only text placement needed restyling

partisaani (approved):

PR #5633, opened 2026-01-26
Ab#44 Flex bus improvements

sharhio:
## Proposed Changes

  - New booking time parameter in use so that booking rules are taken into account when searching. Routes that would leave before the earliest possible pick up time will not be returned. Depends on https://github.com/opentripplanner/OpenTripPlanner/pull/7221
  - Fixes to travel companion call agency styles.
  - Route page that is returned from route search now only shows notifications for call agency routes as there is no route or schedule. The route page map shows the service area of the call agency.
  - Removed some older call agency code, which is no longer in use. 
  - Refactored the local call agency util function, and notification propTypes that caused warnings. 

## Pull Request Check List

## Review

  - Merge the pull request


app/configurations/config.hsl.js
```diff
@@ -738,6 +738,32 @@ export default {
...
+        fi: [
+          'Linja toimii ennakkotilauksella. Varmistaaksesi matkan, tee varaus etukäteen. Tarkemmat tiedot palveluntarjoalta.',
+        ],
+        en: [
+          'This service operates by advance booking. To ensure your ride, please book ahead of time. More information is available from the service provider. ',
+        ],
+        sv: [
+          'Linjen fungerar med förhandsbokning. Boka din resa i förväg för att säkerställa resan. Mer information från tjänsteleverantören.',
+        ],
+      },
```
vesameskanen, line 759: I believe this part could be inherited from the default config, no need to duplicate common parts here. Just add the links.

vesameskanen (changes requested):
For some reason, I did not get any call transport itineraries any more in Kirkkonummi area.

vesameskanen (approved):

PR #5642, opened 2026-01-30
AB#420: Change deprecated node-sass package to dart-sass

Antiik91:
## Proposed Changes

  - Deprecated Node-sass package is dependant on the old gyp version that uses python 2.7. This will cause intermittent problems with ARM processors, and will likely cause more problem with the future.
  
 - Change foundation-apps to foundation sites, because the foundation-apps (deprecated) does not support dart-sass.

- Old cheerio version is needed so our enzyme testing environment can work. (We should change enzyme to some modern testing framework, so we could remove this resolution)

# Notes

- This update will prompt some sass changes that needs to be done before dart-sass is updated to v3. 

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen (approved):
This is good step in updating ancient dependencies.

PR #5644, opened 2026-01-30
AB#74 update park cards

vesameskanen:
- Car and bike park view contents updated
- Opening hours and wheelchair accessibility is shown
- Accessibility errors fixed
- Citybike view shows route to nearest station
- Park views show route to park using correct street mode (not walking)
- New place search after moving on map updates walk/bike/car routes
- Favourites near you button is hidden if no favourites 
- Lots of refactoring


app/util/ParkAndRideUtils.js
```diff
@@ -37,26 +37,27 @@ const ParkAndRideUtils = {
...
-          const day = dateString.substring(6, 8);
-          return {
-            date: new Date(year, month - 1, day),
-            timeSpans: openingHour?.timeSpans && openingHour?.timeSpans[0],
-          };
-        });
-        return openingHoursByDay;
+      const osm = openingHours?.osm;
+      if (osm) {
+        if (osm === 'Mo-Fr 0:00-23:59; Sa 0:00-23:59; Su 0:00-23:59') {
```
partisaani, line 54: Where is this magic string originating from?

vesameskanen: OpenStreetMap  data models opening hours that way. It if unexpectedly changes, we still get a sensible value with 3 rows and translated day names.

partisaani (approved):

PR #5647, opened 2026-02-05
Accessibility fixes for near you view from AB#74

vesameskanen:
These changes fix some bad accessibility problems in V3. Needed to enable the release  next week.
- near you buttons on front page needed two tab key presses to pass, fixed
- After opening the near you page, focus was often in wrong tab. It was possibe to move the focus to a hidden tab using tab key.

### DIFF
--- a/app/component/nearyou/NearYouPage.js
+++ b/app/component/nearyou/NearYouPage.js
@@ -155,7 +155,9 @@ function NearYouPage(
   }, []);
 
   useEffect(() => {
-    window.addEventListener('keydown', tabHandler);
+    // rendering of hidden sub components of inactive tabs
+    // sets tabIndex=0 to hidden elements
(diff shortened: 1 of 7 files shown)


app/component/nearyou/NearYouPage.js
```diff
@@ -61,6 +61,12 @@ function getModes(config) {
   return modes.map(nearYouMode => nearYouMode.toUpperCase());
 }
 
+function tabHandler(e) {
+  if (e.key === 'Tab') {
+    setFocusables();
+  }
+}
+
```
partisaani, line 69: I would add a small JSDoc here to explain why is there a handler for tab key press all of a sudden

partisaani (approved):

partisaani (approved):

partisaani (approved):

PR #5651, opened 2026-02-09
Refactor  some map components

vesameskanen:
- Map.js and GeoJSON.js converted from class to function component
- Complex and redundant bottom sheet position management simplified radically
- Language is taken from config in all map components
- Component prop cleanup

### DIFF
--- a/app/component/map/GeoJSON.js
+++ b/app/component/map/GeoJSON.js
@@ -1,10 +1,10 @@
 import L from 'leaflet';
 import PropTypes from 'prop-types';
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import uniqBy from 'lodash/uniqBy';
 import { default as Geojson } from 'react-leaflet/es/GeoJSON';
 import PointFeatureMarker from './PointFeatureMarker';
-import { configShape, geoJsonFeatureShape } from '../../util/shapes';
+import { geoJsonFeatureShape, configShape } from '../../util/shapes';
 import {
   isMultiPointTypeGeometry,
   isPointTypeGeometry,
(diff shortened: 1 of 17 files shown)


app/component/map/GeoJSON.js
```diff
@@ -177,59 +160,71 @@ class GeoJSON extends React.Component {
...
+            {...rest}
+          />
+        ))}
+    </React.Fragment>
+  );
 }
 
+GeoJSON.propTypes = {
+  // eslint-disable-next-line
+  bounds: PropTypes.object,
```
partisaani: `PropTypes.shape({})` works without eslint comment

app/component/map/GeoJSON.js
```diff
@@ -177,59 +160,71 @@ class GeoJSON extends React.Component {
...
   // eslint-disable-next-line camelcase
-  UNSAFE_componentWillMount() {
-    const {
-      data: { features },
-    } = this.props;
-    this.icons = getIcons(features);
-  }
+  useEffect(() => {
+    const { features } = data;
+    setIcons(getIcons(features));
```
partisaani: useEffect is not mandatory as this could be moved into the initial state setup:

```
const [icons, setIcons] = useState(getIcons(data.features));
```

partisaani (changes requested):

app/component/map/GeoJSON.js
vesameskanen: fixed

app/component/map/GeoJSON.js
vesameskanen: I tried useRef (state setIcons would be unused, eslint starts crying) and for some reason which I don't quite undestand component crashes. Re-render caused by useEffect seems to work fine.

vesameskanen: UseRef works fine after all.

PR #5666, opened 2026-02-26
AB#450 next personalisation info modal

vesameskanen:
Adds a modal with some information about personalisation. Link to hsl.fi favourite management included. CrazyEgg survey button currently does nothing.

Some refactoring also included, for example external link now has an option for arrow icon which is often used in new designs.


app/component/itinerary/customizesearch/PrModal.js
```diff
@@ -0,0 +1,81 @@
...
+import { useTranslationsContext } from '../../../util/useTranslationsContext';
+
+export default function PrModal({ closeModal }) {
+  const config = useConfigContext();
+  const intl = useTranslationsContext();
+
+  return (
+    <Modal
+      appElement="#app"
+      contentLabel="personlaisation_info_modal"
```
sharhio: typo

sharhio (changes requested):
Looks good. There is a warning on the console about the infomodal relating to screen reader functionality, but it seems to work without issues, at least for me. 

A couple of very minor design notes:
- The space under the personalisation setting section is too small when compared to the design and the other setting sections.
- The rounding of the infobox is sharper than the design.

PR #5706, opened 2026-03-23
AB#535 - Adjust defaultSettings

VillePihlava:
## Proposed Changes

  - Adjust values:
    - Increase transfer penalty
    - Reduce walk and bike reluctance


vesameskanen (commented):
- Are the new bike settings different from OTP defaults? If not, adding them to UI config does not make a difference.
- HSL transfer penalty is already 180. Is this change needed for testing something?

vesameskanen (approved):

PR #5712, opened 2026-03-25
AB#238 - Add least transfer filtering for flex

VillePihlava:
## Proposed Changes

  - Add filtering to internal and external flex for filtering out itineraries with many transfers
  - Some refactoring related to isCallAgencyLeg


app/configurations/config.hsl.js
```diff
@@ -817,5 +811,5 @@ export default {
   },
 
   showRouteDescNotification: IS_DEV,
-  personalisation: IS_DEV,
+  personalisation: false,
```
vesameskanen: set this to IS_DEV

VillePihlava: I changed it to IS_DEV

PR #5726, opened 2026-04-07
AB#238 - Make IS_DEV util function and change configs to use it

VillePihlava:
## Proposed Changes

  - Create IS_DEV and IS_PROD util functions

### DIFF
--- a/app/util/envUtils.js
+++ b/app/util/envUtils.js
@@ -1,6 +1,18 @@
-/** Check if application is running in a dev environment. RUN_ENV is defined in kubernetes-deploy for dev instances. For running dev locally, NODE_ENV is checked * */
-export const isDevelopmentEnvironment = config => {
+/**
+ * Check if application is running in a dev environment.
+ * RUN_ENV === 'development' in digitransit-kubernetes-deploy for dev instances.
+ * For running dev locally, NODE_ENV is checked.
+ */
+export function IS_DEV() {
   return (
-    config?.RUN_ENV === 'development' || process.env.NODE_ENV === 'development'
+    process.env.RUN_ENV === 'development' ||
+    process.env.NODE_ENV !== 'production'
   );
-};
+}
+
+/**
+ * Check if application is running in a prod environment.
+ */
+export function IS_PROD() {
+  return !IS_DEV();
+}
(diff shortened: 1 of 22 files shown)


VillePihlava:
@vesameskanen what should be done to the environment checks that only use `NODE_ENV`? For example:
https://github.com/HSLdevcom/digitransit-ui/blob/00a43b2142edd363203ca44d450227311c690745/app/client.js#L99

app/util/envUtils.js
```diff
@@ -1,6 +1,18 @@
...
+    process.env.RUN_ENV === 'development' ||
+    process.env.NODE_ENV !== 'production'
   );
-};
+}
+
+/**
+ * Check if application is running in a prod environment.
+ */
+export function IS_PROD() {
```
tkalvas: There's a change here to avoid the question of whether staging environments are test environments by just having IS_PROD and use either IS_PROD() or !IS_PROD() all over.

VillePihlava: Maybe we should just have one IS_DEV function and remove the IS_PROD function that I added. The question becomes more complicated when taking into account that we have two environment variables that indicate the dev environment. What happens when we only want to take one of them into account. See my comment above

VillePihlava: or only IS_PROD

tkalvas: I'm pretty sure we should not be using two different logics at all. Why would we want to only take one of the variables into account?

VillePihlava: I agree, this seems pretty weird. One of the variables comes from the Kubernetes deployment and one is set in the Dockerfile. Lets see what Vesa says.

tkalvas: I get the two variables thing, but I think every place in the code should use the exact same logic, using the two variables (through a helper function). The fact that some places only check one is utterly confusing.

vesameskanen (changes requested):
It seems that IS_DEV is used only as a feature switch. I guess the minimal implementation is to set it to a config field in config.default.js. Other configs will then inherit it. Usage will be easy: `if(config.is_dev) ....` or `this.is_dev` .
The value usually comes from kube deployment, but it has to be set as env value in package.json's  'yarn dev' script as well so that it works in local ui server.

There is already RUN_ENV field in config but it seems mostly unused. I suggest removing it and using simpler is_dev boolean instead.

NODE_ENV is more complicated as some build tools use it, so we can leave it as it is now. 
 
IS_PROD() is never used, I suggest removing it.

vesameskanen (approved):

PR #5739, opened 2026-04-10
AB#532 - Update to node v24

VillePihlava:
## Proposed Changes

  - Update node to v24.14.1
  - I tested running the UI and unit tests and it worked
  - I removed an old node14-patch

## Problems
  - A lot of new warnings
  - babel-esm-loader.mjs is workaround for mocha ESM syntax problems
  - Not everything has been tested so some things might break in the future


vesameskanen (approved):
Seems to work fine. I am surprised that leaflet patch can be removed now.

vesameskanen (approved):

test/unit/helpers/babel-esm-loader.mjs
```diff
@@ -0,0 +1,47 @@
...
+
+  const source =
+    typeof loaded.source === 'string'
+      ? loaded.source
+      : Buffer.from(loaded.source).toString('utf8');
+
+  const transformed = await transformAsync(source, {
+    filename: fileURLToPath(url),
+    caller: {
+      name: 'mocha-esm-loader',
```
tkalvas: 'babel-esm-loader'

tkalvas (changes requested):

PR #5745, opened 2026-04-13
Refactor font styles

vesameskanen:
- Use consistently variable defined font weights, not hard coded values
- Remove too specialized weight variables
- Remove some sass variables which just duplicate another variable 
- Remove unnecessary font-family settings because inheritance works fine


vesameskanen:
Can be merged after next UI release

PR #5746, opened 2026-04-14
AB#450  use HSL design system modal for personalisation info

vesameskanen:
https://www.figma.com/design/4axMbTogOGOqIXeZNt6suk/Personalisation?t=XHTSMzeIYyDPo1y0-0

Note: to test, set personalisation=true in hsl config. Personalisation requires login, so other themes do not support it yet.


sharhio (approved):
LGTM but a minor conflict needs to be resolved.

PR #5751, opened 2026-04-15
Convert IS_DEV function to a constant

hjvuor:
## Proposed Changes

  - Environment variables wont change during runtime so no need to call IS_DEV as a function each time it is needed

## Pull Request Check List

## Review

  - Merge the pull request


VillePihlava (approved):
Good catch, should have done it like this originally.

PR #5753, opened 2026-04-16
Refactor translation system

vesameskanen:
- Get rid of complicated and unnecessary StoreListeningIntlProvider.js
- Split translations to one file per language
- Load only one single language bundle for the current language
- Refactor message bar related utils to use new single language localization
- Remove about 200 unused translations per language
- Remove outdated complicated polyfill code
- Remove some unused styles
- Update unit tests


test/unit/translations.test.js
```diff
@@ -1,27 +1,29 @@
...
  * from the translations file altogether.
  */
 const ignoredTerms = [
-  'destination-label-change',
-  'origin-label-change',
-  'splash-welcome',
+  //  'destination-label-change',
+  //  'origin-label-change',
+  //  'splash-welcome',
 ];
```
VillePihlava, line 17: Should this just be removed?

server/passport-openid-connect/openidConnect.js
```diff
@@ -135,7 +135,7 @@ export default function setUpOIDC(app, port, indexPath, hostnames) {
         secure: process.env.NODE_ENV === 'production',
         httpOnly: process.env.NODE_ENV === 'production',
         maxAge: 1000 * 60 * 60 * 24 * 60,
-        sameSite: 'none',
+        sameSite: process.env.NODE_ENV === 'production' ? 'none' : 'lax',
```
VillePihlava, line 138: What does `lax` do?

app/util/geolocationMessages.js
VillePihlava, line 5: Should we resolve this TODO while we're here?

app/util/geolocationMessages.js
```diff
@@ -26,35 +24,26 @@ const events = {
...
-          type: s,
-          content: section,
-        });
-      }
+      message.content[language].push({
+        type: s,
+        content: translations[key],
+      });
     });
+    geolocationMessages[e] = message;
```
VillePihlava: I don't like the mutable state here. How much work would it take to just use the return value from this function?

VillePihlava: In that case the function should be renamed to `getGeolocationMessages`

app/util/messageUtils.js
```diff
@@ -9,49 +9,51 @@ export const favouriteTypes = [
...
+      messages[key] = {
+        id: key,
+        persistence: 'repeat',
+        priority: 4,
+        icon: 'caution_white_exclamation',
+        iconColor: '#dc0451',
+        backgroundColor: '#fdf0f5',
+        type: 'error',
+        content,
+      };
```
VillePihlava: Again, mutable state vs return value.

app/util/messageUtils.js
```diff
@@ -9,49 +9,51 @@ export const favouriteTypes = [
...
-    backgroundColor: '#fdf0f5',
-    type: 'error',
-    content,
-  };
+export function failedFavouriteMessage(type, isSave) {
+  const t = favouriteTypes.includes(type) ? type : favouriteTypes[0];
+  const key = isSave
+    ? `add-favourite-${t}-failed-heading`
+    : 'delete-favourite-failed-heading';
+  return messages[key];
```
VillePihlava: If the messages are e.g. put in the global context, then this could be a util function that takes the current set of messages as a parameter

VillePihlava (changes requested):
- Can you rename the `intl` folder to `translations`? Makes a bit more clearer. If theres a good reason to name it `intl` then its fine.
- I did not look at individual translations files.
- We should update to react-intl version 3. It should be compatible with react version 16. This probably removes a lot of code. I will rereview once this has been done.

server/passport-openid-connect/openidConnect.js
vesameskanen: Janne's fix from next to make local UI login work.

app/util/geolocationMessages.js
vesameskanen: These messages are a global resource which is not used from react components. Messages are built here once.

We could, of course, pass translations and current language to this file and build each message dynamically when needed with a 'getMessage' func, but we still niin the init call to store language and translations for later use.

PR #5762, opened 2026-04-21
Miscellaneous minor fixes

vesameskanen:
- Update axios
- Add back one accidentally lost translation
- Clean up LangSelect
- Simplify language config


app/component/LangSelect.js
```diff
@@ -25,36 +24,21 @@ const language = (lang, highlight, match, intl) => {
...
 
-LangSelect.displayName = 'LangSelect';
-
-LangSelect.propTypes = {
-  currentLanguage: PropTypes.string.isRequired,
-};
-
 LangSelect.contextTypes = {
-  executeAction: PropTypes.func.isRequired,
   config: configShape.isRequired,
```
hjvuor: config can be removed from contextTypes

app/component/LangSelect.js
```diff
@@ -1,9 +1,8 @@
-import PropTypes from 'prop-types';
 import React from 'react';
-import { matchShape, routerShape } from 'found';
+import { matchShape } from 'found';
 import { useIntl } from 'react-intl';
-import connectToStores from 'fluxible-addons-react/connectToStores';
 import { configShape } from '../util/shapes';
+import { useConfigContext } from '../configurations/ConfigContext';
 
 const language = (lang, highlight, match, intl) => {
```
hjvuor: isn't this essentially a functional component? maybe the name should be capitalized and used with jsx syntax instead of a function call

hjvuor (changes requested):

app/component/LangSelect.js
VillePihlava: Can this be removed now as well?

app/component/LangSelect.js
```diff
@@ -25,36 +24,21 @@ const language = (lang, highlight, match, intl) => {
...
   match: matchShape.isRequired,
 };
 
-const connected = connectToStores(
-  LangSelect,
-  ['PreferencesStore'],
-  context => ({
-    currentLanguage: context.getStore('PreferencesStore').getLanguage(),
-  }),
-);
```
VillePihlava, line 58: I don't have a good understanding on how these stores work, so I trust you that this should be removed

VillePihlava (commented):
Just a small comment. Also, I would rename LangSelect to LanguageSelect. I think there is no need to use abbreviations here and it makes what the class does more clear.

PR #5775, opened 2026-04-24
Convert front page to a function

vesameskanen:
- React class to function component
- Access intl, config, router and match using react hooks
- Refactoring and cleanup


app/component/IndexPage.js
```diff
@@ -46,99 +46,80 @@ import {
...
-      this.pendingDestination &&
-      isEqual(this.pendingDestination, destination)
+      pendingDestinationRef.current &&
+      isEqual(pendingDestinationRef.current, destination)
     ) {
-      delete this.pendingDestination;
+      pendingDestinationRef.current = null;
     }
-    if (this.pendingOrigin || this.pendingDestination) {
-      // not ready for navigation yet
```
VillePihlava, line 126: I think this comment can be left here above the if statement for clarity

app/component/IndexPage.js
```diff
@@ -46,99 +46,80 @@ import {
...
     const { location } = match;
 
     const currentLocation =
       config.startSearchFromUserLocation &&
-      !this.props.origin.address &&
-      this.props.locationState?.hasLocation &&
-      this.props.locationState;
+      !origin.address &&
+      locationState?.hasLocation &&
+      locationState; // assign locationState conditionally
```
VillePihlava: I think its better to put comments above the line instead of appending them to the end of a statement. This is also the convention in OTP

app/component/IndexPage.js
```diff
@@ -46,99 +46,80 @@ import {
...
     }
+
     scrollTop();
-  }
+  }, []);
 
-  componentDidUpdate() {
-    const { origin, destination } = this.props;
+  useEffect(() => {
+    const { origin, destination, locationState } = props;
```
VillePihlava, line 94: I think this should either be in the outer scope or the calls should be props.*. For example, the dependency array uses props.*

app/component/IndexPage.js
```diff
@@ -198,36 +180,32 @@ class IndexPage extends React.Component {
...
 
-  NearStops() {
-    const { intl, config } = this.context;
-    const { colors, fontWeights, language } = config;
+  const renderNearStops = () => {
+    const nearYouModes = getNearYouModes(config, props.favourites);
 
-    const nearYouModes = getNearYouModes(config, this.props.favourites);
-
-    // If nearYouModes is configured, display those. Otherwise, display all configured transport modes
```
VillePihlava, line 230: I think this comment can be left. Its not instantly clear that the fallback is all configured transport modes

app/component/IndexPage.js
```diff
@@ -280,193 +257,213 @@ class IndexPage extends React.Component {
...
+    fontWeights,
+    colors,
+    modeSet: iconModeSet,
+    geocodingSize: 25,
+  };
 
-    if (config.stopSearchFilter) {
-      stopRouteSearchProps.filterResults = results =>
-        results.filter(config.stopSearchFilter);
-      stopRouteSearchProps.geocodingSize = 40; // increase size to compensate filtering
```
VillePihlava, line 358: Again comment can be left above the corresponding line

app/component/IndexPage.js
```diff
@@ -509,7 +505,7 @@ const IndexPageWithStores = connectToStores(
     newProps.origin = origin;
     newProps.destination = destination;
     newProps.currentTime = context.getStore('TimeStore').getCurrentTime();
-    newProps.query = query; // defines itinerary search time & arriveBy
```
VillePihlava, line 512: Again, comment can be left

VillePihlava (commented):
A couple of comments

app/component/IndexPage.js
vesameskanen: There seems to be a scope problem. Origin and destination are used as variable names in some other places. I think it is best to leve the code as it is.

VillePihlava (approved):

PR #5792, opened 2026-05-04
Ab#483 Route page improvement

sharhio:
## Proposed Changes

  - Fixed agency name fading away unnecessarily
  - Fixed stop menu disappearing under a misplaced component
  - Updated another few route page components to functional
  - Refactored route page and schedule page tests for consistent implementation

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/AgencyInfo.js
+++ b/app/component/AgencyInfo.js
@@ -10,9 +10,7 @@ function AgencyInfo({ agencyName, url }) {
     return (
       <div className="agency-link-container">
         <ExternalLink className="agency-link" href={link}>
-          <div className={agencyName.length > 30 ? 'overflow-fade' : ''}>
-            {agencyName}
-          </div>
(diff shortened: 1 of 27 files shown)


vesameskanen (commented):
Code looks good, but I found a couple of issues in testing:
- Stop page header should have some gap below the title. I guess this broke because of some card component style changes.
- P train seems to have 2 directions (maybe a data issue?), whereas I train variants are classified as expected.

sharhio:
I fixed the stop page margin. 

The issue with P train seems to be caused by track work - there are two directions currently for P train, because it changes direction in Myyrmäki. The HSL data looks good, but Matka data looks a bit strange, as in there are multiple routes with different schedules. This data difference can be seen in production, too.

vesameskanen (approved):

PR #5795, opened 2026-05-06
AB#503 - Uber changes next

VillePihlava:
## Proposed Changes

  - Finishing touches for Uber changes
    - small changes to 2/3 analytics events (last one coming at some point)
    - refactorings related to some util functions
    - change translations to work with taxis in e.g. itinerarylist
      - strip away zone parenthesis block
      - use 'location' instead of 'stop' with taxis
    - move navigator taxi start prevention into correct place
    - add translations

### DIFF
--- a/app/util/modeUtils.js
+++ b/app/util/modeUtils.js
@@ -397,82 +397,95 @@ export function toggleTransportMode(transportMode, config) {
   return modes;
 }
 
+const TRACK_OR_PIER_OR_PLATFORM_TEXT_SHORT_MSGS = {
+  [TransportMode.Rail]: { id: 'track', defaultMessage: 'Track' },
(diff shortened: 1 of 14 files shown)


app/util/modeUtils.js
```diff
@@ -520,3 +520,24 @@ export function getTerminalOrStationText(intl, mode) {
     defaultMessage: 'Station',
   });
 }
+
+export function getFirstDepartureStopTypeText(intl, mode) {
```
vesameskanen, line 522: Using a simple mode->id map (defined outside the function)  instead of repeating if's would make this much nicer.

vesameskanen (changes requested):

app/util/modeUtils.js
VillePihlava: This should be fixed now for the other similar functions in the file as well

vesameskanen (approved):

PR #5807, opened 2026-05-13
AB#503 - More Uber changes

VillePihlava:
## Proposed Changes

  - Remove `stripFlexZoneInfo` because its a hack for weird data
  - Add `ItineraryImpression` analytics event for taxis
  - Refactor state names in `ItineraryPage`
  - Refactor description summary text in `Itinerary` and add screen reader text for taxi
  - Refactor uses of `<Duration>` to use `durationToString`
    - Logic of `durationToString` has changed a little bit (now uses locale formatting AND does not skip '0 min' when 0 minutes, string looks like e.g. '2h' instead of '2 h 0 min')
   - Refactor `ItineraryLine` to function component
   - Related to [AB#521](https://dev.azure.com/digitransit/49576af7-d55f-4602-9722-2d62a5d321b2/_workitems/edit/521)


app/component/map/ItineraryLine.js
```diff
@@ -40,6 +40,7 @@ import {
 class ItineraryLine extends React.Component {
   static contextTypes = {
     config: configShape.isRequired,
+    intl: PropTypes.object.isRequired, // eslint-disable-line
```
vesameskanen: Why not useIntl() ?

app/component/itinerary/ItineraryPage.js
```diff
@@ -1159,11 +1167,34 @@ export default function ItineraryPage(props, context) {
...
+        });
+      }
+    }
+  }, [
+    combinedState,
+    relaxState,
+    combinedExternalRelaxState,
+    settingsState.settingsChanged,
+  ]);
+
```
vesameskanen: Does this generate many analytics events when new search is triggered? Should it monitor only a single plan?

vesameskanen (approved):

PR #5811, opened 2026-05-18
Ab#483 Route page refactoring and new design for v3

sharhio:
## Proposed Changes

  - 

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen (commented):
- I think it would be correct to bump versions of each library whose dependencies changed
- Disruption tab seems to always render empty although there are disruptions

vesameskanen:
What is the status of route page changes? Can they all go to prod in the next ui release 1.6 ?  This PR will change the origin icon, which probably is OK to go to prod. We should at this point sync all digitransit-component libraries from next to dev as follows:
- merge v3 to next, to get a recent autousuggest fix from Heikki
- bump autosuggest and autosuggest-panel versions in next to force npm publish when v3 gets the changes
- include all digitransit-component changes made into next in your PR so that after it gets merged, v3 and next have identical digitransit-component code

app/component/routepage/RouteAlertsContainer.js
```diff
@@ -82,11 +87,17 @@ function RouteAlertsContainer({ currentTime, route, pattern }, { config }) {
   );
 
   return (
-    <AlertList
-      showLinks={false}
-      cancelations={cancelations}
-      serviceAlerts={serviceAlerts}
-    />
+    <div
+      id="route-disruption-panel"
```
hjvuor, line 91: for some reason it's this div that causes alerts not to show

hjvuor (changes requested):

vesameskanen (approved):

PR #5830, opened 2026-05-22
AB#410 - Add airport icons to the map for stops and terminals.

VillePihlava:
## Proposed Changes

  - Lollipop icon for stops
image

  - Square airplane icon for terminals

**Note**
- Terminals have not been tested because the current airports I could find are only stops


vesameskanen (changes requested):
HSL svg-sprite is not valid XML. Please fix.

vesameskanen (approved):

PR #5855, opened 2026-06-09
AB#613 Handle gtfs entities with colons in ids

hjvuor:
## Proposed Changes

  - Adds a utility function that splits a gtfsId into a feedId and entityId
    - Does so by splitting at the index of the first colon instead of splitting into array by colons
  - replaces old functionality that is prone to fail when colons are introduced in id's

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen (approved):

vesameskanen (changes requested):
stop page/timetable tab seems to crash.

vesameskanen (approved):

PR #5858, opened 2026-06-10
AB#579 Traffic now footer

sharhio:
## Proposed Changes

  - Added HSL footer for traffic now. Extensive default content visible for now. To opt for the compact, just uncomment the SiteFooter variant parameter.
  - Added padding under the filters to avoid the filter component touching the footer, and the scroll looking uneven between the filters and the alert contents.

## Pull Request Check List

## Review

  - Merge the pull request


hjvuor (changes requested):
When the list is scrolled in mobile and some disruption is opened to show details, scroll position is maintained and so only the footer is visible initially.

hjvuor (approved):

PR #5862, opened 2026-06-12
AB#399 Traffic Now Fix minor issues with styles

hjvuor:
## Proposed Changes

  - Fixes to font sizes and margins etc..
  - Sticky header button with a dynamic shadow on mobile
  - Mobile filter modal now has sticky header and footer buttons
  - Date formatting tweaked, shows start year only when it differs from end
  - Badge icon now triangle on warnings too
  - RouteBadge margins now differ on stop and vehicle icons, added an extra margin after each group

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/trafficnow/TrafficNow.js
+++ b/app/component/trafficnow/TrafficNow.js
@@ -38,6 +38,16 @@ const TrafficNow = () => {
   const isMobileCanceledTripsView = !!mode && mobile;
   const isDetailsView = !!alertId;
 
+  const [top, setTop] = useState(true);
+
+  useEffect(() => {
+    const scrollHandler = () => {
+      setTop(window.scrollY <= 320);
+    };
+    window.addEventListener('scroll', scrollHandler);
+    return () => window.removeEventListener('scroll', scrollHandler);
+  }, [top]);
+
   return (
     <div className="traffic-now design-system">
       {!isMobileCanceledTripsView && !(isDetailsView && mobile) && (
@@ -84,7 +94,12 @@ const TrafficNow = () => {
                     <Filters />
                   </div>
                 ) : (
-                  <div className="traffic-now__filters-button-container">
+                  <div
+                    className={cx(
+                      'traffic-now__filters-button-container',
+                      !top && 'scrolled',
+                    )}
+                  >
                     <FiltersModal
                       isOpen={showFiltersModal}
                       onClose={() => setShowFiltersModal(false)}
--- a/app/component/trafficnow/TrafficNow.js
+++ b/app/component/trafficnow/TrafficNow.js
@@ -17,6 +17,9 @@ import Filters from './filters/Filters';
 import { FilterContextProvider } from './filters/FiltersContext';
 import FiltersModal from './filters/FiltersModal';
 
+// defines the scroll position at which a drop shadow is applied to the header button on mobile
+const HEADER_HEIGHT = 320;
+
 const TrafficNow = () => {
   const {
     match: {
@@ -42,11 +45,11 @@ const TrafficNow = () => {
 
   useEffect(() => {
     const scrollHandler = () => {
-      setTop(window.scrollY <= 320);
+      setTop(window.scrollY <= HEADER_HEIGHT);
     };
     window.addEventListener('scroll', scrollHandler);
     return () => window.removeEventListener('scroll', scrollHandler);
-  }, [top]);
+  }, []);
 
   return (
     <div className="traffic-now design-system">
--- a/app/component/trafficnow/components/DisruptionStatus.js
+++ b/app/component/trafficnow/components/DisruptionStatus.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { useIntl } from 'react-intl';
+import { DateTime } from 'luxon';
 import Icon from '../../Icon';
 import { getFormattedTimeDate } from '../../../util/timeUtils';
 import { useConfigContext } from '../../../configurations/ConfigContext';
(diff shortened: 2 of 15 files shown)


app/component/trafficnow/components/DisruptionStatus.js
```diff
@@ -24,9 +25,16 @@ export default function DisruptionStatus({
       ? active
       : now > effectiveStartDate * 1000 && now < effectiveEndDate * 1000;
 
+  const endsSameYear =
+    effectiveStartDate &&
```
sharhio, line 30: Confusing naming. Ends the same year but uses only start date for comparison?

app/component/trafficnow/filters/Filters.js
```diff
@@ -49,12 +49,41 @@ const Filters = ({ onApplyClick = undefined, onResetClick = () => {} }) => {
...
+            variant="blue"
+            value={intl.formatMessage({
+              id: 'traffic-now_filters_view-results',
+            })}
+            onClick={onApplyClick}
+          />
+        )}
+        <Button
+          type="button"
+          size={mobile ? 'medium' : 'small'}
```
sharhio: This always returns 'medium' because this code only ever happens when 'mobile' is true.

app/component/trafficnow/filters/VehicleModesFilter.js
```diff
@@ -38,16 +42,30 @@ const VehicleModesFilter = ({ filterId }) => {
     [],
   );
   return (
-    <fieldset>
+    <fieldset className={isMobile && 'mobile'}>
```
sharhio: Causes a console warning when isMobile is false. (Received `false` for a non-boolean attribute `className`.) 
`cx({ mobile: isMobile })` would solve this.

app/component/trafficnow/TrafficNow.js
```diff
@@ -38,6 +38,16 @@ const TrafficNow = () => {
...
 
+  const [top, setTop] = useState(true);
+
+  useEffect(() => {
+    const scrollHandler = () => {
+      setTop(window.scrollY <= 320);
+    };
+    window.addEventListener('scroll', scrollHandler);
+    return () => window.removeEventListener('scroll', scrollHandler);
+  }, [top]);
```
sharhio: Keeps recreating the event listener whenever user scrolls up due to "top" being a dependency. Empty dependency array would be more appropriate so this would only be run once. 

320 is also pretty magical. Perhaps a named constant?

app/component/trafficnow/filters/VehicleModesFilter.js
```diff
@@ -38,16 +42,30 @@ const VehicleModesFilter = ({ filterId }) => {
...
-          <label htmlFor={`vehicleModes-${option}`} className="input-label">
+        <div
+          key={option}
+          role="checkbox"
+          aria-checked={selectedFilters[filterId]?.includes(option) || false}
+          aria-label={intl.formatMessage({ id: option.toLowerCase() })}
+          tabIndex={0}
+          className={cx(
+            'traffic-now__filters-mode-option',
+            isMobile && 'mobile',
```
sharhio: Same as above.

app/component/trafficnow/filters/Filters.js
```diff
@@ -53,12 +49,41 @@ const Filters = ({ onApplyClick = undefined, onResetClick = () => {} }) => {
...
-      })}
-    >
+  return mobile ? (
+    <div>
+      <form className="traffic-now__filters traffic-now__filters-mobile">
+        {components.map(({ id, Component }) => (
+          <Component key={id} filterId={id} />
+        ))}
+      </form>
+      <div className={cx('traffic-now__filters-form-buttons')}>
```
sharhio: Doesn't really require cx.

app/component/trafficnow/filters/Filters.js
```diff
@@ -53,12 +49,41 @@ const Filters = ({ onApplyClick = undefined, onResetClick = () => {} }) => {
...
+        {onApplyClick && (
```
sharhio: Buttons are exactly the same for mobile and desktop. Perhaps that part of code could be shared?

app/component/trafficnow/filters/VehicleModesFilter.js
```diff
@@ -56,13 +74,13 @@ const VehicleModesFilter = ({ filterId }) => {
...
             {intl.formatMessage({ id: option.toLowerCase() })}
           </label>
-          <input
-            id={`vehicleModes-${option}`}
-            type="checkbox"
-            checked={selectedFilters[filterId]?.includes(option)}
-            value={option}
-            onChange={() => handleCheck(option)}
-          />
+          <span
```
sharhio, line 77: What is the reasoning behind changing this from an actual input to a span? Generally, the native components support accessibility better than custom creations.

sharhio (changes requested):
Everything looks good, but a few minor issues to fix in the code.

app/component/trafficnow/filters/VehicleModesFilter.js
hjvuor: Did not find a way to define a checkbox with the custom svg symbol as icon using native elements. Also since we wanted the whole div to be clickable and not just the checkbox I think it made sense to just think of the "checkbox" as a sort of selection indicator.

app/component/trafficnow/components/DisruptionStatus.js
```diff
@@ -25,15 +25,18 @@ export default function DisruptionStatus({
...
       : now > effectiveStartDate * 1000 && now < effectiveEndDate * 1000;
 
-  const endsSameYear =
+  // We want to hide start year when both dates are the current year
+  const hideStartYear =
     effectiveStartDate &&
-    DateTime.fromSeconds(effectiveStartDate).year === DateTime.now().year;
+    effectiveEndDate &&
+    DateTime.now().year === DateTime.fromSeconds(effectiveStartDate).year &&
+    DateTime.now().year === DateTime.fromSeconds(effectiveStartDate).year;
```
sharhio: One of these should be the effectiveEndDate.

sharhio (changes requested):

PR #5863, opened 2026-06-12
AB#511 - Show direct and transit flex itineraries when enabled

VillePihlava:
## Proposed Changes

  - This PR adds the capability to flex to show both a direct and transit route. In order for this to happen both `direct`, `transit`, and the new config field `showBothDirectAndTransitResults` need to be enabled.
  - I removed a check that was always true because comparing numbers to the string TAXI doesn't make sense `(includeTaxiSuggestions || leg.route?.type !== 'TAXI'),`
  - Refactoring changes

### DIFF
--- a/app/component/itinerary/ItineraryPageUtils.js
+++ b/app/component/itinerary/ItineraryPageUtils.js
@@ -11,7 +11,11 @@ import {
 import { PlannerMessageType, ExtendedRouteTypes } from '../../constants';
 import { addAnalyticsEvent } from '../../util/analyticsUtils';
 import { boundWithMinimumArea } from '../../util/geo-utils';
-import { compressLegs, getTotalBikingDistance } from '../../util/legUtils';
+import {
+  compressLegs,
+  getTotalBikingDistance,
+  isDirectFlexItinerary,
+} from '../../util/legUtils';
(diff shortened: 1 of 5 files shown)


app/component/itinerary/ItineraryPageUtils.js
```diff
@@ -411,23 +402,20 @@ export function filterItineraries(edges, modes) {
...
-    edge.node.legs.some(
-      leg =>
-        types.includes(leg.route?.type) &&
-        (includeTaxiSuggestions || leg.route?.type !== 'TAXI'),
-    ),
+    edge.node.legs.some(leg => types.includes(leg.route?.type)),
   );
 }
 
+function selectFlexEdges(edges, types, showBothDirectAndTransitResults) {
```
sharhio, line 424: This function assumes a few things. Firstly, that it receives only flex edges, and also that the flex edges are in the correct order. The logic works due to how this function is currently called and how otp works, but I think this should be at least documented. The reason being that there is no logic here that guarantees that it always gives the expected results (either just one transit+flex journey, or also a direct flex journey).

VillePihlava: I added documentation to this function and `sortAndMergePlans`

sharhio (approved):

PR #5865, opened 2026-06-12
Ab#623 Adjustments to traffic now overview cards

sharhio:
## Proposed Changes

  - Each transport mode has its own card (for the same alert) to avoid cluttering the cards.
  - Max 5 routes are shown at a time, then  "+ number of additional hidden ones".
  - Only route-related icons are shown when alerts relate to routes as well as stops (stops are included in the drill-down view). 
  - Stop-only alerts have their own card.

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/configurations/config.default.js
+++ b/app/configurations/config.default.js
@@ -731,8 +731,11 @@ export default {
   vehicles: false,
   showVehiclesOnStopPage: false,
   showVehiclesOnItineraryPage: false,
-  trafficNowLink: false,
+  trafficNowLink: true,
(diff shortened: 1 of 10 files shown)


app/configurations/config.default.js
```diff
@@ -731,8 +731,11 @@ export default {
   vehicles: false,
   showVehiclesOnStopPage: false,
   showVehiclesOnItineraryPage: false,
-  trafficNowLink: false,
+  trafficNowLink: true,
```
hjvuor: why is this changed to true?

hjvuor (approved):

PR #5866, opened 2026-06-12
AB#468 Add translations and fix styles of stop and route page disruption view

hjvuor:
## Proposed Changes

  - Replaces placeholder translations with proper ones
  - fix minor spacing issues
  - route badge background corresponds to the mode
  - remove shadow from disruption card, adds a border and a hover effect, make the whole item clickable

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/Disruption.js
+++ b/app/component/Disruption.js
@@ -9,7 +9,7 @@ import DisruptionBadge from './trafficnow/DisruptionBadge';
 import { useConfigContext } from '../configurations/ConfigContext';
 import { routePagePath, stopPagePath } from '../util/path';
 import IconBackground from './icon/IconBackground';
-import { getRouteMode } from '../util/modeUtils';
+import { getRouteMode, transitIconName } from '../util/modeUtils';
 import { getStartTimeWithColon } from '../util/timeUtils';
 import { entityShape, stopTimeShape } from '../util/shapes';
 import {
@@ -86,7 +86,7 @@ export default function Disruption({
       <div
         className="alert-row"
         role="button"
-        aria-label={alertDescriptionText}
+        aria-label={`${alertHeaderText} ${buttonLabel}`}
         onClick={toggleDetails}
         tabIndex={0}
         onKeyDown={e => {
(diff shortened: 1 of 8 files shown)


app/component/Disruption.js
```diff
@@ -81,97 +82,120 @@ export default function Disruption({
...
+            }
+            e.stopPropagation();
+          }
+        }}
+      >
+        {toggleDetails && (
+          <button
+            type="button"
+            onClick={e => {
+              if (toggleDetails) {
```
sharhio: toggleDetails should always be true due to the earlier check

app/component/Disruption.js
```diff
@@ -81,97 +82,120 @@ export default function Disruption({
...
+        <div className="alert-row-badges">
+          {groupedEntities &&
+            Object.entries(groupedEntities).map(
+              ([key, { typename, mode, items }]) => {
+                const isStop = typename === AlertEntityType.Stop;
+                return (
+                  <Fragment key={key}>
+                    <Icon
+                      img={`icon_${
+                        mode.toLowerCase() === 'bus-express'
```
sharhio: Could consider using the modeUtils function transitIconName here.

app/component/disruption.scss
```diff
@@ -105,16 +111,37 @@
...
+
+    .icon-container {
+      color: $theme-primary-color;
+      display: flex;
+      align-items: center;
+    }
+  }
+
+  .alert-row:hover {
+    border: 1px solid #0074bf;
```
sharhio: Should this always be blue or should it be themed?

app/component/Disruption.js
```diff
@@ -81,97 +82,120 @@ export default function Disruption({
...
+          if (e.key === ' ' || e.key === 'Enter') {
+            if (toggleDetails) {
+              toggleDetails();
+            }
+            e.stopPropagation();
+          }
+        }}
+      >
+        {toggleDetails && (
+          <button
```
sharhio: Nested interactive elements isn't really great for accessibility. Because the entire card is a button, is it really necessary to have a separate button inside of it?

sharhio (commented):
Looks good, but there are a couple of questions.

app/component/disruption.scss
hjvuor: I asked from design and they clarified that we'll use the themed primary color 👍

app/component/Disruption.js
hjvuor: I removed onclick behaviour and replaced button with a div, I also added the button label (View details/View timetable) to the aria description of the actual div acting as a button.

hjvuor:
added a fix to a bug where disruption id was not passed correctly to toggledetails for future alerts

PR #5870, opened 2026-06-15
AB#605 reactivation of personalization

vesameskanen:
A modal asks if personalization history should be kept or emptied. Opens only if there is existing history. 

This PR also adds feedback element as a teaser to itinerary detail view of non-logged users which have not yet turned personalization off. 
 
Some components modernized as well.

### DIFF
--- a/app/component/itinerary/Feedback.js
+++ b/app/component/itinerary/Feedback.js
@@ -124,18 +124,16 @@ export default function Feedback({
   }, []);
 
   const handleGiveFeedback = value => {
-    if (giveFeedback) {
-      if (isAnimating) {
-        return;
-      }
-
-      setIsAnimating(true);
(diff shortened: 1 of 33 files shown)


app/component/itinerary/Feedback.js
```diff
@@ -104,16 +124,18 @@ export default function Feedback({
   }, []);
 
   const handleGiveFeedback = value => {
-    if (isAnimating) {
-      return;
-    }
+    if (giveFeedback) {
+      if (isAnimating) {
+        return;
+      }
```
VillePihlava:
```
if (!giveFeedback || isAnimating) {
  return;
}
```

VillePihlava (commented):
Looks good, just one small thing (and the z-index stuff if thats still relevant)

PR #5883, opened 2026-07-03
AB#650 - Add car pickup taxi sandbox support

VillePihlava:
## Proposed Changes

**BEFORE MERGING VERIFY ITINERARY LOGIC, ESPECIALLY RELATED TO RENTAL FIRST LEG LOGIC**

  - Support for car pickup zone taxis through flex data.
  - Lots of refactorings, including:
    - `FirstLegStartTime` logic for scooter and bike rentals were not working at all, now enabled **NOTE** this could cause errors because it was unused code
    - Lots of util function refactoring
    - Big Itinerary.js refactoring to make code more clean
    - Remove unused variables/functionality in Itinerary.js related to `onlyIconLegsLength` and `iconLegsInPercents `
    
**Whats missing**
- How should emissions be handled?
- Should a relaxed query be done similar to external flex (taxis)?


app/component/itinerary/Itinerary.js
```diff
@@ -510,123 +497,31 @@ const Itinerary = ({
       );
     }
   });
-  const normalLegs = legs.length - onlyIconLegs;
+  const normalLegs = legs.length - onlyIconLegCount;
```
tkalvas: normalLegCount

tkalvas (changes requested):
Change the variable name to xxxCount as above, to be in line with other similar changes in this PR.

app/component/itinerary/Itinerary.js
VillePihlava: fixed

PR #5896, opened 2026-07-31
Fix trafficnow filtering

sharhio:
## Proposed Changes

  - When single alerts are split into mode specific cards the mode filtering was not applied correctly and unwanted mode cards were shown. Splitting now takes into account the selected filters.
  - Also did some refactoring and added tests

## Pull Request Check List

## Review

  - Merge the pull request

### DIFF
--- a/app/component/trafficnow/Disruptions.js
+++ b/app/component/trafficnow/Disruptions.js
@@ -64,8 +64,7 @@ export default function Disruptions() {
   );
 
   const disruptionCards = useMemo(
-    () =>
-      buildDisruptionCards(disruptions, selectedFilters.vehicleModes, config),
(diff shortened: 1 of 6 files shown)


hjvuor (changes requested):
Filtering by mode works now but if searching for a specific route or just favorite routes the duplicate cards for other modes are shown.

hjvuor (changes requested):
I think this might've been an already existing bug but the bus filter seems to leave out bus-local and bus-express entities (might probably also be bugged for speedtrams and replacement buses but couldn't verify quickly). modeUtils.js contains the getRouteMode function that has the correct mapping but it does not seem like such is used anywhere in the trafficnow page.

hjvuor (approved):

PR #5897, opened 2026-08-04
AB#659 - Add first leg text for citybikes and scooters

VillePihlava:
## Proposed Changes

  - Scooters and citybikes now have their own first leg text.
  - Scooters, citybikes, and taxis all now use their mode-specific translation keys.
  - Combine screen reader and visual text keys into one util function
  - Remove unused translation keys
  - Rename translation keys for clarity


vesameskanen (commented):
Only remark is that we are trying to get rid of defaultProps and use default values in function component params (see FirstLegStartTime)

PR #5904, opened 2026-08-10
AB#675 refactor and fix transit mode translations for screen readers 

vesameskanen:
- Translations included two versions for transit mode names, plain mode and longer version with shortname and headsign postfix. This is plain redundance because we can generate postfixed version by appending parameters in code.
- Some special modes such as bus-local did not have translations.  Screen reader outputted nonsense. Fixed by the refactoring above.
- Airplane leg used uppercase mode, so that translation was never found. Special AirplaneLeg component removed completely as it adds nothing useful to Transitleg. 
- Broken occupancy status screen reader translations fixed by refactoring


app/component/itinerary/Itinerary.js
```diff
@@ -461,15 +462,7 @@ const Itinerary = ({
...
-        formatMessage(
-          {
-            id: `${leg.mode.toLowerCase()}-with-route-number`,
-          },
-          {
-            routeNumber: routeName,
-            headSign: '',
-          },
-        ),
+        `${formatMessage({ id: leg.mode.toLowerCase() })} ${routeName || ''}}`,
```
VillePihlava: There is an unfortunate thing related to this. Because of legacy reasons, the `rail` translation is `Commuter train`.

You would have to change rail to the correct general term `train`, but support for HSL's local trains should stay.

VillePihlava: Also, I think you have an extra `}`

app/component/itinerary/TransitLeg.js
```diff
@@ -467,14 +467,8 @@ export default function TransitLeg({
...
-            <FormattedMessage
-              id={`${mode}-with-route-number`}
-              values={{
-                routeNumber: leg.route?.shortName,
-                headSign: leg.trip?.tripHeadsign,
-              }}
-              defaultMessage={`${mode} {routeNumber} {headSign}`}
-            />
+            `${intl.formatMessage({ id: mode })} ${leg.route?.shortName || ''} $
+            {leg.trip?.tripHeadsign || ''}`
```
VillePihlava: The formatting on this is weird with $ on a separate line to the {} block, our linter configuration should maybe be looked at.

VillePihlava: Actually, you are missing the surrounding {} from the expression

app/component/CapacityModal.js
```diff
@@ -36,7 +36,7 @@ const CapacityModal = ({ config }) => {
           </div>
           <p className="capacity-info-explanation">
             <FormattedMessage
-              id="capacity-modal.many-seats-available-body"
+              id="capacity-modal.many-seats-available"
```
VillePihlava, line 39: Is the `capacity-modal.full-capacity` translation missing from this list?

test/unit/util/occupancyUtil.test.js
```diff
@@ -51,40 +51,40 @@ describe('occupancyUtil', () => {
   describe('capacityToTranslationId', () => {
     it('should map EMPTY and MANY_SEATS_AVAILABLE to many-seats heading', () => {
```
VillePihlava: The text on this line is wrong

VillePihlava: Also, check the other tests as well

VillePihlava (changes requested):
Nice deletions! However, I think the legacy `rail` mode = `Commuter train` will cause a lot of work.

app/component/itinerary/LegInfo.js
```diff
@@ -33,7 +33,7 @@ export default function LegInfo({
   mobile = false,
   isTransitLeg,
 }) {
-  const intl = useIntl();
+  const { formatMessage } = useIntl();
```
VillePihlava, line 36: I would keep it intl = useIntl(); this makes it very clear that it comes from intl, not some other formatMessage function

VillePihlava (changes requested):
`settings-alternative-name-rail` is still used in `NaviUtils.js`

app/component/CapacityModal.js
vesameskanen: hmm good point. It seems that  icon mapping and capacity modal want to map full to crushed-standing-room, whereas screen reader messages use full status as well.

This capacity handling mess has been a real pain.

app/component/itinerary/LegInfo.js
vesameskanen: Hmm this is normal destructring, which is recommended by eslint. Why to use 2 steps for accessing a 
single field?

hjvuor: imo this destructuring syntax is cleaner, it discards of any unnecessary stuff in the intl object and is also commonly used with other hooks

VillePihlava: Similarly named functions can currently be found in:
- https://github.com/HSLdevcom/digitransit-ui/blob/v3/scripts/sort-translations.mjs
- https://github.com/HSLdevcom/digitransit-ui/blob/v3/server/reittiopasParameterMiddleware.js

If in the future someone were to add a `format*` function, this would make it very clear that this one comes from `intl`. I also think having clarity on whether a parameter comes from the config is important. So I disagree with e.g. `{ colors } = useConfigContext()`. However, this is an opinion on code style. I like the intl.* prefix because its nicer to grep.

hjvuor: i'm not entirely sure how js optimizes stuff but in the config example you can either have the entire config in memory or just the relevant `colors` attribute so its not just a code style difference if im not mistaken.

VillePihlava (approved):

PR #5909, opened 2026-08-19
AB#666 personalization style fixes

vesameskanen:
- Use of hsl design system elements increased to apply exact design styles in personalization and popover
- Various style leaks into design system elements fixed
- Default color theme overruling of design system did not work in prod build, fixed by style sheet load order rearrangement in app/server.js
- Regression in custom setting popover visibility fixed


app/component/itinerary/customizesearch/PrModal.js
```diff
@@ -72,16 +80,20 @@ export default function PrModal({ closeModal }) {
...
+              <Text variant="text-xs">
+                {formatMessage({ id: 'personalization-beta' })}
+              </Text>
+              <Spacer size="xxs" />
+              <div>
+                <ButtonLink
+                  size="s"
+                  variant="primary"
+                  expandOnMobile
+                  href="/foo"
```
VillePihlava, line 92: Is it correct to have a placeholder here? E.g. should it come from the config?

app/component/itinerary/customizesearch/customize-search.scss
VillePihlava: I think this can be removed. The styles are incorrect because they have '', and there is another style that overrides this one.

app/component/itinerary/customizesearch/customize-search.scss
```diff
@@ -221,7 +227,9 @@
       }
 
       span {
-        font-size: $font-size-small;
+        font-size: var(--Font-Size-Routes-Extra-small, 12px);
+        line-height: var(--Line-Height-Routes-XS, 14px);
+        letter-spacing: var(--letter-spacing-03, -0.3px);
```
VillePihlava, line 224: Does this make sense? Shouldn't the fallback be in defaults? Also, are the variables even defined somewhere?

app/component/itinerary/personalization.scss
```diff
@@ -236,3 +236,11 @@
...
   }
 }
+
+.mobile {
+  .feedback-section {
+    .thumb-button {
+      margin-left: var(--space-fixed-xs);
+    }
+  }
+}
```
VillePihlava, line 246: This file should be name `personalization.scss`

VillePihlava (commented):
I wrote a couple of comments

app/component/itinerary/customizesearch/PrModal.js
vesameskanen: True, but I am still waiting crazyEgg url, to be added into HSL config.

app/component/itinerary/customizesearch/customize-search.scss
vesameskanen: Variables apparently are under way to the design system.

VillePihlava (approved):

PR #5914, opened 2026-08-21
Ab#619 Add stop status badges to search results

sharhio:
## Proposed Changes

  - Alert based statuses are fetched by UI at startup and applied at search time
  - Calendar based statuses come from pelias search results 

Depends on https://github.com/HSLdevcom/pelias-gtfs/pull/23
AND https://github.com/HSLdevcom/pelias-data-container/pull/108, https://github.com/HSLdevcom/pelias-gtfs/pull/24

## Pull Request Check List

## Review

  - Merge the pull request


vesameskanen (commented):
I wonder if alert parsing should be moved to geocoding data loading? In other words, GTFS addendum would also contain stop alert related information. GTFS loader could make a graphql query before starting the actual GTFS parsing streams.

Now HSL app and hsl.fi  will have to duplicate and maintain similar alert mapping code.

pelias-gtfs would then need API keys teh same way as bikes-pelias, see https://github.com/HSLdevcom/bikes-pelias/blob/master/lib/importPipeline.js

sharhio:
> I wonder if alert parsing should be moved to geocoding data loading? In other words, GTFS addendum would also contain stop alert related information. GTFS loader could make a graphql query before starting the actual GTFS parsing streams.

Sure. I didn't do it right away, because the GTFS loader doesn't currently do queries like that and there was talk of perhaps avoiding too much extra information in the addendum. The app having to duplicate the mapping is a very good point, though.

digitransit-component/packages/digitransit-component-suggestion-item/src/index.js
```diff
@@ -197,9 +197,23 @@ const SuggestionItem = memo(
...
         <Icon color={iconColor} img={iconId} />
+        {stopStatusBadge && (
+          <svg
+            width="13"
+            height="13"
+            viewBox="0 0 40 40"
+            className={styles['suggestion-status-badge']}
+            aria-hidden="true"
+          >
+            <use href={`#${stopStatusBadge}`} />
```
vesameskanen: I wonder how this reference resolves - to an icon from the sprite sheets?
It would be nice if suggestion item would not depend from some external asset wiring. Digitransit components have an icon library for cases like this.

vesameskanen (commented):
I think use of SearchContext is just unnecessary complexity, which makes it hard for hsl.fi to use our search. Please move getStopBadge code to suggestion item, even if it requires duplication of icon name mapping.

digitransit-component/packages/digitransit-component-icon/src/index.js
```diff
@@ -193,18 +202,20 @@ const Icon = ({ color, img, height, width, rotate }) => {
...
   width: PropTypes.number,
   height: PropTypes.number,
   rotate: PropTypes.string,
 };
 
 Icon.defaultProps = {
+  className: undefined,
   color: undefined,
   width: undefined,
   height: undefined,
```
vesameskanen: Please remove default props, especially when default values are useless like here.

digitransit-component/packages/digitransit-component-suggestion-item/src/index.js
```diff
@@ -58,6 +58,72 @@ function isFavourite(item) {
...
+
+// Layers eligible for a stop status badge in search suggestions.
+const STATUS_LAYERS = new Set([
+  'stop',
+  'favouriteStop',
+  'station',
+  'favouriteStation',
+]);
+
+function extractGtfsId(item) {
```
vesameskanen: I guess a minimally simple boolean check is enough here. Actual GTFS id is not needed.

digitransit-component/packages/digitransit-component-suggestion-item/src/index.js
```diff
@@ -431,13 +508,15 @@ SuggestionItem.propTypes = {
   getAutoSuggestIcons: PropTypes.objectOf(PropTypes.func),
   colors: PropTypes.objectOf(PropTypes.string),
   modeSet: PropTypes.string,
+  showStopStatusMarkers: PropTypes.bool,
 };
 
 SuggestionItem.defaultProps = {
```
vesameskanen: Good opportunity to move default props into function parameter declaration.

vesameskanen (changes requested):

vesameskanen (approved):

PR #5921, opened 2026-08-26
Fluxible timeStore converted to a react context

vesameskanen:


app/component/AlertList.js
```diff
@@ -12,6 +11,7 @@ import {
   isAlertValid,
 } from '../util/alertUtils';
 import { alertShape } from '../util/shapes';
+import { withCurrentTime } from '../hooks/TimeContext';
```
hjvuor: instead of wrapping AlertList in withCurrentTime HOC, you should use the useCurrentTime hook inside the component since its a function, and only use withCurrentTime in case of class components.

app/component/AlertList.js
```diff
@@ -129,12 +129,6 @@ AlertList.defaultProps = {
...
 };
 
-const connectedComponent = connectToStores(
-  withBreakpoint(AlertList),
-  ['TimeStore'],
-  context => ({
-    currentTime: context.getStore('TimeStore').getCurrentTime(),
-  }),
-);
+const connectedComponent = withCurrentTime(withBreakpoint(AlertList));
```
hjvuor: With useCurrentTime, theres no need to define connectedComponent and you can just export the component itself. Also the connectedComponent name is misleading since it no longer connects the component to any store.

app/component/DisruptionBanner.js
hjvuor, line 1: Same thing here as with AlertList

app/component/IndexPage.js
hjvuor, line 1: here we can also use useCurrentTime directly inside the function component

hjvuor (changes requested):
Noticed some possible improvements in case of functional components where we should use the useTimeContext hook inside the components instead of the HOC. This should be done everywhere possible in addition to the few files i commented on.

hjvuor (approved):

PR #5926, opened 2026-08-27
Combine providers using reduce to avoid nesting jsx syntax

hjvuor:
## Proposed Changes

  - Defines an AppReducer component that has a list of all context providers and uses reduceRight to wrap them in order around the children.

This type of approach is from https://newsletter.daishikato.com/p/the-past-and-future-of-render-optimization-with-react-context and https://medium.com/@d3d.me/react-provider-composition-exorcising-provider-hell-d98cd3a78b29

### DIFF
--- a/app/client.js
+++ b/app/client.js
@@ -52,6 +52,34 @@ const { config } = window;
 const app = appCreator(config);
 const context = app.createContext({ config });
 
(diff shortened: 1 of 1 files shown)


vesameskanen (approved):
nice

PR #5929, opened 2026-08-30
Fix build/dev environment checking, and remove old test mode

VillePihlava:
## Why

`app/util/envUtils.js` exported one blurred `IS_DEV` constant mixing two unrelated axes
(`process.env.RUN_ENV === 'development' || process.env.NODE_ENV !== 'production'`). Since #5726
dropped its `config` parameter and deleted `config.RUN_ENV`, run-environment detection has been
**dead in the browser** — `process.env.RUN_ENV` doesn't exist there and `NODE_ENV` is baked to
`"production"`, so `IS_DEV` was permanently `false` in the bundle (e.g. `showCitybikeNetwork`'s
dev branch never ran). Client code also carried ~8 `process.env.NODE_ENV === 'test'` guards that
were likewise dead — `yarn build` runs webpack `mode: production` (webpack rejects `mode: test`),
so the test value never reached any bundle.

## What changed

**`app/util/envUtils.js`** — one blurred const → two focused signals:

| | meaning | source |
|---|---|---|
| `IS_DEV_BUILD` | dev build / local `yarn dev` server | `NODE_ENV === 'development'` (baked into the client bundle by webpack, live read in Node). Unset `NODE_ENV` (CI, mocha, bare `node server/server`) ⇒ real build. |
| `isDevRunEnv(config)` | non-production deployment tier | `RUN_ENV` (`development` / `production`; unknown ⇒ production). Client passes `window.config`; server omits the arg to read `process.env.RUN_ENV`. |

- **Config assembly** (`config.*.js`, `realtimeUtils.js`): `IS_DEV` → `isDevRunEnv()`.
- **`config.default.js`**: adds `RUN_ENV` and `SURVEY_SHARE` to the config object, so the client
  can read the tier via `window.config.RUN_ENV` and `analyticsUtils` no longer reads
  `process.env` from a client-shipped module.
- **Client build-mode checks** (`client.js`, `AppBarHsl`, `BackButton`) and **Node build-mode
  checks** (`app/config.js`, `app/server.js`, `server/server.js`,
  `server/passport-openid-connect/openidConnect.js`): `process.env.NODE_ENV` comparisons →
  `IS_DEV_BUILD`.
- **Dead `NODE_ENV === 'test'` guards removed**: `TimeContext` (frozen clock), `AppBar`
  (`<MessageBar>`), `ItineraryPageUtils`, `TransitLeg`, `RouteControlPanel` (3, incl. a
  `config.NODE_ENV` reference to a field that never existed), `realTimeClientAction`, `fareUtils`.
- **`showCitybikeNetwork` / `showRentalVehiclesOfType`** take an optional `config` and call
  `isDevRunEnv(config)` — dev/staging deployments show every citybike network off-season again
  (regressed by #5726).
- **`package.json`**: `start-test` removed; e2e jest-playwright configs run `yarn start` directly.
  `dev` / `dev-nowatch` set `RUN_ENV=development`. `start` deliberately does **not** set `RUN_ENV`
  so Kubernetes can (a script-level `RUN_ENV=production` would override the pod env).
- Docs: `Installation.md` / `Docker.md` describe the two signals and point at the `envUtils`
  JSDoc.

## Behavior changes to be aware of

All in the fail-safe direction (identical under a normal `NODE_ENV=production` deployment):

- **OIDC session cookies** (`openidConnect.js`): `secure` / `httpOnly` / `sameSite` were
  `NODE_ENV === 'production'`, now `!IS_DEV_BUILD`. Under an unset `NODE_ENV` they flip on (were
  off). Local `yarn dev` sets `NODE_ENV=development`, so dev login is unchanged.
- **500 responses** (`server/server.js`): full stack trace only when `NODE_ENV=development` now
  (was: anything except `production`). Stack is still in the server log.
- **e2e** now boots via `yarn start` instead of `start-test` (`NODE_ENV=test`): `window.config`
  carries prod backend URLs (old `start-test` had `IS_DEV` true ⇒ dev URLs) and 500s are terse.
  The suite mocks GraphQL/weather via `page.route()` and is not in CI (snapshots from 2022), so
  this only matters when snapshots are next regenerated.

`window.config.RUN_ENV` / `SURVEY_SHARE` serialize as `undefined` when their env vars are unset;
every reader normalizes (`isDevRunEnv`'s `|| production`, `analyticsUtils`' `|| 250`).

## Follow-up (not this PR)

The e2e visual suite is dormant (not in CI, 2022 snapshots) and its determinism relied on
request mocking, not the deleted guards. A proper revival — bump Playwright off `1.16.2`, freeze
time with `page.clock`, mock the service-alerts query + block the MQTT socket in the harness,
regenerate snapshots, add `test:e2e` to CI — is its own task. `config.RUN_ENV === 'test'` is
available as a single app hook if that work wants one.


app/client.js
```diff
@@ -107,7 +108,7 @@ async function init() {
 
   window.context = context;
 
-  if (process.env.NODE_ENV === 'development') {
+  if (IS_DEV_BUILD) {
```
vesameskanen: This must stay a literal `process.env.NODE_ENV === 'development'` check, not `IS_DEV_BUILD` (or any other imported/computed value).

Webpack's DefinePlugin replaces the literal `process.env.NODE_ENV` expression with a constant string at each call site, and webpack's own parser (not just the minifier) performs dead-code elimination on conditions written in exactly this form *during the build/graph phase*. When the condition folds to `false`, the whole branch — including any `require()` inside it — is dropped before webpack resolves it, so it's never added to the module graph.

`IS_DEV_BUILD` is just an imported variable, so webpack can't statically fold it away. The `if` branch survives, and webpack still has to resolve the dynamic `require(`../sass/themes/${config.CONFIG}/main.scss`)` inside it. Since `config.CONFIG` isn't statically known, webpack builds a context module that bundles *every* theme's `main.scss` (all ~23 regions) into the shared CSS chunk — which is how e.g. Oulu's pink theme colors leaked into a production HSL build.

TL;DR: this isn't just a truthy/falsy check, it's a build-time dead-code-elimination trigger that only works for the exact `process.env.NODE_ENV === 'development'` pattern. Swapping in a helper function or imported constant breaks the elimination and causes all themes to be bundled together.

vesameskanen (changes requested):

vesameskanen (approved):

PR #5960, opened 2026-09-09
Replace FavouriteStore (Fluxible) with a React Context

vesameskanen:
## Summary

Migrates favourites state management from the Fluxible `FavouriteStore` to a React Context-based implementation (`FavouriteProvider` + `useFavourites`/`useFavouriteStatus`/`useFavouriteActions` hooks), reducing reliance on the legacy Fluxible architecture. Along the way, the framework-agnostic singletons this introduced (and a related one for search) were consolidated into a new `app/data/` folder, reserved for non-Fluxible, non-React "data"/service classes produced by the ongoing hooks migration - each following the same `init()`-once-at-bootstrap pattern.

### Favourites: Fluxible store → React Context + data singleton

- Added `app/hooks/FavouriteContext.js` (Provider + hooks); wired into `client.js`.
- Migrated ~17 consumers (e.g. `FavouritesContainer`, `FavouriteStopContainer`, `IndexPage`, `NearYouPage`, `AppBarContainer`, `trafficnow` components) from `connectToStores`/`FavouriteActions` to the new hooks.
- Removed `FavouriteActions.js` (Fluxible action creators, now unneeded).
- Moved `app/store/FavouriteStore.js` → `app/data/FavouriteData.js`: renamed to avoid confusion with the legacy Fluxible stores it replaces, and consolidated the standalone pure query functions (`isFavourite`, `getFavouriteByGtfsId`, `getFavouriteRouteGtfsIds`, `getFavouriteStopsAndStations`, `getFavouriteVehicleRentalStations`, `getFavouritePlaces`, `countLocations`, etc.) into instance methods on the singleton, each still accepting an optional explicit favourites array override.
- Moved `app/store/localStorage.js` → `app/data/localStorage.js` for the same reason (it's a plain persistence helper, not a Fluxible store).
- Removed the now-redundant pass-through wrappers (`getFavouriteLocations`/`getFavouriteRoutes`/`getFavouriteStops`/`getFavouriteVehicleRentalStations`) from `app/util/storeUtils.js`; call sites now call the `FavouriteData` singleton's methods directly.
- Removed a duplicate `favouriteStore.init(config)` call from `FavouriteProvider`'s mount effect: `client.js` already calls it before `ReactDOM.render()` ever mounts the provider.
- Removed `getPositionStore` from `storeUtils.js` - it returned the `PositionStore` Fluxible store directly but had no callers anywhere in the app.

### Search context: merged singleton + initializer

- Merged `app/util/searchContext.js` (a plain object of no-op defaults) and `app/util/DTSearchContextInitializer.js` (the function that populated it with real implementations from the Fluxible context/config) into a single class-based singleton, `app/data/SearchContext.js`.
- The singleton is now initialized once, eagerly, in `client.js` (`searchContext.init(context.getComponentContext())`), mirroring `favouriteStore.init(config)` and running strictly before anything renders. `WithSearchContext.js` no longer lazily calls `init()` from `componentDidMount`.
- Because nothing can ever observe the singleton in an uninitialized state anymore, removed the ~70 lines of per-method no-op default implementations (`getPositions`, `getFavouriteLocations`, etc.) and the no-op field defaults (`context = null`, `feedIDs = []`, `URL_PELIAS = ''`, etc.) that existed purely as a safety net for that (now nonexistent) window. Only `initialized` remains as a field, since it's genuinely read by the `init()` guard.
- Removed the unused, mistyped `getFavouriterRoutes` default and the never-referenced `positionStore` field.
- This singleton's shape remains the contract consumed by the external `@digitransit-search-util` packages (e.g. `digitransit-search-util-execute-search-immidiate`), which only rely on its fields/methods and are unaffected by this internal reorganization. Its `context` field is still the bridge to the not-yet-migrated Fluxible `PositionStore`/`OldSearchesStore`.

### Modernization cleanup on touched components

- Removed `defaultProps` in favor of destructured default parameters.
- Removed dead props (`isFetching`, `favouriteModalAction`).
- Collapsed redundant dual `default`/named `Component` exports to a single default export.
- Removed unnecessary prop-renaming.
- Removed prop-drilling of the static `title` config value (`TopLevel` → `AppBarContainer` → `AppBar` → `LogoSmall`); components now read it from `useConfigContext()` directly.
- Documented the "avoid `defaultProps`" convention in `.github/copilot-instructions.md`.


app/data/FavouriteData.js
hjvuor, line 1: I think the content of this file should be moved somewhere outside the store/ folder since that contains the fluxible stores. should also be renamed to FavouriteData or something to make it distinct from fluxible stores

hjvuor (changes requested):
I would maybe define the FavouriteData singleton in one file and then have the favourite related helper functions somewhere else like inside /utils

app/data/FavouriteData.js
vesameskanen: I agree, I had the same concern myself.

vesameskanen:
Creating data classes and putting the logic which knows what to do with the data to 'function libraries' is anti-OO, old fashioned procedural programming pattern. favourite processing is now handled using class methods.

hjvuor:
> Creating data classes and putting the logic which knows what to do with the data to 'function libraries' is anti-OO, old fashioned procedural programming pattern. favourite processing is now handled using class methods.

should the useFavourites hook then just return the class instance so we wouldn't need to import it in components? Now theres hooks in functions where the return value of useFavourites is not used but all the data comes from the imported class

vesameskanen:
usefavourites hook - although its return value is usually not used - is what makes components update when favourites change. Returning singleton class, which never changes, would stop updates.

vesameskanen:
This was not an easy transition to react context because digitransit components access favourite data without react context. That's why there is the data singleton class and a separate react context provider.

hjvuor (approved):

PR #5966, opened 2026-09-11
AB#698 store personalization data as a favourite

vesameskanen:
Draft until favourite service supports the new type


hjvuor (approved):
looks good

PR #5967, opened 2026-09-11
Message flux store to react context

vesameskanen:
## Summary

Converts the legacy Fluxible `MessageStore` (message bar / notifications) to a React
context (`MessageContext`), and fixes several bugs uncovered along the way in how
message-bar state interacts with re-rendering and layout.

## Changes

- **refactor: convert MessageStore Flux store to React context** — replaces the
  Fluxible `MessageStore` with `app/hooks/MessageContext.jsx`, exposing
  `MessageProvider`/`useMessages()`/`useMessageActions()` for adding and dismissing
  messages.
- **refactor: remove unreachable unsupported-browser check and IE message** — dead
  code cleanup no longer relevant after the store conversion.
- **fix: closing a message/alert banner no longer requires a reload** — messages
  dismissed via the new context are removed immediately instead of only after a
  full page reload.
- **fix: reconnect NaviContainer to MessageContext for immediate re-render** /
  **fix: undefined default prop** — follow-up fixes to keep `NaviContainer`
  subscribed correctly to message state.
- **refactor: remove unnecessary wrapper divs in MessageBar** — simplified
  `MessageBar.jsx` markup (removed redundant nested `<div>`s, merged two adjacent
  wrapper divs), verified against SCSS selectors and existing tests.
- **fix: propagate service-alert and geolocation message dismissal to
  NaviContainer** —
  - `MessageContext`'s `MARK_READ` reducer previously bailed out with the same
    state reference when the target id wasn't tracked in `state.messages`. Live
    service alerts (`MessageBar`) and geolocation messages (`PositionActions`)
    are never added via `addMessage`, so marking them read never changed the
    context value identity, and `useMessages()` subscribers (e.g.
    `NaviContainer`) never re-rendered.
  - `PositionActions.updateGeolocationMessage` still dispatched legacy Fluxible
    actions (`AddMessage`/`MarkMessageAsRead`) that have had no listener since
    `MessageStore` was removed — geolocation messages never reached
    `MessageContext` at all. Added a `messageActions` bridge so plain
    (non-React) modules can call `addMessage`/`markMessageAsRead` on the
    current `MessageProvider`.
- **fix: measure NaviContainer's containerTopPosition after DOM commit** —
  `containerTopPosition` was read from `mapLayerRef.getBoundingClientRect()`
  directly during render, which returned a stale value because React runs all
  render functions of components re-rendering in the same batch before
  committing DOM changes. This caused the navi cards to only move up on the
  next unrelated re-render (5-10s later), and stale/overlapping layout when
  reopening the navigator. Now measured in a `useLayoutEffect`, which runs
  after DOM commit but before paint.

## Testing

- `yarn eslint app/` — 0 errors.
- `yarn test-unit` — 1153 passing, 1 pending, 0 failing.
- Manually verified in dev server: message bar with multiple static messages,
  closing service alerts and geolocation messages, and reopening the navigator
  view — no more delayed layout shift or header/message overlap.


app/client/client.jsx
```diff
@@ -71,7 +72,8 @@ const AppProviders = props => {
     ],
     [IntlBridge],
     [RelayEnvironmentProvider, { environment: props.environment }],
-    [FavouriteProvider, { context: props.context }],
+    [FavouriteProvider],
+    [MessageProvider],
```
sharhio: The order of these has to be swapped, because FavouriteProvider uses MessageProvider. Currently favouriteContext doesn't have access to the real addMessage function in the messageContext, because it's not a descendant of the MessageProvider, so any failure message related to favourites is never shown.

sharhio (changes requested):

app/client/client.jsx
vesameskanen: Really good catch!

PR #5974, opened 2026-09-13
Fix empty string parameter causing fetch in `MessageStore`

VillePihlava:
## Summary

Fixes a browser console error on `matka` deployments where the static-messages
fetch ran with an empty URL and choked parsing the resulting HTML page as
JSON.

- `app/store/MessageStore.js`: `addConfigMessages`'s guard checked
  `staticMessagesUrl !== undefined`, so an empty string (e.g. the
  Dockerfile's default `STATIC_MESSAGE_URL=''`) still passed and triggered
  `fetch('')` — which resolves to the current page and returns HTML, not
  JSON. Changed to a truthiness check.
- Added a regression test asserting no fetch happens when
  `staticMessagesUrl` is `''`.


VillePihlava:
Added to another PR

PR #5990, opened 2026-09-16
Fix vertical alignment of stop code/station text in NearYou header

vesameskanen:
The .itinerary-stop-code chip in the near-you view's second header row had a fixed height but was not itself a flex container, so its text content was not vertically centered inside the bordered box. A margin-bottom hack had been added previously to visually compensate.

Align the styling with the working equivalent on the stop page (route-stop-list.scss): make .itinerary-stop-code a flex container with align-items: center, and add align-items: center to the parent .stop-near-you-info row. Remove the now-unnecessary margin-bottom workaround.


VillePihlava:
Very small difference, but the new one is a bit better

**old**
image

**new**
image

VillePihlava (approved):

PR #5994, opened 2026-09-16
AB#707 prevent stop code plate from overlapping name row

vesameskanen:
The stop-code span's border/padding were painted outside the CSS line box, causing the plate to bleed into the name row above it. Make it an inline-block with an explicit line-height, and add a small margin-top on the label row for clear spacing.


vesameskanen:
version bump after approval

sharhio (approved):
LGTM

PR #5999, opened 2026-09-18
Fix autosuggest always selecting the first item + improvements to state handling

hjvuor:
## Proposed Changes

  - moving away from defaultProps to native defaults caused a bug where the first suggestion was always selected
    - this was caused by the default objects being recreated, fixed by defining them as constants
  - additionally, improve state handling by moving side effects away from downshifts stateReducer
  - replaced the pending enter handling with a React.useRef instead of keeping it in state

## Pull Request Check List

## Review

  - Merge the pull request


hjvuor:
version bump after approval

vesameskanen (approved):

